PR #2, opened 2021-04-21
Config management - Key-Value management support in CLI

tobiscr:


tobiscr:
Closing PR... opening as Draft

cmd/config/get/bucket/cmd.go
```diff
@@ -0,0 +1,38 @@
...
+			return Run(o)
+		},
+	}
+	return cmd
+}
+
+func Run(o *cli.Options) error {
+
+	pgConnFac := &db.PostgresConnectionFactory{
+		Host:     "localhost",
```
crabtree, line 26: This should be read from the configuration

tobiscr: Agree.. but I dropped that already. Per default is now SQLite used. The CLI is already using a config-file to retrieve DB credentials.

pkg/config/cacheentrryentity.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (ce *CacheEntryEntity) checksum() string {
+	if ce.Checksum == "" && ce.Data != "" {
+		ce.Checksum = ce.md5()
+	}
+	return ce.Checksum
+}
+
+func (ce *CacheEntryEntity) md5() string {
+	md5 := md5.Sum([]byte(ce.Data)) //nolint - MD5 is used for change detection, not for encryption
```
crabtree, line 37: Maybe you can come up with different name than md5 for that variable? It covers the md5 import.

pkg/config/entity.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	if time, ok := value.(time.Time); ok {
+		return time, nil
+	}
+	return nil, fmt.Errorf("Failed to convert value '%s' (kind: %s) for field 'Created' to Time struct",
+		value, reflect.TypeOf(value).Kind())
+}
+
+func requireValidBucketName(value interface{}) (interface{}, error) {
+	bucketName := fmt.Sprintf("%s", value)
```
crabtree, line 30: Is the empty string a valid bucket name?

pkg/db/marshaller.go
```diff
@@ -0,0 +1,120 @@
...
+	var err error
+	switch field.Kind() {
+	case reflect.Int64:
+		int64Value, ok := value.(int64)
+		if !ok {
+			return es.fireCastError(reflect.Int64, field, value)
+		}
+		err = field.Set(int64Value)
+	case reflect.Bool:
+		//some DBs handle booleans als integer values (0/1)
```
crabtree, line 94:
```suggestion
		//some DBs handle booleans as integer values (0/1)
```

Makefile
```diff
@@ -21,15 +21,25 @@ build: build-linux build-darwin build-linux-arm
...
+	@rm cover.out
+
+.PHONY: clean
+clean:
+	rm -rf bin
 
 .PHONY: all
-all: resolve build lint 
\ No newline at end of file
+all: resolve build test lint
```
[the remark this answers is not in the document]
tobiscr: oh, right.. sorry C&P failure. Fix is coming...

cmd/config/config.go
```diff
@@ -0,0 +1,135 @@
...
+			SslMode:  viper.GetBool("configManagement.db.postgres.sslMode"),
+			Debug:    o.Verbose,
+		}, nil
+	default:
+		return nil, fmt.Errorf("Database driver '%s' not supported yet", dbDriver)
+	}
+}
+
+func fileExists(file string) bool {
+	_, err := os.Stat(file)
```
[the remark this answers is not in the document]
tobiscr: Very good point! Thanks for the hint.. will enhance the check to make sure a found inode won't be a dir.

pkg/config/bucketmerger_test.go
```diff
@@ -0,0 +1,134 @@
...
+			bucket = fmt.Sprintf("%s-%s", bucket, subBucket)
+		}
+		kvMap, err := loadYaml(t, fmt.Sprintf("%s.yaml", bucket))
+		if err == nil {
+			importKVMap(t, kvRepo, bucket, kvMap)
+		}
+	}
+}
+
+func fileExists(file string) bool {
```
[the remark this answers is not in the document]
tobiscr: Omg.. you are really good! :) Redundant code is removed.

pkg/config/cacheentrryentity.go
tobiscr: Definitely, var is now renamed!

pkg/config/entity.go
tobiscr: no it isn't, but the validation should only run when the bucket string is defined. Marshaller are also executed in cases when an empty entity is provided (which is valid in cases of a delete-query, like here: https://github.com/kyma-incubator/reconciler/pull/2/files#diff-dae7efcf1d65d1b2a95798b43d9e45f920597f0889bd70a1e68fdf5a2e944909R70)

The validation for non-empty strings happens already via the annotation (see https://github.com/kyma-incubator/reconciler/pull/2/files#diff-2a76185f573c76ec3b23aff0fe2057dae5d020b1866407aadec53fd4a0fbef91R14). I'm actually abusing here the marshaller to define another validation for the value. A marshaller should normally only convert data-types from Go-Types to a format the DB understands and only fail if the value is not convertible. Here I let it fail when the provided value is not fitting to my expectations.

crabtree (dismissed):

crabtree (dismissed):

docs/configuration-management.md
```diff
@@ -0,0 +1,110 @@
...
+      * Name of the bucket the value belongs to.
+      * Creation date of the entry.
+      * User who created the configuration value entry.
+    * When a configuration value entity is created, it will per-default be mapped to the latest version of a configuration key entity.
+    * Configuration value entities are immutable and versioned: a change of the value or any metadata leads to a new version of the configuration value entity.
+    * Before a configuration value entry can be created, the assigned validation logic (part of the the configuration key metadata) has to be executed. If the validation was successful (return value `true`), the entry gets persisted otherwise an error will be returned.
+* **Caching**
+  * A cache entry is the merge result of multiple buckets and specific to one particular cluster.
+  * It has to be possible to identify in which cache entry a specific configuration value was used.
+  * If a configuration value has changed (other attributes coupled to a configuration entry don't matter), any cache entries which were using this configuration value have to be invalidated.
```
NHingerl, line 33:
```suggestion
  * If a configuration value has changed (other attributes coupled to a configuration entry don't matter), any cache entries that were using this configuration value must be invalidated.
```

docs/configuration-management.md
```diff
@@ -0,0 +1,110 @@
...
+  * The invalidation of a cache entry leads to a reconciliation for the related clusters (scheduling of a reconciliation run).
+* **CLI**
+  * A CLI is used to access the configuration management
```
NHingerl, line 36:
```suggestion
  * A CLI is used to access the configuration management.
```

docs/configuration-management.md
```diff
@@ -0,0 +1,110 @@
...
+
+The data of the configuration management can be distinguished in two domains:
+
+* Configuration data
+
+  Any information related to the configuration entires (metadata per configuration key, different versions of configuration entries etc.).
+
+* Caching data
+
+  A cache entity contains the merged configuration for one cluster. It also trackes which configuration entry was used in which cache entity to identify outdated cluster configurations.
```
NHingerl, line 53:
```suggestion
  A cache entity contains the merged configuration for one cluster. It also tracks which configuration entry was used in which cache entity to identify outdated cluster configurations.
```

docs/configuration-management.md
```diff
@@ -0,0 +1,110 @@
...
+
+|Column|Description|Data Type|Primary Key|Example|
+|--|--|--|--|--|
+|key|The configuration key|string|Yes|`my.config.key`|
+|version|Incremental version number|Integer|Yes|`1`|
+|datatype|Data type of the stored information|String|No|`integer`|
+|encrypted|Flag whether the value is plain text or encrypted|Boolean|No|`false`|
+|created|Timestamp when the entry was created|Integer|No|`123456789`|
+|user|User who created the entry|String|No|`i98765`|
+|validator|Optional logic which is executed to validated the value. Return value has to be a boolean: `true`=valid / `false`=invalid|String|No|`it >= 1 && it < 10`
```
NHingerl, line 72:
```suggestion
|validator|Optional logic that is executed to validate the value. Return value must be a boolean: `true`=valid / `false`=invalid|String|No|`it >= 1 && it < 10`
```

docs/configuration-management.md
```diff
@@ -0,0 +1,110 @@
...
+|trigger|Optional trigger functions which are executed by the reconciler|String|No|`restartPod("app=appLabel,component=abc")`
```
NHingerl, line 73:
```suggestion
|trigger|Optional trigger functions that are executed by the reconciler|String|No|`restartPod("app=appLabel,component=abc")`
```

docs/configuration-management.md
```diff
@@ -0,0 +1,110 @@
...
+|bucket|The bucket the configuration entry belongs to|String|Yes|`cust1`|
+|value|The value of the configuration entry|String|No|`123`|
+|created|Timestamp when the entry was created|Integer|No|`123456789`|
+|user|User who created the entry|String|No|`i98765`|
+
+#### Cache Table
+
+Requirements for the data structure layout:
+
+* Cache bucket merge result per cluster.
```
NHingerl, line 91: Why is this one not a sentence! All the others are full sentences! :O

tobiscr: Hm, good point. Will fix that when I'm updating the tech-docs in general. This file contains currently only some basic information.

alexandra-simeonova (approved):

crabtree (approved):

PR #25, opened 2021-06-23
Implement cluster-inventory

rJankowski93:


cmd/example/register-cluster_complicated-exemple.json
```diff
@@ -0,0 +1,280 @@
...
+    "labels": {
+      "scenarios": [
+        "DEFAULT"
+      ]
+    }
+  },
+  "kymaConfig": {
+    "version": "1.19.0",
+    "profile": "Production",
+    "conflictStrategy": "Replace",
```
piotrmiskiewicz, line 14: do we need conflict strategy? I think we will drop such logic.

cmd/example/register-cluster_complicated-exemple.json
```diff
@@ -0,0 +1,280 @@
...
+      "test2@test.pl"
+    ]
+  },
+  "metdata": {
+    "runtimeID": "6f8ec1bc-f300-4f00-bb65-6fec5dcd0ba5",
+    "globalAccountID": "3e64ebae-38b5-46a0-b1ed-9ccee153a0ae",
+    "subAccountID": "f7d129fa-b2fe-11eb-8529-0242ac130003",
+    "serviceID": "47c9dcbf-ff30-448e-ab36-d3bad66ba281",
+    "servicePlanID": "4deee563-e5ec-4731-b9b1-53b42d855f0c",
+    "shootName": "4deee563"
```
piotrmiskiewicz, line 277: I would add instanceID. Usually runtimeID is enough - unique RuntimeID -> unique instance ID. The relation is 1 to 1. Biut in the scenario with suspension one instance could have more than one runtimes (one runtime at the same time, but many runtimes in the history). Such instance ID could makes some operations easier in the future

cmd/example/register-cluster_complicated-exemple.json
```diff
@@ -0,0 +1,280 @@
...
+          }
+        ]
+      }
+    ],
+    "administrators": [
+      "test2@test.pl"
+    ]
+  },
+  "metdata": {
+    "runtimeID": "6f8ec1bc-f300-4f00-bb65-6fec5dcd0ba5",
```
piotrmiskiewicz, line 272: runtimeID is a significant identifier, wich identifies the cluster, let's move iot to the root

cmd/example/register-cluster_complicated-exemple.json
rJankowski93: Done

cmd/example/register-cluster_complicated-exemple.json
rJankowski93: Done

cmd/example/register-cluster_complicated-exemple.json
rJankowski93: I moved this to the runtimeInput, is it ok?

rJankowski93: Done

configs/db/postgres/000001_configuration-management.up.sql
```diff
@@ -53,12 +53,63 @@ CREATE TABLE config_cachedeps (
...
+    "sub_account_id"    text,
+    "service_id"       text,
+    "service_plan_id"   text,
+    "shoot_name"       text,
+    "instance_id"      text,
+	"created" TIMESTAMP WITHOUT TIME ZONE DEFAULT (NOW() AT TIME ZONE 'utc')
+);
+
+CREATE TABLE configurations (
+     id uuid PRIMARY KEY CHECK (id <> '00000000-0000-0000-0000-000000000000'),
```
tobiscr, line 72: As the configuration will be immutable, it could be helpful to use incremental IDs instead of UUIDs. They would have the benefit, that they implicitly reflect the timely order of confiurations.

configs/db/postgres/000001_configuration-management.up.sql
```diff
@@ -53,12 +53,63 @@ CREATE TABLE config_cachedeps (
...
+    "cluster_id" uuid NOT NULL,
+    "kyma_version" text,
+    "kyma_profile" text,
+    "created" TIMESTAMP WITHOUT TIME ZONE DEFAULT (NOW() AT TIME ZONE 'utc'),
+    CONSTRAINT fk_clusters FOREIGN KEY(cluster_id) REFERENCES clusters(id) ON DELETE CASCADE
+);
+
+CREATE TABLE components (
+      "id" uuid PRIMARY KEY CHECK (id <> '00000000-0000-0000-0000-000000000000'),
```
tobiscr, line 81: Just an idea to make it maybe simpler:we could store the received JSON as string in the configuration entity and parse it on-the-fly when we need it. This would have the benefit, that a could store the contract-version between KEB/Reconciler per configuration-entity which was valid at the time as we received the data. By knowing the contract-version, we can implement contact-specific JSON-hanlder which evaluate the JSON depending on the contract-version.

If we change the contact later, we would only have to add another JSON-renderer version to our application but  we won't have to modify the database structure too.

pkg/db/columnhandler.go
```diff
@@ -147,6 +147,39 @@ func (ch *ColumnHandler) ColumnNamesCsv(onlyWriteable bool) string {
 	return buffer.String()
 }
 
+// TODO
```
tobiscr, line 150: If we go with immutable objects (which would be good to have automatically also the history and auditing-problem addressed) we don't need updates actually.

pkg/model/clusteradministratorentity.go
```diff
@@ -0,0 +1,54 @@
+package model
+
```
tobiscr, line 2: Maybe we can handle it in the same way that we just store the JSON and don't use dedicated entities. But let's discuss that what we need and what wold be simpler for us.

suleymanakbas91 (approved):

PR #27, opened 2021-06-27
Add workspace factory

tobiscr:


pkg/test/expensive.go
```diff
@@ -0,0 +1,26 @@
+package test
+
+import (
+	"os"
+	"strings"
+)
+
+const (
+	EnvExpensiveTests = "RECONCILER_EXPENSIVE_TESTS"
```
rJankowski93, line 9: Can we add information about it to readme?

NHingerl (dismissed):

pkg/workspace/test/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
suleymanakbas91, line 2: Do we need this file here or was it committed by mistake?

pkg/git/clone.go
```diff
@@ -0,0 +1,58 @@
...
+var defaultCloner repoCloner = &remoteRepoCloner{}
+
+// CloneRepo clones the repository in the given URL to the given dstPath and checks out the given revision.
+// revision can be 'main', a release version (e.g. 1.4.1), a commit hash (e.g. 34edf09a).
+func CloneRepo(url, dstPath, rev string) error {
+	repo, err := defaultCloner.Clone(url, dstPath, true)
+	if err != nil {
+		return errors.Wrapf(err, "Error downloading repository (%s)", url)
+	}
+	if rev != "" {
```
suleymanakbas91, line 20: I think setting rev to a default value like `main` or returning an error in case rev is empty would be better than swallowing the problem and just returning nil.

suleymanakbas91 (changes requested):

NHingerl (dismissed):

rJankowski93 (dismissed):

NHingerl (approved):

suleymanakbas91 (approved):

PR #28, opened 2021-06-28
Add reconciler service cmd

tobiscr:
Add support to start reconciler service into CLI


NHingerl (dismissed):

NHingerl (approved):

cmd/service/start/options.go
```diff
@@ -0,0 +1,48 @@
...
+		_, err = tls.X509KeyPair(crt, key)
+		if err != nil {
+			return errors.Wrap(err,
+				fmt.Sprintf("Provided TLS certificate '%s' and key '%s' is invalid", o.SSLCrt, o.SSLCrt))
+		}
+	}
+	return nil
+}
+
+func (o *Options) SslSupport() bool {
```
suleymanakbas91, line 46: To make it consistent with the rest, can we rename this function to SSLSupport?

cmd/main.go
```diff
@@ -1,15 +1,124 @@
...
+
+func initViper(o *cli.Options) func() {
+	return func() {
+		//read configuration from ENV vars
+		viper.SetEnvPrefix(envVarPrefix)
+		viper.AutomaticEnv()
+
+		//read configuration from config file
+		cfgFile := getConfigFile()
+		if !file.Exists(cfgFile) {
```
suleymanakbas91, line 82: Would it make sense to move this check also to getConfigFile so that the both checks for the existence of file together?

suleymanakbas91 (changes requested):

cmd/main.go
tobiscr: aaah.. right. Makes sense. Will change it

cmd/service/start/options.go
tobiscr: right, will fix it

suleymanakbas91 (approved):

Issue #18, opened 2021-06-17
Implement metrics exporter


tobiscr:
**Description**

The metrics exporter is responsible to expose to external monitoring systems (e.g. Prometheus from SREs) the list of clusters which are currently in a failure, error or transition state.

These data will be evaluated by the SRE monitoring system and used to reduce false-positive alerts during reconciliation runs. SRE will disable the alerting for a particular time range when a cluster enters a transition state.

Output of the metrics exporter are Prometheus metrics, containing all clusters which are in a transition or error/failure state inclusive the date when the transition state started.

ACs:
* Prometheus metrics has to be exposed over an HTTP socket
* Exposed data have to include all clusters which are in transition/error/failure state inclusive the timestamp when this state started

**Reasons**

Reduce false-positive alerts on SRE side for clusters which are currently reconciled.

**Attachments**

suleymanakbas91:
Metrics endpoint is exposed, and it emits go metrics by default. On top, these three custom metrics have been implemented: 

```
reconciler_reconciliation_status{"cluster_id", "runtime_id", "cluster_version", "configuration_version"} - the status of the reconciliation
reconciler_reconciliation_not_ready_total - total number of clusters currently reconciling or in error/failed state
reconciler_reconciliation_waiting_total - total number of clusters waiting to be reconciled
```

PR #29, opened 2021-06-30
Add the first custom metrics

suleymanakbas91:
See https://github.com/kyma-incubator/reconciler/issues/18


pkg/metrics/reconciliation.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+func (c *ReconciliationWaitingCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.waitingClustersDesc
+}
+
+// Collect implements the prometheus.Collector interface.
+func (c *ReconciliationWaitingCollector) Collect(ch chan<- prometheus.Metric) {
+	clusters, err := c.inventory.ClustersToReconcile()
```
tobiscr, line 37: Arg.. I think we have to add another method the inventory. This will return only the clusters which have to be scheduled for reconciliation.

What we need is a list of all clusters which are currently reconciling or in an error / failed state.

Let's quickly sync tomorrow, but I think only the method-name has to be adjusted in your code. The rest should still work as before.

pkg/metrics/reconciliation_status.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/kyma-incubator/reconciler/pkg/cluster"
+	"github.com/kyma-incubator/reconciler/pkg/model"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	prometheusSubsystem = "reconciler"
+
+	statusError            float64 = 0
```
tobiscr, line 14: Maybe we can integrate this mapping into the status-struct we are using in our model. If we change laters our statuses, we would not have to update the metrics as well.

pkg/metrics/reconciliation_status.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (c *ReconciliationStatusCollector) OnClusterStateUpdate(state *cluster.State) error {
+	status, err := state.Status.GetClusterStatus()
+	if err != nil {
+		return err
+	}
+
+	c.reconciliationStatusGauge.
+		WithLabelValues(state.Cluster.Cluster, state.Cluster.Runtime, fmt.Sprintf("%d", state.Cluster.Version), fmt.Sprintf("%d", state.Configuration.Version)).
+		Set(status.ID)
```
tobiscr, line 53: Shouldn't it be `status.GetClusterStatus()` ?

tobiscr (changes requested):

pkg/metrics/reconciliation_status.go
```diff
@@ -0,0 +1,56 @@
...
+
+const (
+	prometheusSubsystem = "reconciler"
+)
+
+// ReconciliationStatusCollector provides the following metrics:
+// - reconciler_reconciliation_status{"cluster_id", "runtime_id", "global_account_id", "configuration_id"}
+// These gauges show the status of the reconciliation.
+// The value of the gauge could be:
+// 0 - Error
```
tobiscr, line 18: Not sure if we still need this comment after we moved it to the cluster-status object.

tobiscr (changes requested):

pkg/metrics/reconciliation_status.go
suleymanakbas91: Check line 46

pkg/metrics/reconciliation_status.go
suleymanakbas91: You're right, we don't need this here.

tobiscr (approved):

pkg/metrics/reconciliation_status.go
tobiscr: sorry, overlooked it

PR #39, opened 2021-07-11
Add chart provider

tobiscr:


pkg/chart/provider.go
```diff
@@ -0,0 +1,171 @@
...
+					fmt.Sprintf("Updating namespace of component '%s' with namespace provided by KEB: '%s' => '%s'",
+						clComp.Name, clComp.Namespace, comp.namespace))
+				compList.Components[idx].Namespace = comp.namespace
+			}
+		}
+	}
+	return compList, nil
+}
+
+func (p *Provider) comps(comps []*Component) map[string]*Component {
```
suleymanakbas91, line 136: The name of the function can be more explanatory of what it is doing.

tobiscr: Very good point... will change it to `componentMap`

suleymanakbas91 (approved):

suleymanakbas91 (approved):

PR #66, opened 2021-07-30
Interceptor - add managed-by label

rJankowski93:


pkg/reconciler/model.go
```diff
@@ -20,6 +20,11 @@ const (
 	Success    Status = "success"
 )
 
+const (
+	ManagedByLabel       = "managed-by"
+	LabelReconcilerValue = "reconciler"
```
tobiscr, line 25: to make these labels unique to Kyma, it would be good to add a prefix like `reconciler.kyma-project.org/managed-by` or similar to them (do we have a common prefix we normally use in Kyma)?

tobiscr (changes requested):

pkg/reconciler/model.go
rJankowski93: e.g. for serverless `serverless.kyma-project.io/managed-by` so changed this to `"reconciler.kyma-project.io/managed-by"`

tobiscr (approved):

PR #76, opened 2021-08-05
Add Busola migrator patch job

dbadura:
Rewrite the virtual service migrator script for busola to `Go`.

Issue: https://github.com/kyma-project/kyma/issues/11795


pkg/reconciler/instances/busola/test_files/virtual-service.yaml
```diff
@@ -0,0 +1,94 @@
...
+    "labels": {
+      "app": "dex",
+      "app.kubernetes.io/managed-by": "Helm",
+      "chart": "dex-1.0.0"
+    },
+    "managedFields": [
+      {
+        "apiVersion": "networking.istio.io/v1alpha3",
+        "fieldsType": "FieldsV1",
+        "fieldsV1": {
```
dbadura, line 20: I think that I should remove all those empty objects

pkg/reconciler/kubernetes/adapter/adapter.go
```diff
@@ -151,6 +151,7 @@ func (g *kubeClientAdapter) deployManifest(ctx context.Context, manifest, namesp
 			resource, err := g.kubeClient.ApplyWithNamespaceOverride(&unstruct, namespace)
 			if err != nil {
 				g.logger.Errorf("Failed to apply Kubernetes unstructured entity: %s", err)
+				g.logger.Debugf("Used JSON data: %+v", unstruct)
```
dbadura, line 154: To remove

pkg/reconciler/instances/busola/action.go
```diff
@@ -0,0 +1,112 @@
...
+	clientSet, err := helper.KubeClient.Clientset()
+	if err != nil {
+		return errors.Wrapf(err, "while getting client set from kubeclient")
+	}
+	restClient := clientSet.Discovery().RESTClient()
+
+	logger.Infof("Launching pre install busola migrator job, version: %s ", version)
+	for _, virtSvcToPatch := range p.virtSvcsToPatch {
+		logger.Infof("Patching virtual service: %s in namespace: %s", virtSvcToPatch.Name, virtSvcToPatch.Namespace)
+		if err := p.patchVirtSvc(ctx, restClient, virtSvcToPatch.Name, virtSvcToPatch.Namespace, logger); err != nil {
```
tgorgol, line 54: What will happen if the virtual service doesn't exist?

pkg/reconciler/instances/busola/busola.go
```diff
@@ -0,0 +1,46 @@
...
+
+	//configure reconciler
+	reconciler.
+		//list dependencies (these components have to be available before this component reconciler is able to run)
+		WithDependencies("istio", "dex", "console").
+		//register reconciler pre-action (executed BEFORE reconciliation happens)
+		WithPreReconcileAction(&VirtualServicePreInstallPatch{
+			name:            "pre-action",
+			virtSvcsToPatch: virtSvcs,
+			suffix:          "-old",
```
tgorgol, line 43: These values used to be configured via helm values, do we want to hardcode them?

tgorgol (commented):
Few comments from me

pkg/reconciler/instances/loader.go
```diff
@@ -5,6 +5,8 @@ package instances
 import (
 	//import required to register component reconciler 'base' in reconciler registry
 	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/base"
+	//import required to register component reconciler 'busola' in reconciler registry
+	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/busola"
```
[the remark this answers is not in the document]
tobiscr: The package name is not restricted or has to follow a naming convention. You are free to choose any name you like.

pkg/reconciler/instances/busola/vs_client.go
```diff
@@ -0,0 +1,74 @@
...
+func NewVirtSvcClient() VirtSvcClient {
+	return &client{}
+}
+
+const (
+	istioCRPath  = "/apis/networking.istio.io/v1alpha3"
+	resourceName = "virtualservices"
+)
+
+type virtSvc struct {
```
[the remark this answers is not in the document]
tobiscr: Please have in mind, that adding further APIs increases also the surface for potential security issues. You are of course not limited to use only APIs we have in place and it's ok to add furhter APIs. But please have in the amount of additional dependencies in mind.. the ISTIO client has a lot of them: https://github.com/istio/client-go/blob/master/go.sum

pkg/reconciler/instances/loader.go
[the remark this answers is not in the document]
dbadura: I renamed the pacakge to `busolamigrator`

pkg/reconciler/instances/busola/busola.go
```diff
@@ -0,0 +1,46 @@
...
+		{
+			Name:      "dex-virtualservice",
+			Namespace: "kyma-system",
+		},
+	}
+
+	//configure reconciler
+	reconciler.
+		//list dependencies (these components have to be available before this component reconciler is able to run)
+		WithDependencies("istio", "dex", "console").
```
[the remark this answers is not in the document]
dbadura: As `Virtual Service` is need, we only need to check for `istio` dependency. 
I removed the old ones.

pkg/reconciler/instances/busola/action.go
dbadura: There will be an error and the step will fail.
I added test case for this.

pkg/reconciler/instances/loader.go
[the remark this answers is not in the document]
tobiscr: 👍

tobiscr (dismissed):

pkg/reconciler/instances/busola/action.go
tgorgol: Then it behaves differently than the current helm hook implementation and having it work without dex/console components installed was one of the requirements if I remember correctly.

dbadura: You're right. I am going to fix it

pkg/reconciler/instances/busola/busola.go
dbadura: Are we going to persinalize it?
I think no.

pkg/reconciler/instances/busola/vs_client.go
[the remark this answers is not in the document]
dbadura: Busola migrator is simple job, so we can go with rest client.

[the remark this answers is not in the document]
tobiscr: 👍

tgorgol (approved):
LGTM

tobiscr (approved):

Issue #44, opened 2021-07-14
Kubeconfig has to be passed from KEB-contract model to reconciler-DB entity


tobiscr:
**Description**

The kubeconfig is currently not passed to the reconciler from KEB and missing in the contract.

The contract has to be adjuted and the kubeconfig-value has to be considered when creating new cluster entities in the reconciler.

See `inventory.createCluster` function for further details.

**Expected result**

Kubeconfig is another attribute of a `ClusterEntity` and stored in the DB.

**Actual result**

Kubeconfig is missing in models and in DB.

**Steps to reproduce**

**Troubleshooting**

tobiscr:
Please review TODO comment in https://github.com/kyma-incubator/reconciler/pull/78/files#diff-f921d302f6db810d79d7007b3c8a95dc107a6a7e4da95f92f3a96cbb6f6589aaL88 when working on this ticket

suleymanakbas91:
Implemented in #83.

PR #78, opened 2021-08-08
Stabilize reconciler e2e tests

tobiscr:


pkg/cluster/inventory.go
```diff
@@ -85,7 +85,7 @@ func (i *DefaultInventory) createCluster(contractVersion int64, cluster *keb.Clu
 		Cluster:    cluster.Cluster,
 		Runtime:    string(runtime),
 		Metadata:   string(metadata),
-		Kubeconfig: "fixme", //TODO: use correct model field as soon as kubeconfig is provided
+		Kubeconfig: cluster.Kubeconfig, //TODO: use correct model field as soon as kubeconfig is provided
```
suleymanakbas91, line 88: Do we still need the TODO comment?

tobiscr: Currently, we haven't yet agreed with Gophers where the kubeconfig will be located in the JSON. I was just adding it now to the top level to get the end2end scenario working.

We have also issue #44 open for aligning it - I would leave it for now and drop it as part of this issue (will add a comment in the issue ticket to drop this TODO)

suleymanakbas91 (approved):

PR #80, opened 2021-08-09
Create component-reconcilers chart

suleymanakbas91:


Makefile
```diff
@@ -1,6 +1,7 @@
 APP_NAME = reconciler
 IMG_NAME := $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME)
 TAG := $(DOCKER_TAG)
+COMPONENTS := $(shell (cd pkg/reconciler/instances && printf -- '%s' */ | tr '/' ',' | sed -e "s/^example,//" -e "s/,$$//" && cd ../../..))
```
tobiscr, line 4: wow, that looks quite hard to read :) not sure if the following command is simpler to read, but what about:

`find pkg/reconciler/instances -d 1 -type d -not -path '*/example' -exec sh -c 'echo $(basename {})' \;`

resources/reconciler/charts/component-reconcilers/values.yaml
```diff
@@ -0,0 +1,11 @@
+# Default values for component-reconcilers.
+
+nameOverride: ""
+fullnameOverride: ""
+
+components:
+- istio
```
tobiscr, line 7: Maybe better to fail if the list is empty? Otherwise we should to update this list when ading/removing new components to stay in-sync

tobiscr (changes requested):

Makefile
tobiscr: This works better  - without line breaks in the output:

`find pkg/reconciler/instances -d 1 -type d -not -path '*/example' -execdir echo -n '{} ' \;`

tobiscr (approved):

PR #89, opened 2021-08-11
Extract reconciler invocation

skhalash:
# What has been done?

1. Extract remote invocation of component reconcilers into a reconcile invoker. Hid it behind an interface to replaceable in the case of local reconciliation (will be used by the CLI). Split `WorkerFactory` into `local` and `remote`. They provide the corresponding `Invoker` implementation.
2. Implement a local scheduler, which installs Kyma on a single provided cluster.
3. Cosmetic renaming: `Fct` -> `Func`


crabtree (approved):

JeremyHarisch (commented):
Looks good 👍🏻

Issue #91, opened 2021-08-12
Align naming of operation-/correlation-ID and review cluster-status updates


tobiscr:
**Description**

This can be achieved by replacing the ticker-channel with a channel which is used by the operation registry to send status-update to the worker. The mothership-reconciler should track the cluster status on base of these conditions:

* If all component-reconcilers are reporting a `RECONCILING` status: cluster state becomes `RECONCILING`
* If one ore more component-reconcilers are reporting an `ERROR` status: cluster state becomes `ERROR`
* If all component-reconcilers are reporting a `SUCCESS` status: cluster state becomes `READY`

(Also documented in the [Wiki](https://cxwiki.sap.com/display/ysf/Reconciler+Architecture#ReconcilerArchitecture-Statesreflectedindatabase))

Additionally, the naming of the shared processing ID between mothership-reconciler and component-reconciler should be aligned. In the mothership-reconciler is the variable normally called `operationID` but on component-reconciler side we use the naming `correlationID`. This should be aligned to one name to make the code base more consistent.

ACs:
* Establish consistent naming for operationID/correlationID in mothership- and component-reconciler
* Verify that the cluster status is updated accordingly to the listed conditions above

**Reasons**

Reduce risk of lost data which can happen caused by time-based updates and make code base easier readable by using consistent naming.

**Attachments**

PR #106, opened 2021-08-23
Review cluster-status updates

suleymanakbas91:
- Review cluster status updates as described in #91 
- Rename leftover operationIDs as correlationIDs


pkg/scheduler/scheduler.go
```diff
@@ -32,19 +32,21 @@ type Scheduler interface {
...
 	inventoryWatch InventoryWatcher
 	workerFactory  WorkerFactory
 	mothershipCfg  reconciler.MothershipReconcilerConfig
 	poolSize       int
 	logger         *zap.SugaredLogger
 }
 
-func NewRemoteScheduler(inventoryWatch InventoryWatcher, workerFactory WorkerFactory, mothershipCfg reconciler.MothershipReconcilerConfig, workers int, debug bool) (Scheduler, error) {
+func NewRemoteScheduler(inventory cluster.Inventory, inventoryWatch InventoryWatcher, workerFactory WorkerFactory, mothershipCfg reconciler.MothershipReconcilerConfig, workers int, debug bool) (Scheduler, error) {
 	l, err := logger.NewLogger(false)
```
tobiscr, line 44: We could add a `NewWatcher()` method to the inventory interface and move the creation o the watcher into the inventory implementation. This would help to simplify this method signature and also bundle the creation of the `InventoryWatcher` with the underneath used inventory implementation (e.g. database-lookup based, or in case of an in-memory inventory it could be callback based) .. what do you think?

pkg/scheduler/scheduler.go
```diff
@@ -112,31 +114,55 @@ func (rs *RemoteScheduler) schedule(state cluster.State) {
...
+	_, err = rs.inventory.UpdateStatus(&state, model.Reconciling)
+	if err != nil {
+		rs.logger.Infof("Failed to update cluster status as reconciling: %s", err)
+	}
+
 	//Reconcile CRD components first
 	for _, component := range components {
 		if rs.isCRDComponent(component.Component) {
-			rs.reconcile(component, state, schedulingID, doInstallCRD, concurrencyNotAllowed)
+			rs.reconcile(nil, component, state, schedulingID, doInstallCRD, concurrencyNotAllowed)
```
tobiscr, line 126: Maybe we can rely on the Waiting-Group parameter to decide whether concurrency should be used, e.g. 

      func useConcurreny(wg *sync.WaitingGroup) bool {
          return wg != nil
      }

pkg/scheduler/scheduler.go
```diff
@@ -112,31 +114,55 @@ func (rs *RemoteScheduler) schedule(state cluster.State) {
...
+			rs.logger.Infof("Failed to update cluster status as ready: %s", err)
+		}
 	}
 }
 
-func (rs *RemoteScheduler) reconcile(component *keb.Components, state cluster.State, schedulingID string, installCRD bool, concurrent concurrency) {
-	fn := func(component *keb.Components, state cluster.State, schedulingID string) {
+func (rs *RemoteScheduler) reconcile(wg *sync.WaitGroup, component *keb.Components, state cluster.State, schedulingID string, installCRD bool, concurrent concurrency) {
+	fn := func(wg *sync.WaitGroup, component *keb.Components, state cluster.State, schedulingID string, concurrent concurrency) {
+		if bool(concurrent) {
```
tobiscr, line 163: as above.. maybe we use the `wg` as indicator for concurrency..

e.g. 

     func useConcurreny(wg *sync.WaitingGroup) bool {
         return wg != nil
     }

tobiscr (changes requested):

pkg/scheduler/scheduler.go
tobiscr: Other option could be to adda a getter for the inventory to the `InventoryWatcher`, e.g.

     func (w *DefaultInventoryWatcher) Inventory() *cluster.Inventory {
          return w.inventory
     }

pkg/scheduler/worker.go
```diff
@@ -121,10 +112,29 @@ func (w *Worker) process(component *keb.Components, state cluster.State, schedul
 			return false, err
 		}
 		return false, nil
-	case StateNew, StateInProgress, StateFailed:
+	case StateNew, StateInProgress:
+		// Operation still being processed by the component reconciler
```
tobiscr, line 116: Currently we change the status in two different places... failures are handled in the workers and successful-runs are handled in the scheduler.

Could be worth to move the status-update logic into one place. One option could be to implement a small and simple state-machine which will be updated by the workers (e.g. using a channel) and the state-machine decides which cluster-state should be persisted in the database.

pkg/scheduler/statusupdater.go
```diff
@@ -0,0 +1,131 @@
...
+	_, err := su.inventory.UpdateStatus(&su.clusterState, status)
+	if err != nil {
+		su.logger.Infof("Failed to update cluster status as ready: %s", err)
+	}
+}
+
+func (su *ClusterStatusUpdater) statusChangeAllowed(status model.Status) error {
+	latestState, err := su.inventory.GetLatest(su.clusterState.Cluster.Cluster)
+	if err != nil {
+		su.logger.Infof("Failed to update cluster status as ready: %s", err)
```
tobiscr, line 106: Error message doesn't fit to this step

pkg/scheduler/statusupdater.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (su *ClusterStatusUpdater) Run() {
+	timeout := time.After(progressTimeout)
+	for {
+		select {
+		case update := <-su.updateChannel:
+			su.statusMap[update.component] = update.operationState
+			if update.operationState == StateInProgress {
+				su.reconciling()
```
tobiscr, line 49: To save DB requests, it would be better to change the reconciling state of the cluster only at the beginning to `reconcile`. 

In this loop it would be enough to check only for DONE or ERROR updates.

pkg/scheduler/worker.go
```diff
@@ -72,18 +66,14 @@ func (w *Worker) Reconcile(component *keb.Components, state cluster.State, sched
...
-				}
 				return nil
 			}
 		}
 	}
 }
 
 func (w *Worker) process(component *keb.Components, state cluster.State, schedulingID string, installCRD bool) (bool, error) {
-	w.logger.Debugf("Processing the reconciliation for a component %s, correlationID: %s", component.Component, w.correlationID)
+	w.logger.Debugf("Processing the reconciliation for a compoent %s, correlationID: %s", component.Component, w.correlationID)
```
tobiscr: little typo in `component`

tobiscr (changes requested):

pkg/scheduler/remotescheduler.go
```diff
@@ -106,17 +107,20 @@ func (rs *RemoteScheduler) schedule(state cluster.State) {
 		return
 	}
 
+	statusUpdater := NewClusterStatusUpdater(rs.inventoryWatch.Inventory(), state, components, rs.logger)
+	go statusUpdater.Run()
```
tobiscr, line 111: Oh, sorry missed that one: we should also pass the CTX to the status-updater and check the `ctx.Done()` channel in the loop for a graceful shutdown of the goroutine

tobiscr (changes requested):

tobiscr (approved):

Issue #75, opened 2021-08-05
Implement end2end test for reconciler (includes mothership- and component-reconcilers)


tobiscr:
**Description**

To establish a reliable delivery process, a full end-to-end test for the interaction between:

1.  KEB to the mothership-reconciler 
2. mothership-reconciler to component-reconcilers 

The used mothership- and component-reconcilers run as dedicated services and communicate via REST with each other. The deployment of the mothership- and component-reconcilers has to be based on a Kubernetes deployment (comparable to the deployment used for the productive setup in KCP).

AC:

* End-to-end tests are added as dedicated pipeline for the reconciler in PROW
* The end-end-test simulates the reconciliation of a cluster. A dedicated Kubernetes cluster has to be created which is used b the component-reconcilers to reconcile Kyma (e.g. a new created Gardener cluster or a new K3d/s cluster).
* The end-to-end test covers the reconciliation of a Kyma installation (same components as used in SKR), considering several expected us-cases (ordered by their priority):
     1. Installation of the previous Kyma release on a new cluster (no Kyma installed yet)
     2. Upgrade of Kyma on an existing cluster using the latest Kyma release
     3. Reconciliation of a cluster with a manipulated Kyma installation (e.g. a Kyma component was deleted)
     4. Reconciliation of a cluster with a manipulated Kyma installation which cannot be recovered (e..g component was changed in a way that a reconciliation won't fix it)
     5. Reconciliation of a cluster which is not reachable (K8s down / access to API blocked)
     6. Reconciliation of a cluster with insufficient permissions (e.g. K8s-user has no permissions on `kyma-system` namespace)
     7. Reconciliation of a cluster with a defective Kyma component (e.g. syntax error in HELM template)
     8. ... further scenarios missing?...
* The end-to-end test covers the communication between the mothership and component-reconcilers  considering several expected use-cases (ordered by their priority):
    1. Happy path: mothership calls component reconcilers and gets successful result
    2. At least one component reconcilers returns consistently an error
    3. One or more component reconcilers are not reachable via network (down)
    4. A component reconcilers returns an invalid response (e.g. 500 server error)
* The end-to-end test covers the communication between the KEB and mothershop-reconciler considering several expected use-cases (ordered by their priority):
    1.  Happy path: KEB registers or upgrades a cluster
    2. KEB registers a cluster which already exists
    3. KEB sends an invalid request
    4. KEB requests status of a non-existing cluster

**Reasons**

[...]

PR #114, opened 2021-08-26
Add e2e test job

sayanh:
Related issue:
https://github.com/kyma-incubator/reconciler/issues/75


Makefile
```diff
@@ -1,7 +1,8 @@
 APP_NAME = reconciler
 IMG_NAME := $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(APP_NAME)
 TAG := $(DOCKER_TAG)
-COMPONENTS := $(shell recons=$$(find pkg/reconciler/instances -d 1 -type d -not -path '*/example' -execdir echo -n '{} ' \; | xargs) && echo $${recons// /,})
+# COMPONENTS := $(shell recons=$$(find pkg/reconciler/instances -d 1 -type d -not -path '*/example' -execdir echo -n '{} ' \; | xargs) && echo $${recons// /,})
```
mfaizanse, line 4: Reason for changing this was that the `find` command was failing, because `find` has different standard in GNU shell (in debian, prow-job pod). Also, the substitution for `recons` in echo command was also failing in GNU.

The error was:
```
find: paths must precede expression: `1'
/bin/sh: 1: Bad substitution
```

suleymanakbas91 (approved):

suleymanakbas91 (approved):

PR #122, opened 2021-08-31
Introduce values flag for CLI

pbochynski:


cmd/local/options.go
```diff
@@ -22,15 +22,22 @@ type Options struct {
...
 	profile        string
 	components     []string
+	values         []string
 }
 
-func defaultComponentList() []keb.Components {
-	defaultComponents := []string{"cluster-essentials", "istio-configuration@istio-system",
+func defaultComponents() []string {
+	return []string{"cluster-essentials", "istio-configuration@istio-system",
 		"certificates@istio-system", "logging", "tracing", "kiali", "monitoring", "eventing", "ory", "api-gateway",
```
tobiscr, line 30: Can'w we generate this list from this URL: https://github.com/kyma-project/kyma/blob/main/installation/resources/components.yaml ??

The list of existing components should come from Kyma and not be defined in the reconciler.

cmd/local/options.go
```diff
@@ -22,15 +22,22 @@ type Options struct {
...
-	return componentsFromStrings(defaultComponents)
-
+}
+func defaultValues() []string {
+	return []string{
+		"tracing.authProxy.config.useDex=false",
+		"tracing.authProxy.configDocsLink=https://kyma-project.io/docs",
+		"kiali.authProxy.config.useDex=false",
+		"kiali.authProxy.configDocsLink=https://kyma-project.io/docs",
+	}
```
tobiscr, line 40: Such information should go into the particular component and not become part of the reconciler.

cmd/local/options.go
```diff
@@ -55,16 +63,31 @@ func componentsFromStrings(list []string) []keb.Components {
 		if len(s) >= 2 {
 			namespace = s[1]
 		}
-		components = append(components, keb.Components{Component: name, Namespace: namespace})
+		configuration := []keb.Configuration{}
+		for _, keyValue := range values {
+			splitKeyValue := strings.Split(keyValue, "=")
```
tobiscr, line 68: Risky if the string includes a `=` in the value. Option to solve this could be to use a properties-reader for it (e.g. https://pkg.go.dev/github.com/magiconair/properties)

tobiscr (changes requested):

cmd/local/options.go
tobiscr: Code for reading the component list is available here: https://github.com/kyma-incubator/reconciler/pull/123/files#diff-ca4a1604ae2c727c0927d025dc68f9e56e274b5dbbd55ad6d66c45bebaf15bc8

cmd/local/options.go
pbochynski: Fixed.

cmd/local/options.go
pbochynski: Removed.

cmd/local/options.go
pbochynski: Fixed. Default is taken from workspace. You can override it with your own file using `--components-file` option.

cmd/local/options.go
```diff
@@ -40,13 +32,47 @@ func NewOptions(o *cli.Options) *Options {
...
+		[]string{}, // values
+		"",         // componentsFile
 	}
 }
 func (o *Options) Kubeconfig() string {
 	return o.kubeconfig
 }
 
-func componentsFromStrings(list []string) []keb.Components {
+type KymaComponentList struct {
```
tobiscr, line 43: You can skip this block (line 43-62) and use `data := test.NewComponentList(path)` instead

tobiscr (changes requested):

cmd/local/options.go
pbochynski: `NewComponentList(t *testing.T, compListFile string)` has testing dependency. I just reused structs and did Unmarshall call directly.

tobiscr (approved):

PR #123, opened 2021-08-31
Test rendering of all Kyma components

tobiscr:


pkg/test/componentlist.go
```diff
@@ -0,0 +1,55 @@
...
+	kymaNamespace        = "kyma-system"
+)
+
+type KymaComponentList struct {
+	DefaultNamespace string `yaml:"defaultNamespace" json:"defaultNamespace"`
+	Prerequisites    []Component
+	Components       []Component
+}
+
+type ComponentList struct {
```
pbochynski, line 22: Is it used somewhere?

pkg/test/componentlist.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+	"io"
+	"net/http"
+	"testing"
+)
+
+const (
+	kymaComponentListURL = "https://raw.githubusercontent.com/kyma-project/kyma/main/installation/resources/components.yaml"
```
pbochynski, line 12: Why not get the file from `filepath.Join(workspace.InstallationResourceDir, "components.yaml")`?

tobiscr: Would be an option, but this would require to clone the whole repository at the beginning of the unittest to finally use just one file of it. From resource perspective, it's cheaper to just pick the required file.

tobiscr: But right - as the unit-tests which are using this feature are anyway cloning the repository up-front, it doens't make sense to download it here again.

Will change it to accept a path to the file as input-parameter.

pkg/test/componentlist.go
tobiscr: Right, thanks for pointing it out! That's dead code...will drop it

crabtree (approved):

PR #125, opened 2021-09-01
Private Registries Support

ralikio:
WIP Private Registries Support


pkg/reconciler/instances/connectivityproxy/connectivityproxy.go
```diff
@@ -0,0 +1,33 @@
+package connectivityproxy
```
tobiscr, line 1: JFYI - as we discussed in our previous call, I created a bug for the issue with hyphens in pkg name: https://github.com/kyma-incubator/reconciler/issues/126

pkg/reconciler/kubernetes/kubeclient/kubeclient.go
```diff
@@ -30,12 +29,25 @@ type KubeClient struct {
 	mapper        *restmapper.DeferredDiscoveryRESTMapper
 }
 
+func NewInCluster() (*KubeClient, error) {
+	config, err := rest.InClusterConfig()
```
tobiscr, line 33: The name is a big tricky to understand without looking into the code, can we make it a bit more self-explaining (e.g. `NewInClusterConfig` or similar)?

pkg/reconciler/kubernetes/kubeclient/kubeclient.go
```diff
@@ -82,11 +94,11 @@ func (kube *KubeClient) ApplyWithNamespaceOverride(u *unstructured.Unstructured,
 	helper := resource.NewHelper(restClient, restMapping)
 
 	if namespaceOverride == "" {
-		SetDefaultNamespaceIfScopedAndNoneSet(u, helper)
-	} else {
-		SetNamespaceIfScoped(namespaceOverride, u, helper)
+		namespaceOverride = "default"
```
tobiscr, line 97: like it! 👍

pkg/reconciler/model.go
```diff
@@ -37,11 +42,20 @@ type Reconciliation struct {
 	CallbackURL     string          `json:"callbackURL"` //CallbackURL is mandatory when component-reconciler runs in separate process
 	InstallCRD      bool            `json:"installCRD"`
 	CorrelationID   string          `json:"correlationID"`
+	Repo            Repo            `json:"repo"`
```
tobiscr, line 45: Can we use "repository" in the JSON as the other fields were also avoiding abbreviations?

pkg/reconciler/model.go
```diff
@@ -112,3 +126,54 @@ type MothershipReconcilerConfig struct {
...
+func (r *Repo) ReadToken(clientSet core.CoreV1Interface, namespace string) error {
+	secretKey, err := MapSecretKey(r.URL)
+	if err != nil {
+		return err
+	}
+
+	secret, err := clientSet.
+		Secrets(namespace).
+		Get(context.Background(), secretKey, v1.GetOptions{})
+
```
tobiscr, line 144: QQ - this secret would be configured by SREs? If yes -we have to verify with them how we deal with it during the deployment.

Alternatively providing it as part of the comonent-configuration won't be an option, right (means, the credentials would be part of the component-configuration and the component-reconciler would create the secret in K8s if it's missing)?

Just thinking how we can avoid a coupling of our software to data in the underneath cluster. Also testing will become more difficult for us.

tobiscr (changes requested):

pkg/reconciler/git/secret_copy.go
```diff
@@ -0,0 +1,138 @@
+package git
```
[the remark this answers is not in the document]
ralikio: Moved to connectivityproxy package.

pkg/reconciler/kubernetes/kubeclient/kubeclient.go
ralikio: I have changed it to `NewInClusterClient`

pkg/reconciler/git/secret_copy.go
```diff
@@ -0,0 +1,138 @@
...
+		return err
+	}
+	err = c.SecretTo.createSecret(secret)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+type SecretTo struct {
```
rafalpotempa, line 34: I'd move it above `SecretCopy`

pkg/reconciler/git/secret_copy.go
```diff
@@ -0,0 +1,138 @@
...
+		SecretTo: SecretTo{
+			Namespace:       configs[prefix+".secret.namespace"],
+			Name:            configs[prefix+".secret.name"],
+			Key:             configs[prefix+".secret.key"],
+			targetClientSet: targetClientSet,
+		},
+	}
+}
+
+func NewFromSecret(prefix string, targetClientSet k8s.Interface, inClusterClient k8s.Interface, configs map[string]string) *SecretCopy {
```
rafalpotempa, line 57: What if we make a single constructor:
`func NewSecretCopy(prefix string, targetClientSet k8s.Interface) *SecretCopy` 

and two methods:
`func (s *SecretCopy) FromURL(configs map[string]string)` and 
`func (s *SecretCopy) FromSecret(inClusterClient k8s.Interface, configs map[string]string)`.

So it can be chained `NewSecretCopy(...).FromURL(...)`.

Anyway I don't like the fact that `configs` is map[string]string, since you don't know what should be inside. It looks more like dynamic typesetting from Python or JS, rather than static typesetting in Go.

pkg/reconciler/instances/connectivityproxy/connectivityproxy.go
```diff
@@ -0,0 +1,33 @@
...
+		WithPreReconcileAction(&CustomAction{
+			name: "pre-action",
+		}).
+		WithReconcileAction(&CustomAction{
+			name: "install-action",
+		}).
+		WithPostReconcileAction(&CustomAction{
+			name: "post-action",
+		})
+}
```
rafalpotempa, line 33: Interesting way in how it is designed...

pkg/reconciler/model_test.go
```diff
@@ -0,0 +1,99 @@
...
+			URL:   "https://localhost",
+			Token: "",
+		}
+
+		err := repo.ReadToken(client.CoreV1(), "non-existing")
+
+		assert.Error(t, err)
+	})
+
+	t.Run("Should read correct token secret", func(t *testing.T) {
```
rafalpotempa, line 63: This test has the same description as the test in line 36

pkg/reconciler/service/test.go
```diff
@@ -2,6 +2,7 @@ package service
 
 import (
 	"context"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler"
```
rafalpotempa, line 5: You have some problem with formatting here... It should go below and join other `github.com` imports.

pkg/reconciler/service/test.go
```diff
@@ -20,7 +21,7 @@ func (c *cleanup) removeKymaComponent(t *testing.T, version, component, namespac
 	t.Logf("Cleanup of component '%s' (version: %s, namespace: %s) started", component, version, namespace)
 
 	//render manifest
-	chartProv, err := c.reconciler.newChartProvider()
+	chartProv, err := c.reconciler.newChartProvider(&reconciler.Repo{})
```
rafalpotempa, line 24: This looks weird. Just like you are passing an empty repository and than, the required information is appearing out of nowhere... What if the `newChartProvider` was left as was, but it had an additional method `WithRepo` or sth like that, instead?

e.g., `UsingRepo(*reconciler.Repo)`

tobiscr:
@ralikio : there is another PR coming which is also changing the GIT API. Do you think we can merge your PR soon? Otherwise, we should quickly aling the change from @jeremyharisch  (https://github.com/kyma-incubator/reconciler/pull/180) with your modifications to avoid bigger merge conflicts.

pkg/reconciler/service/action.go
```diff
@@ -10,11 +13,15 @@ import (
...
-	ChartProvider    *chart.Provider
+	KubeClient         kubernetes.Client
+	WorkspaceFactory   *workspace.Factory
+	Context            context.Context
+	Logger             *zap.SugaredLogger
+	ChartProvider      *chart.Provider
+	Kubeconfig         string
+	InClusterClientSet k8s.Interface
+	ConfigsMap         map[string]interface{}
+	ClientSet          k8s.Interface
```
tobiscr, line 24: This is not needed. Please use `context.KubeClient.ClientSet()`

pkg/reconciler/service/runner.go
```diff
@@ -80,22 +81,46 @@ func (r *runner) reconcile(ctx context.Context, model *reconciler.Reconciliation
...
+	if err != nil {
+		return err
+	}
+
+	configs := model.ConfigsToMap()
+	err = model.Repository.ReadToken(inClusterClientSet.CoreV1(), configs["repo.token.namespace"])
+	if err != nil {
+		return err
+	}
+
```
tobiscr, line 104: Is this really globally required? Can't we move this part into the `connectiviy-proxy` itself?

Also the fields in the `ActionContext` are only required by the connectivity-proxy. Do we really have a need to make them available to all component reconcilers?

rafalpotempa (approved):

pkg/reconciler/service/action.go
```diff
@@ -10,11 +13,15 @@ import (
...
-	Context          context.Context
-	Logger           *zap.SugaredLogger
-	ChartProvider    *chart.Provider
+	KubeClient         kubernetes.Client
+	WorkspaceFactory   *workspace.Factory
+	Context            context.Context
+	Logger             *zap.SugaredLogger
+	ChartProvider      *chart.Provider
+	Kubeconfig         string
+	InClusterClientSet k8s.Interface
```
tobiscr, line 22: Isn't this part specific to the connectivity proxy? Not sure if we really need it also in the `ActionContext`to make it accessible for all other component reconcilers.

pkg/reconciler/service/action.go
```diff
@@ -10,11 +13,15 @@ import (
...
+	ConfigsMap         map[string]interface{}
```
tobiscr, line 23: This should not go into the ActionContext as it's component specific. The context is intended to give component reconcilers access to cross-cutting features of the reconciler.

pkg/reconciler/service/runner.go
```diff
@@ -80,22 +81,46 @@ func (r *runner) reconcile(ctx context.Context, model *reconciler.Reconciliation
...
-		Logger:           r.logger,
-		ChartProvider:    chartProvider,
+		InClusterClientSet: inClusterClientSet,
+		KubeClient:         kubeClient,
+		ClientSet:          clusterClientSet,
+		WorkspaceFactory:   wsFactory,
+		Context:            ctx,
+		Logger:             r.logger,
+		ChartProvider:      chartProvider,
+		ConfigsMap:         configs,
```
tobiscr, line 123: This should not go into the `ActionContext` as it's purpose is to give access to non-component specific functionality (cross-cutting features).

I would suggest to move the `ConfigToMap()` function from the model into the connectivity-proxy (currently only used there) OR  we pass this map directly to the Actions (I can image this makes life in general easier for our colleagues).

     type Action interface {
	Run(version, profile string, configuration map[string]interface{}, helper *ActionContext) error
     }

tobiscr (changes requested):

tobiscr (approved):
As agreed, the comments will be addresses in a following PR afterwards

pkg/reconciler/git/client.go
```diff
@@ -0,0 +1,31 @@
...
+	"context"
+	"fmt"
+
+	"github.com/go-git/go-git/v5"
+	gitp "github.com/go-git/go-git/v5/plumbing"
+	"github.com/pkg/errors"
+)
+
+type Client struct {
+	repo *git.Repository
```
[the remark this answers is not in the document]
ralikio: To be handled on separate branch.

pkg/reconciler/git/client.go
```diff
@@ -0,0 +1,31 @@
...
+	var err error
+	d.repo, err = git.PlainCloneContext(ctx, path, isBare, o)
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("Failed to clone the repository: '%s'", err))
+	}
+	return d.repo, nil
+}
+
+func (d *Client) Worktree() (*git.Worktree, error) {
+	return d.repo.Worktree()
```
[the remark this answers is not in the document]
ralikio: To be handled on separate branch.

pkg/reconciler/service/action.go
ralikio: To be handled on separate branch.

pkg/reconciler/service/runner.go
ralikio: To be handled on separate branch.

pkg/reconciler/service/action.go
ralikio: To be handled on separate branch.

pkg/reconciler/service/action.go
ralikio: To be handled on separate branch.

pkg/reconciler/service/runner.go
ralikio: To be handled on separate branch.

tobiscr (dismissed):
Approved, as agreed with @ralikio we will address the comments in a second PR afterwards

pkg/reconciler/service/action.go
ralikio: Refactored.

pkg/reconciler/service/runner.go
ralikio: Refactored.

pkg/reconciler/service/action.go
ralikio: Refactored.

pkg/reconciler/service/runner.go
ralikio: Refactored.

pkg/reconciler/service/action.go
ralikio: Refactored.

tobiscr (approved):

suleymanakbas91 (approved):

suleymanakbas91 (approved):

PR #132, opened 2021-09-02
Reconciler int tests

tobiscr:


pkg/scheduler/operations.go
```diff
@@ -35,6 +35,28 @@ type OperationsRegistry interface {
 	SetFailed(correlationID, schedulingID, reason string) error
 }
 
+type OperationNowFoundError struct {
```
suleymanakbas91, line 38: Should be `OperationNotFoundError` :)

suleymanakbas91 (changes requested):

pkg/scheduler/operations.go
tobiscr: haha...  but Now would actually also be right! :D Let me fix this typo

suleymanakbas91 (approved):

PR #134, opened 2021-09-03
Downgraded spf13/cobra

skhalash:
Downgraded spf13/cobra to the version used by the cli to avoid dependency problems.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.7.1
 	github.com/spf13/cast v1.4.1 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.7.1
 	github.com/spf13/cast v1.4.1 // indirect
-	github.com/spf13/cobra v1.2.1
```
tobiscr, line 26: Would be better to upgrade the consumer of this API - our build mechanism is normally upgrading dependencies automatically when new versions are available.

Can't promise that this version will stay long :/

tobiscr (approved):

PR #135, opened 2021-09-03
Persisted operations impl

crabtree:
Persist operations data in postgres database


tobiscr (approved):

pkg/scheduler/operations.go
```diff
@@ -57,36 +44,198 @@ func IsOperationNotFoundError(err error) bool {
...
+
+		q, err := db.NewQuery(or.Conn, &model.OperationEntity{})
+		if err != nil {
+			return nil, err
+		}
+		whereCond := map[string]interface{}{
+			"CorrelationID": correlationID,
+			"SchedulingID":  schedulingID,
+		}
+		_, err = q.Delete().
```
tobiscr, line 133: do you think we should have a kind of ops-archive to have the reconciliation history available for a longe period?

I could images SREs could have a use-case for reviewing the last reconciliation protocl at a later time, what do you think?

tobiscr (approved):

PR #137, opened 2021-09-03
Downgraded spf13/viper

skhalash:
Downgraded spf13/viper to the version used by the cli to avoid dependency problems. As soon as we have our changes ready, we will bump the versions of spf13/cobra and spf13/viper in the cli as well as the reconciler.


go.mod
```diff
@@ -13,22 +13,30 @@ require (
 	github.com/iancoleman/strcase v0.1.3
 	github.com/imdario/mergo v0.3.12
 	github.com/jonboulle/clockwork v0.1.0
+	github.com/json-iterator/go v1.1.11 // indirect
```
tobiscr, line 16: Can't this be done by using `replace` in the CLI go.mod ?

tobiscr: https://golang.org/ref/mod#go-mod-file-replace

tobiscr (changes requested):

go.mod
tobiscr: This comment is for all downgrades intended .. don'T want to repeat it several times :)

tobiscr (changes requested):

skhalash:
Good idea, @tobiscr. I will replace the dependency on the cli side.

PR #144, opened 2021-09-07
Fix scaffolding script

suleymanakbas91:


pkg/reconciler/instances/reconcilerctl.sh
```diff
@@ -57,12 +57,12 @@ function addReconciler {
 
   mv "./${pkgName}/example.go" "./${pkgName}/${pkgName}.go"
 
-  echo "Adjusting init function for component '${reconName}' (OS is '$(uname)')"
+  echo "Adjusting init function for component '${pkgName}' (OS is '$(uname)')"
```
tobiscr, line 60: Good idea to handle it in that way! The only disadvantage we are now running into, that we cannot support components which include dashes in their name (e.g. `istio-configuration`).

What about this solution:

1. We adjust the `example.go` file and use as component name `example-component`.
2. We use two SED-calls for the replacement:

       sed -i '' "s/example/${pkgName}/g" "$file"
       sed -i '' "s/example-component/${reconName}/g" "$file"

tobiscr (changes requested):

pkg/reconciler/instances/reconcilerctl.sh
tobiscr: As agreed in our call, we should support dashes in component names and will use a little helper script in our build process to get these component names via Golang.

For the fix of this shell-script, I think we can go with distinguishing between package-name and comp-name.

tobiscr (approved):

PR #151, opened 2021-09-08
Rename keb.components to keb.component

tobiscr:
Changing plural name to singular for component entity.


pkg/keb/contract.go
```diff
@@ -19,17 +19,17 @@ type Configuration struct {
 	Secret bool        `json:"secret"`
 }
 
-type Components struct {
+type Component struct {
 	Component     string          `json:"component"`
```
suleymanakbas91, line 23: Does it make sense to rename this as `Name`? What do you think?

suleymanakbas91: Then it'd be `Component.Name` instead of `Component.Component`.

tobiscr: Agree, that would be better.. will rename it

PR #157, opened 2021-09-10
Improve Istio Reconciler Installation scenario

mjakobczyk:
**Description**:
- wrap `istioctl` command
- establish interfaces
- introduce mocks
- provide tests


pkg/reconciler/instances/istio/actions/actions.go
```diff
@@ -0,0 +1,146 @@
...
+	kubeConfig    string
+	istioOperator string
+	kubeClient    kubernetes.Client
+	logger        *zap.SugaredLogger
+}
+
+// NewDefaultIstioPerformer creates a new instance of the DefaultIstioPerformer.
+func NewDefaultIstioPerformer(kubeConfig, manifest string, kubeClient kubernetes.Client, logger *zap.SugaredLogger, cmder istioctl.Commander) (*DefaultIstioPerformer, error) {
+	istioctlPath := resolveIstioctlPath()
+	if istioctlPath == "" {
```
tobiscr, line 55: would recommend to let the `resolveIstioctlPath` throw the error instead checking for an empty string outside of it

      istioctlPath, err := resolveIstioctlPath()
      if err != nil {
         return err
      }

pkg/reconciler/instances/istio/actions/actions.go
```diff
@@ -0,0 +1,146 @@
...
+	err = c.kubeClient.PatchUsingStrategy("MutatingWebhookConfiguration", "istio-sidecar-injector", "istio-system", patchContentJSON, types.JSONPatchType)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func resolveIstioctlPath() string {
+	return os.Getenv(istioctlBinaryPathEnvKey)
```
tobiscr, line 134: For local testig, it would be good to have an option to use also a locally installed istioctl command as fallback, like

    func resolveIstioctlPath (string, err) {
       istioctlPath, ok := os.Lookup(istioctlBinaryPathEnvKey)
       if ok {
         if file.Exists(istioctlPath) {
           return istioctlPath, nil
         }
         return "", fmt. Errorf("istioctl binary '%s' defined in env-var '%s' but file does not exist", istioctlPath, istioctlBinaryPathEnvKey)
       }
       istioctlPath, err := exec.Command("which istioctl").Output()
       if err != nil {
           return "", fmt.Errorf("istiotctl binary not defined in env-var '%' and not available in $PATH", istioctlBinaryPathEnvKey)
       }
       return istioctlPath, nil
    }

pkg/reconciler/instances/istio/actions/actions.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func extractIstioOperatorContextFrom(manifest string) string {
+	defs := strings.Split(manifest, "---")
```
tobiscr, line 138: oh, be careful with string operations as they can lead to bigger memory allocations and also don't deal with edge cases well (e.g. `-- #some comment here )

To process manifests, you can use a feature of the `kubeclient`API, see https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/chart/provider.go#L62-L78

pkg/reconciler/instances/istio/istioctl/commander.go
```diff
@@ -0,0 +1,29 @@
...
+// DefaultCommander provides a default implementation of Commander.
+type DefaultCommander struct{}
+
+func (c *DefaultCommander) Install(istioCtlPath, istioOperatorPath, kubeconfigPath string) error {
+	cmd := exec.Command(istioCtlPath, "apply", "-f", istioOperatorPath, "--kubeconfig", kubeconfigPath, "--skip-confirmation")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
+		return err
```
tobiscr, line 25: Maybe worth to add also the STDERR + STDOUT output to the error (I haven't checked what the error object already includes, maybe it's already done like this by Golang)

pkg/reconciler/instances/istio/actions/actions.go
```diff
@@ -0,0 +1,146 @@
...
+// NewDefaultIstioPerformer creates a new instance of the DefaultIstioPerformer.
+func NewDefaultIstioPerformer(kubeConfig, manifest string, kubeClient kubernetes.Client, logger *zap.SugaredLogger, cmder istioctl.Commander) (*DefaultIstioPerformer, error) {
+	istioctlPath := resolveIstioctlPath()
+	if istioctlPath == "" {
+		return nil, errors.New("Istioctl binary could not be found under ISTIOCTL_PATH env variable")
+	}
+
+	istioOperator := extractIstioOperatorContextFrom(manifest)
+	if istioOperator == "" {
+		return nil, errors.New("Istio Operator definition could not be found in manifest")
```
tobiscr, line 61: Same here, better to return the err directly from the `extractIstioOperatorContextFrom` to have it properly encapsulated

tobiscr (dismissed):

pkg/reconciler/instances/istio/actions/actions.go
mjakobczyk: It would be cool to have such quality-of-life improvement, BUT:
- Windows has it's own equivalent to `which` command
- the binary does not have to be called exactly `istioctl` - it may contain prefixes or suffixes
- you may have multiple `istioctl` binaries available on your machine - which one should be chosen?
We might come back to the discussion how to handle these, but for now I think it's not crucial. As a user you simply have to set your env var on the local machine and this Reconciler will still work!

tobiscr (approved):

PR #160, opened 2021-09-10
Validate SQL queries

suleymanakbas91:


pkg/db/sqlite.go
```diff
@@ -55,13 +72,46 @@ func (sc *SqliteConnection) Query(query string, args ...interface{}) (DataRows,
...
+		}
+	}
 	result, err := sc.db.Exec(query, args...)
 	if err != nil {
 		sc.logger.Errorf("Sqlite3 Exec() error: %s", err)
 	}
 	return result, err
 }
 
+func (sc *SqliteConnection) validate(query string) bool {
```
tobiscr, line 88: From SQL perspective, the differences between SQLite and Postgres are (except timestamps and custom DB-functions) shoulnd't be that extreme. Would it be possible to centralize the SQL validation logic in a SQL-validator object and pass that object to the DB-connections?

By making the validator configurable to enable/disable the validation internally, e.g.

    type sqlValidator struct {
        blockQueries bool
    }
    
    func newValidator(blockQueries bool) *sqlValidator {
        return &sqlValidator{blockQueries}
    }
    
     func (s *sqlValidator) Validate(sql string) error {
         //validte SQL query here...
         msg := fmt.Sprintf("Found potential SQL injection: %s", sql)
         if s.blockQueries {
            return errors.New(msg)
        } 
        log.Warnf(msg)
        return nil
     }

pkg/db/postgres.go
```diff
@@ -53,13 +71,46 @@ func (pc *PostgresConnection) Query(query string, args ...interface{}) (DataRows
...
+		}
+	}
 	result, err := pc.db.Exec(query, args...)
 	if err != nil {
 		pc.logger.Errorf("Postgres Exec() error: %s", err)
 	}
 	return result, err
 }
 
+func (pc *PostgresConnection) validate(query string) bool {
```
tobiscr, line 87: Please see me comment below.. maybe we can centlraize this logical part in a dedicated validator object to encapsulate this part and make it easier testable.

tobiscr (changes requested):

pkg/db/factory.go
```diff
@@ -21,14 +22,15 @@ func NewConnectionFactory(configFile string, debug bool) (ConnectionFactory, err
 	}
 
 	dbToUse := viper.GetString("db.driver")
+	executeUnverified := viper.GetBool("db.executeUnverified")
```
tobiscr, line 25: We could use a kind of `sqlValidator` instance and pass this instance it the conn-factories (under the assumption, that it's stateless and can be used in parallel without issues).

tobiscr (approved):

Issue #35, opened 2021-07-06
Implement basic istioctl support in the component reconciler


varbanv:
**Description**

Provide a basic implementation of Istio configuration using istioctl + configuration template. 

**Reasons**

Since Istio is a special case for component installations right now, we'll attempt to reduce risk and dev time by implementing custom logic to install Istio via istioctl.

**Links**
https://github.com/kyma-project/kyma/issues/11635

tobiscr:
Obsolete after moving to components reconciler

varbanv:
We need to do the basic integration with istoctl and passing a config like we do with kubectl.

tobiscr:
Do we still have to handle that? Would be more a topic for goats team as part of their component reconciler.

tobiscr:
see PR #67

strekm:
TODO:
- scan istioctl 1.10.x & 1.9.x binaries
- figure out how to get access to rendered istiooperator manifest
- support for installation & upgrade
- apply istio proxy reset job after upgrade
- do NOT support istio delete
- trigger thread modelling with sec team

@tobiscr will we get already "helm template" istiooperator file? do we already have helm client available? do we already have some helper methods for that?

tobiscr:
@strekm : we would close this ticket as our part for supporting remote and embedded reconcilers is done. Please follow-up the mentioned TODOs in a dedicated ticket.

PR #171, opened 2021-09-15
Feature/istio reconciler upgrade

mjakobczyk:
**Description**:
- introduce Istio upgrade scenario

**Links**:
See: #35


pkg/reconciler/instances/istio/action.go
```diff
@@ -23,20 +40,174 @@ func (a *ReconcileAction) Run(version, profile string, config map[string]interfa
...
+		if err != nil {
+			return errors.Wrap(err, "Could not install Istio")
+		}
+
+		err = a.performer.PatchMutatingWebhook(context.KubeClient, context.Logger)
+		if err != nil {
+			return errors.Wrap(err, "Could not patch MutatingWebhookConfiguration")
+		}
+	} else {
+		if !isClientVersionAcceptable(ver) {
```
veichtj, line 63: Maybe you think about moving that several checks into a new function li,e "isUpgradePossible" to make it more expressive

pkg/reconciler/instances/istio/action.go
```diff
@@ -23,20 +40,174 @@ func (a *ReconcileAction) Run(version, profile string, config map[string]interfa
...
+	if shouldInstall(ver) {
+		context.Logger.Info("No Istio version was detected on the cluster, performing installation...")
+
+		err = a.performer.Install(context.KubeClient.Kubeconfig(), manifest.Manifest, context.Logger)
+		if err != nil {
+			return errors.Wrap(err, "Could not install Istio")
+		}
+
+		err = a.performer.PatchMutatingWebhook(context.KubeClient, context.Logger)
+		if err != nil {
```
veichtj, line 59: Not sure about the responsiblity of the "performer" but could be an idea to move the whole logic into the "Install" also the "PatchMutatingWebhook"

pkg/reconciler/instances/istio/istioctl/commander.go
```diff
@@ -3,21 +3,66 @@ package istioctl
...
 }
 
+var execCommand = exec.Command
+
 // DefaultCommander provides a default implementation of Commander.
 type DefaultCommander struct{}
 
-func (c *DefaultCommander) Install(istioCtlPath, istioOperatorPath, kubeconfigPath string) error {
-	cmd := exec.Command(istioCtlPath, "apply", "-f", istioOperatorPath, "--kubeconfig", kubeconfigPath, "--skip-confirmation")
+func (c *DefaultCommander) Install(istioOperator, kubeconfig string, logger *zap.SugaredLogger) error {
```
veichtj, line 35: The code of Install and Upgrade is pretty much the same, except the argument "upgrade" passed in this case. Maybe you can think about harmonizing this code into one new common method

pkg/reconciler/instances/istio/reset/proxy/proxy.go
```diff
@@ -0,0 +1,58 @@
+package proxy
```
veichtj, line 1: To have the same pattern, maybe you could think about having similar action like the performer of the istio install/upgrade

veichtj (commented):
As discussed,  my remarks. And some general thoughts / questions

- referring to the [istio upgrade docu](https://istio.io/latest/docs/setup/upgrade/in-place/), the recommended upgrade way is canary upgrade instead of in-place upgrade. There might be well-defined reasons why we are doing in-place, but please double check and document it shortly in the code

- for the in-place upgrade istio proposes three main steps, the precheck, the upgrade and restart of the manual update of the data plane. There is a proposed kubectl to do this, in the code we have there a lot of logic for handling this, are we sure that this is not anyway doing what we need? If so, please also add a short comment why this is needed at all.
Also might the precheck also be useful? 

- In the last IR it was mentioned that the istio reconciler needs to support two istioctl versions if I'm not wrong? Is that already in place or does it come later?

pkg/reconciler/instances/istio/action.go
```diff
@@ -23,20 +40,174 @@ func (a *ReconcileAction) Run(version, profile string, config map[string]interfa
 		return err
 	}
 
-	performer, err := actions.NewDefaultIstioPerformer(context.KubeClient.Kubeconfig(), manifest.Manifest, context.KubeClient, context.Logger, &istioctl.DefaultCommander{})
+	ver, err := a.performer.Version(context.WorkspaceFactory, version, istioChart, context.KubeClient.Kubeconfig(), context.Logger)
```
veichtj, line 43: I would propose to rename "Version" to something more meaningful like for e.g. "GetInstalledIstioVersion"

pkg/reconciler/instances/istio/action.go
```diff
@@ -23,20 +40,174 @@ func (a *ReconcileAction) Run(version, profile string, config map[string]interfa
...
+
+	return nil
+}
+
+// shouldInstall checks if istio is already installed
+func shouldInstall(version actions.IstioVersion) bool {
+	return version.DataPlaneVersion == "" && version.PilotVersion == ""
+}
+
+// isClientVersionAcceptable checks if istioctl version is up to date.
```
Tomasz-Smelcerz-SAP, line 119: I would rename it to "isClientVersionUpToDate"

pkg/reconciler/instances/istio/action.go
```diff
@@ -23,20 +40,174 @@ func (a *ReconcileAction) Run(version, profile string, config map[string]interfa
...
+			ImageVersion:          fmt.Sprintf("%s-distroless", ver.TargetVersion),
+			RetriesCount:          retriesCount,
+			DelayBetweenRetries:   delayBetweenRetries,
+			SleepAfterPodDeletion: sleepAfterPodDeletion,
+			Kubeclient:            kubeClient,
+			Debug:                 true,
+			Log:                   context.Logger,
+		}
+
+		err = a.istioProxyReset.Run(cfg)
```
Tomasz-Smelcerz-SAP, line 105: It would be nice to make it more readable, too many "if's". How about encapsulating this in a single method? It's reasonable to assume all theses checks/vlidations are always executed.

pkg/reconciler/instances/istio/action.go
```diff
@@ -9,11 +19,18 @@ import (
...
+	sleepAfterPodDeletion = 10
 )
 
 type ReconcileAction struct {
+	commander       istioctl.Commander
+	istioProxyReset proxy.IstioProxyReset
+	performer       actions.IstioPerformer
 }
 
 func (a *ReconcileAction) Run(version, profile string, config map[string]interface{}, context *service.ActionContext) error {
```
werdes72, line 36: This function is very long and quite hard to read, it could be split into few smaller ones.

werdes72, line 36: This function became very long, it could be separated into a few shorter functions.

Tomasz-Smelcerz-SAP (dismissed):

Tomasz-Smelcerz-SAP (approved):

tobiscr (approved):

PR #177, opened 2021-09-16
Separate CRD installation from components

suleymanakbas91:


pkg/reconciler/service/runner.go
```diff
@@ -132,7 +133,13 @@ func (r *runner) reconcile(ctx context.Context, model *reconciler.Reconciliation
 }
 
 func (r *runner) install(ctx context.Context, chartProvider *chart.Provider, model *reconciler.Reconciliation, kubeClient kubernetes.Client) error {
-	manifest, err := r.renderManifest(chartProvider, model)
+	var err error
+	var manifest string
+	if model.Component == "CRDs" {
```
tobiscr, line 138: As this name is a special, we should may be move it as a public connstant to our models ( `model.go` )

tobiscr (approved):

PR #180, opened 2021-09-16
Resolve PR, branch, etc. as installation source

jeremyharisch:
Regarding [this issue](https://github.com/kyma-project/kyma/issues/12037), we need to support PRs and branches as source of a kyma deploy. 
This, was missing until now in the reconciler.


pkg/reconciler/git/resolve.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+type refLister interface {
+	List(repoURL string) ([]*plumbing.Reference, error)
+}
+
+type remoteRefLister struct {
+}
+
+var defaultLister refLister = &remoteRefLister{}
```
tobiscr, line 22: We shouldn't use a variable to change the behaviour of unit tests.  Also the encapsulation isn't very good.

Would recommend to bundle the functions in two different implementations:

      type revisionResolver struct {
         repository *git.Repository
         url string
      }

     func (r *revisionResolver) Resolve(rev string) {...code from resolveRevision ...}

pkg/reconciler/git/resolve_test.go
```diff
@@ -0,0 +1,92 @@
...
+			givenRevision: "nonExistingBranch",
+			expectErr:     true,
+			kind:          "branch",
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.summary, func(t *testing.T) {
+			defaultLister = &fakeRefLister{
```
tobiscr, line 80: For the unit-test, pass the `fakeRefLister` to the resolver instance (please see pseudo-code above).

tobiscr (changes requested):

pkg/reconciler/git/clone.go
```diff
@@ -39,13 +38,13 @@ func (rc *remoteRepoCloner) Clone(url, path string, autoCheckout bool) (*git.Rep
...
 
-func checkout(repo *git.Repository, rev string) error {
+func checkout(repo *git.Repository, rev string, url string) error {
 	w, err := repo.Worktree()
 	if err != nil {
 		return errors.Wrap(err, "error getting the GIT worktree")
 	}
 
-	hash, err := repo.ResolveRevision(plumbing.Revision(rev))
+	hash, err := resolveRevision(repo, url, rev)
```
tobiscr, line 47: Use the `revisionResolver` instead calling a function.

tobiscr (changes requested):

tobiscr:
Adding hold-label as we agreed to merge first #125

tobiscr (approved):

PR #184, opened 2021-09-16
Add cmd flag to execute up migrations

crabtree:


tobiscr (approved):

tobiscr:
@crabtree: do you see any risk that we could run into situations where multiple MS-instances would trigger a migration in parallel (e.g. when multiple replicas are started at once)?

PR #199, opened 2021-09-21
Reconciler - Connectivity Bindings

ralikio:
WIP Bindings in Reconcilers


pkg/reconciler/kubernetes/client.go
```diff
@@ -29,4 +32,8 @@ type Client interface {
 	Delete(ctx context.Context, manifest, namespace string) ([]*Resource, error)
 	PatchUsingStrategy(kind, name, namespace string, p []byte, strategy types.PatchType) error
 	Clientset() (kubernetes.Interface, error)
+
+	GetStatefulSet(ctx context.Context, name, namespace string) (*v1apps.StatefulSet, error)
```
tobiscr, line 36: this is a good idea! Would be worth to create a follow-up ticket to add also support for a few more base-types, like

* DaemonSet
* Pod
* Map
* Deployment
* Job

pkg/reconciler/service/install.go
```diff
@@ -11,7 +11,12 @@ import (
 )
 
 type Install struct {
-	logger *zap.SugaredLogger
```
tobiscr, line 14: as it would lead to nil-pointers if the logger is not passed during initialisation, it's maybe better use a newInstall() fct and make the logger again private

tobiscr (changes requested):

pkg/reconciler/instances/connectivityproxy/loader_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+		result, err := loader.FindSecret(actionContext, &unstructured.Unstructured{Object: map[string]interface{}{
+			"spec": map[string]interface{}{
+				"secretName": secretName,
+			},
+		}})
+
+		require.NoError(t, err)
+		require.Equal(t, secret, result)
+	})
```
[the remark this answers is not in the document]
ralikio: I would recommend being explicit in the assertions to remove ambiguity in how contract for given component is understood.

pkg/reconciler/service/install.go
```diff
@@ -14,6 +14,15 @@ type Install struct {
 	logger *zap.SugaredLogger
 }
 
+func NewInstall(logger *zap.SugaredLogger) *Install {
+	return &Install{logger: logger}
```
rafalpotempa, line 18: What is happening here? 😅

rafalpotempa (dismissed):

pkg/reconciler/instances/connectivityproxy/command.go
```diff
@@ -0,0 +1,86 @@
...
+	targetClientSetFactory NewTargetClientSet
+	install                service.Operation
+	copyFactory            []CopyFactory
+}
+
+func (a *CommandActions) Install(context *service.ActionContext, bindingSecret *apiCoreV1.Secret) error {
+	for key, value := range bindingSecret.Data {
+		configuration := context.Model.Configuration
+		if configuration == nil {
+			context.Model.Configuration = make(map[string]interface{})
```
tobiscr, line 33: Please add this check and the creation of an empty map to the [invoker.go](https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/invoker/invoker.go#L25) - we should not delegate such checks to our comp-reconcilers

Issue #86, opened 2021-08-11
Delete namespaces which were managed by reconciler only if they do not include any further resources


tobiscr:
**Description**

When deleting resources based on a manifest, defined namespaces in the manifest should per default be excluded and processes in a second step:

After the deletion of all manifest resources is finished, the namespace deletion can happen **but only if no further resources from Kyma exist in the namespace**. If a namespace is not empty, the deletion is not allowed.

AC:*

* Resources created by the reconciler are removed (deletion should happen based on the resources defined in the component manifests)
* Reminiang resources have to be verified whether they were created by Kyma or by a customer. Only resources created by Kyma should be deleted.
* A namespace can only be removed if it is completely empty (not including any customer resources).

**Reasons**

Don't delete namespaces which include resources.

**Attachments**

PR #201, opened 2021-09-21
Delete namespaces which were managed by reconciler

rJankowski93:
https://github.com/kyma-incubator/reconciler/issues/86


pkg/reconciler/kubernetes/adapter/adapter.go
```diff
@@ -200,8 +200,11 @@ func (g *kubeClientAdapter) Delete(ctx context.Context, manifest, namespace stri
 		g.logger.Warnf("Watching progress of deleted resources failed: %s", err)
 	}
 
-	//TODO: check if namespace is empty and if yes: remove it
-
+	if err = g.kubeClient.DeleteNamespace(namespace); err != nil {
+		g.logger.Errorf("Failed to delete namespace name='%s': %s",
```
tobiscr, line 204: This is now deleting the namespace without checking for custom resources, right?

rJankowski93: in first version we check if namespace contains any resources. If any resources exist on namespace, we don't remove namespace.

tobiscr (approved):

PR #210, opened 2021-09-23
New scheduler architecture for mothership-reconciler

tobiscr:


pkg/scheduler/invoker/remoteinvoker.go
```diff
@@ -0,0 +1,186 @@
...
+
+func NewRemoteReoncilerInvoker(reconRepo reconciliation.Repository, cfg *config.Config, logger *zap.SugaredLogger) *RemoteReconcilerInvoker {
+	return &RemoteReconcilerInvoker{
+		reconRepo: reconRepo,
+		config:    cfg,
+		logger:    logger,
+	}
+}
+
+func (i *RemoteReconcilerInvoker) Invoke(_ context.Context, params *Params) error {
```
suleymanakbas91, line 36: Instead of just ignoring the context here, we can send the HTTP request with context. What do you think?

pkg/scheduler/invoker/remoteinvoker.go
```diff
@@ -0,0 +1,186 @@
...
+			i.logger.Warnf("Remote invoker failed to dump HTTP response from component reconciler: %s", err)
+		}
+	} else {
+		return resp, errors.Wrap(err, fmt.Sprintf("failed to call remote reconciler (URL: %s)", compRecon.URL))
+	}
+
+	return resp, nil
+}
+
+func (i *RemoteReconcilerInvoker) marshalHTTPResponse(body []byte, respModel interface{}, params *Params) error {
```
suleymanakbas91, line 157: Should be `unmarshalHTTPResponse` instead, right?

pkg/scheduler/invoker/remoteinvoker.go
tobiscr: aah, you mean that it's cancelling the http-call when context got closed?

pkg/scheduler/invoker/remoteinvoker.go
tobiscr: Right!  :)

tobiscr: fixed

pkg/scheduler/invoker/remoteinvoker.go
suleymanakbas91: yes

configs/reconciler.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 db:
-  driver: postgres
+  driver: sqlite
```
rJankowski93, line 3: In default case will we use `sqllite`?

rJankowski93 (commented):
What do you think about add image with new arch to readme?

configs/reconciler.yaml
tobiscr: at least for local development it's simpler to use. For CI pipelines we should think whether we stay with SQLite or run them with Postgres (or maybe both.. but depends on the execution time / resource consumption).

suleymanakbas91 (approved):

PR #219, opened 2021-09-27
Removing `oapi` from `all`

ralikio:
After model generation it is not necessary to run oapi goal during each and every build.


dbadura (changes requested):
This check is required to run on every pre-main job to ensure that nobody forgot about regenerating the models.

pbochynski:
I believe it is not going to be merged.

PR #220, opened 2021-09-28
Include component resources to be installed with chart

mjakobczyk:
**Description**:
- deploy `istio-configuration` k8s resources after performing Istio installation
- omit `IstioOperator` resource from manifest to avoid duplication

### DIFF
--- a/pkg/reconciler/instances/istio/action.go
+++ b/pkg/reconciler/instances/istio/action.go
@@ -1,7 +1,6 @@
 package istio
 
 import (
-	"fmt"
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/chart"
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/instances/istio/actions"
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/instances/istio/istioctl"
@@ -47,8 +46,11 @@ func (a *ReconcileAction) Run(context *service.ActionContext) error {
 		return errors.Wrap(err, "Could not patch MutatingWebhookConfiguration")
 	}
 
-	generated := generateNewManifestWithoutIstioOperatorFrom(manifest.Manifest)
-	fmt.Println(generated)
+	generated, err := generateNewManifestWithoutIstioOperatorFrom(manifest.Manifest)
+	if err != nil {
+		return errors.Wrap(err, "Could not generate manifest without Istio Operator")
+	}
+
 	_, err = context.KubeClient.Deploy(context.Context, generated, istioNamespace, nil)
 	if err != nil {
 		return errors.Wrap(err, "Could not deploy Istio resources")
@@ -57,10 +59,10 @@ func (a *ReconcileAction) Run(context *service.ActionContext) error {
 	return nil
 }
 
-func generateNewManifestWithoutIstioOperatorFrom(manifest string) string {
+func generateNewManifestWithoutIstioOperatorFrom(manifest string) (string, error) {
 	unstructs, err := kubeclient.ToUnstructured([]byte(manifest), true)
 	if err != nil {
-		return ""
+		return "", err
 	}
 
 	builder := strings.Builder{}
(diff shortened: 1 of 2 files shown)


pkg/reconciler/instances/istio/action.go
```diff
@@ -41,5 +42,31 @@ func (a *ReconcileAction) Run(version, profile string, config map[string]interfa
...
+	_, err = context.KubeClient.Deploy(context.Context, generateNewManifestWithoutIstioOperatorFrom(manifest.Manifest), istioNamespace, nil)
+	if err != nil {
+		return errors.Wrap(err, "Could not deploy Istio resources")
+	}
+
 	return nil
 }
+
+func generateNewManifestWithoutIstioOperatorFrom(manifest string) string {
+	separator := "---"
```
pbochynski, line 54: It smells. Can't you parse it (or use k8s client to do it) and remove IstioOperator resource from the list?

pbochynski (changes requested):
String manipulation is not the best idea for managing YAML files.

pkg/reconciler/instances/istio/action.go
tobiscr: You can use our K8s-API features to convert manifests into `[]unstructured.Unstructured` entities and operate on them... see https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/kubernetes/kubeclient/unstructured.go#L15

tobiscr (changes requested):

pkg/reconciler/instances/istio/action.go
```diff
@@ -41,5 +44,36 @@ func (a *ReconcileAction) Run(version, profile string, config map[string]interfa
 		return errors.Wrap(err, "Could not patch MutatingWebhookConfiguration")
 	}
 
+	generated := generateNewManifestWithoutIstioOperatorFrom(manifest.Manifest)
+	fmt.Println(generated)
```
tobiscr, line 48: Probably a left over? If not, please use the  provided `logger` instance from the `ActionContext`.

pkg/reconciler/instances/istio/action.go
```diff
@@ -41,5 +44,36 @@ func (a *ReconcileAction) Run(version, profile string, config map[string]interfa
...
+		if unstruct.GetKind() == "IstioOperator" {
+			continue
+		}
+
+		unstructBytes, err := unstruct.MarshalJSON()
+		if err != nil {
+			return ""
+		}
+
+		builder.WriteString("---\n")
```
tobiscr, line 74: Would recommend not to serialize each `Unstructured` one by one and add YAML delimiters (`---`) manually to it. Better to rely completely on the the K8s API to convert it back to YAML.

Therefore, an option could be to drop just the element from the slice of `Unstructured` entities. Like

      var withoutOperator []unstructured.Unstructured
      for _, unstr := range unstructs {
         if unstruct.GetKind() == "IstioOperator" {
            continue
         }
         withoutOperator = append(withoutOperator, unstr)
      }
      
Finally, serialise it, e.g. by using `unstructured.UnstructuredJSONScheme.Encode()` ([link to API](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/helpers.go#L347-L351))

tobiscr (changes requested):

pkg/reconciler/instances/istio/action.go
tobiscr: It's actually the opposite of this call: https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/kubernetes/kubeclient/unstructured.go#L91

tobiscr: If you like, feel free to add this new function for converting `unstructured.Unstructured` to our API https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/kubernetes/kubeclient/unstructured.go ;)

pbochynski (approved):

tobiscr (dismissed):

tobiscr (approved):

Issue #188, opened 2021-09-17
Allow SRE to administer cluster reconciliation status via CLI


kwiatekus:
**Description**

As an SRE operator I want to use KCP CLI to be able to 

 - disable cluster reconciliation
 - enable cluster reconciliation ( at its "normal" schedule )
 - ensure a cluster will be reconciled in the next cycle
 
 **Reasons**

Allow to control the cluster reconciliation for clusters operated by SRE engineers.

kwiatekus:
Those three use cases should be controlled by the [Clusterstatus](https://github.com/kyma-incubator/reconciler/blob/e1bfce4cc1e0738ee37b28eb301efd67ecd5997a/pkg/model/clusterstatus.go#L9)
 - `ready`- setting this status would mark the cluster as available for reconciliation according to a regular schedule. Will be picked up for reconciliation eventually
 - `reconcile_pending`- this status would mark the cluster for reconciliation within the next cycle ( "as soon as possible" )
 - `disabled` - new status to add. That would mean no reconciliation for this cluster

kwiatekus:
[Updating the status](https://github.com/kyma-incubator/reconciler/blob/main/pkg/cluster/inventory.go#L194-L205) is not yet exposed via http. Would be a part of the task

kwiatekus:
@pbochynski @ebensom @tobiscr @varbanv 
There is one thing to decide on:

**Should kcp CLI update the status at the inventory directly?**
 Or 
**Should it go via KEB?**

KEB is currently the only source of configuration.. and maybe its better to keep it this way.
Also KEB is facing BTP UI. 
If we ( in the future ) want to reflect the status of reconciliation in the BTP cockpit, than we probably cannot make any exceptions and allow bypassing KEB when changing the cluster state

varbanv:
KCP should update the inventory via REST API, no need to go via KEB.
This should not be changing any configuration, just triggering a state change so KEB being the source of truth doesn't change.
KEB is polling the cluster state when it requests changes, there is no reason why it can't poll again if BTP is to incorporate status dashboard. Besides, periodic reconciliation would change the state without notifying KEB, wouldn't it?

pbochynski:
I think these operations should NOT go through KEB.

piotrmiskiewicz:
What should be done with the provisioning proces, if someone disables reconciler when the provisioning (in KEB) is in progress? KEB could fail the process, still wait for reconciling (but finally it will be timed out) or something else?
[...]

PR #221, opened 2021-09-29
Add a new clusteerstatus - `reconcile_disabled`  

pPrecel:
**Description**

Add the `reconcile_disabled`  clusterstatus which should freez reconcilation of the targeted cluster.

Changes proposed:

- Add the `reconcile_disabled`  clusterstatus
- Upgrade the openAPI manifest
- Upgrade the openAPI README.md

**Issues**
https://github.com/kyma-incubator/reconciler/issues/188


pkg/cluster/inventory.go
```diff
@@ -402,7 +402,8 @@ func (i *DefaultInventory) ClustersToReconcile(reconcileInterval time.Duration)
 
 func (i *DefaultInventory) ClustersNotReady() ([]*State, error) {
 	statusFilter := &statusFilter{
-		allowedStatuses: []model.Status{model.ClusterStatusReconciling, model.ClusterStatusReconcileFailed, model.ClusterStatusError},
+		// TODO: ask about allowedStatuses
+		allowedStatuses: []model.Status{model.ClusterStatusReconciling, model.ClusterStatusReconcileFailed, model.ClusterStatusError, model.ClusterStatusReconcileDisabled},
```
kwiatekus, line 406: Where is this method used? If its used to pick up clusters for next cycle than probably we shouldn't be returning those for which reconciliation was disabled

pPrecel: This method is used only to filter clusters to expose the right Prometheus metrics. 
[here](https://github.com/kyma-incubator/reconciler/blob/874713c60a30b8a95870fdb4c9b460e6d667ca17/pkg/metrics/reconciliation_not_ready.go#L40)

PR #249, opened 2021-10-11
Add get list reconcilers

m00g3n:


cmd/mothership/mothership/start/http.go
```diff
@@ -224,6 +244,158 @@ func updateLatestCluster(o *Options, w http.ResponseWriter, r *http.Request) {
...
+			&reconciler.HTTPErrorResponse{Error: err.Error()},
+		)
+		return
+	}
+
+	results := []keb.Reconciliation{}
+
+RESULT_LOOP:
+	for _, reconcile := range reconciles {
+		// FIXME add new method in inventory to fetch multiple statuses via runtimeID and fetch it in 1 go
```
tobiscr, line 293: Agree... having something like `getLatest(runtimeID ...string)` would be nice

pkg/scheduler/reconciliation/filter.go
```diff
@@ -23,6 +26,43 @@ func (ws *WithSchedulingID) FilterByInstance(i *model.ReconciliationEntity) *mod
...
+	var values string
+	for i := range wc.RuntimeIDs {
+		if i == runtimeIDsLen-1 {
+			values = fmt.Sprintf("%s$%d", values, i+1)
+			break
+		}
+		values = fmt.Sprintf("%s$%d,", values, i+1)
+	}
+
+	q.WhereIn("RuntimeID", values, strings.Join(wc.RuntimeIDs, ","))
```
tobiscr, line 48: I mean it expected the values of the placeholders as array, concatenation isn't required actually

tobiscr (changes requested):

pkg/scheduler/reconciliation/filter.go
m00g3n: This is the tricky one.
The `WhereIn` method expects arguments as `args ... interface{}` and the `wc.RuntimeIDs` type is `[]string`.
[If we pass the string slice we will pass it as `interface{}`](https://play.golang.org/p/vQaKSXLPh03) (we cannot spread it because the types do not match `[]interface` vs `[]string`).

tobiscr: oh, that's quite hacky :D 

Can't we use something like this?

     q.WhereIn("RuntimeID", values, func (args []string) []interface{} {
        var result []interface{}
        for _, arg := range args {
           result = append(result, arg)
        }
        return result
     }(wc.RuntimeIDs)...)

m00g3n: Let's do the conversion. I will just create dedicated function to avoid confusion with immediate executed lambda.

Issue #255, opened 2021-10-12
Enhance error responses to KEB


wozniakjan:
**Description**

Reconciler currently provides very limited information regarding failures to KEB.

https://github.com/kyma-incubator/reconciler/blob/376a4ec6627c07e4f3398c95554fc2f5a804c384/pkg/keb/model_gen.go#L26-L32

It's nice to know that reconciliation failed but knowing what caused it would be very helpful not only for KEB to debug but also for the end-user. And seeing `"status": "error"` may not be enough.

It would be great if we can add one more field, e.g. `"details"`, describing the high level root cause of the error.

**Reasons**

This will provide better UX and allow easier debugging.

**Attachments**

originally described here: https://github.com/kyma-incubator/reconciler/issues/250#issuecomment-940052439, but it was discussed that it would be better as a standalone issue.

wozniakjan:
I had some availability yesterday so I tried to tackle it myself - https://github.com/kyma-incubator/reconciler/pull/256. Unfortunately didn't finish and I was assigned a different task at the moment.

PR #282, opened 2021-10-19
Enhance keb failure responses

rJankowski93:
https://github.com/kyma-incubator/reconciler/issues/255


openapi/external_api.yaml
```diff
@@ -254,6 +254,10 @@ components:
         configurationVersion:
           type: integer
           format: int64
+        failures:
+          type: array
+          items:
+            $ref: '#/components/schemas/failure'
```
wozniakjan, line 260: is it possible that there is going to be an error from a reconciler not related to any component? I think these might be also useful for KEB if they can occur.

rJankowski93: Hmmm...I think it's impossible, I think every error will be related with some component. 
@tobiscr what do you think?

cmd/mothership/mothership/start/http.go
```diff
@@ -621,16 +621,39 @@ func sendResponse(w http.ResponseWriter, r *http.Request, clusterState *cluster.
...
 
-func newClusterResponse(r *http.Request, clusterState *cluster.State) (*keb.HTTPClusterResponse, error) {
+func newClusterResponse(r *http.Request, clusterState *cluster.State, reconciliationRepository reconciliation.Repository) (*keb.HTTPClusterResponse, error) {
 	kebStatus, err := clusterState.Status.GetKEBClusterStatus()
 	if err != nil {
 		return nil, err
 	}
+
+	var failures []keb.Failure
+	reconciliations, err := reconciliationRepository.GetReconciliations(&reconciliation.WithRuntimeID{RuntimeID: clusterState.Cluster.RuntimeID})
```
tobiscr, line 631: The current implement is unfortunately not fully reliable to retrieve the correct reconciliation for a cluster.

Quick review of the data model:

`Cluster  ---has many-->  ClusterConfiguration ---has many--> ClusterConfigurationStatus`

A reconciliation is always happening on base of a particular `ClusterConfiguration` and not on a runtimeID. A reconciliation entity has following information we can use to query for it:

* reference to the [clusterConfig-ID](https://github.com/kyma-incubator/reconciler/blob/main/configs/db/postgres/000002_scheduler_operations.up.sql#L12)
* reference to the [clusterConfigStatus-ID](https://github.com/kyma-incubator/reconciler/blob/main/configs/db/postgres/000002_scheduler_operations.up.sql#L13)

I would recommend to query for the clusterConfigStatus-ID as it's  a precise match if you have a `cluster.State` instance available.

But this will require to implement quickly another [`Filter`](https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/reconciliation/filter.go) instance which allows querying a reconciliation based on the reference `clusterConfigStatus` ID.

cmd/mothership/mothership/start/http.go
```diff
@@ -621,16 +621,39 @@ func sendResponse(w http.ResponseWriter, r *http.Request, clusterState *cluster.
...
+		return nil, err
+	}
+	if len(reconciliations) > 0 {
+		operations, err := reconciliationRepository.GetOperations(reconciliations[0].SchedulingID)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, operation := range operations {
+			if operation.State.IsError() {
```
tobiscr, line 642: good idea, but the name `isError` would indicate that we check only for the `[OperationStateError`](https://github.com/kyma-incubator/reconciler/blob/main/pkg/model/operationstate.go#L15). As we are interested in all operations which had Go-errors during their execution, I could imagine that the name `hasErrors()` is a bit better.

openapi/external_api.yaml
tobiscr: yes that'S possible.. but this will return a error-response (4xx or 5xx response code) as these kinds of error are not cluster specific

pkg/model/operationstate.go
```diff
@@ -39,3 +39,7 @@ func NewOperationState(state string) (OperationState, error) {
 	}
 	return result, nil
 }
+
+func (o OperationState) IsError() bool {
```
tobiscr, line 43: See comment above.. name can be a bit misleading

cmd/mothership/mothership/start/http.go
```diff
@@ -621,16 +621,41 @@ func sendResponse(w http.ResponseWriter, r *http.Request, clusterState *cluster.
...
 }
 
-func newClusterResponse(r *http.Request, clusterState *cluster.State) (*keb.HTTPClusterResponse, error) {
+func newClusterResponse(r *http.Request, clusterState *cluster.State, reconciliationRepository reconciliation.Repository) (*keb.HTTPClusterResponse, error) {
 	kebStatus, err := clusterState.Status.GetKEBClusterStatus()
 	if err != nil {
 		return nil, err
 	}
+
+	var failures []keb.Failure
```
tobiscr, line 630:
```suggestion
	var failures []*keb.Failure
```

cmd/mothership/mothership/start/http.go
```diff
@@ -621,16 +621,41 @@ func sendResponse(w http.ResponseWriter, r *http.Request, clusterState *cluster.
...
+		}
+		if len(reconciliations) > 0 {
+			operations, err := reconciliationRepository.GetOperations(reconciliations[0].SchedulingID)
+			if err != nil {
+				return nil, err
+			}
+
+			for _, operation := range operations {
+				if operation.State.IsError() {
+					failures = append(failures, keb.Failure{
```
tobiscr, line 644:
```suggestion
					failures = append(failures, &keb.Failure{
```

Issue #263, opened 2021-10-13
Ory Hydra reconciler upgrade


cnvergence:
**Description**

Currently upgrade is not supported.
We are migrating [secret-migration.sh](https://github.com/kyma-project/kyma/blob/main/resources/ory/files/secret-migration.sh) job functionality to the custom action, as the install phase is done, we need to take care about not changing secrets during Ory and Kyma upgrades.
Previously we were handling them as files in the mounted volume from secret `ory-hydra-credentials`. There is an init container in ory-hydra [deployment](https://github.com/kyma-project/kyma/blob/main/resources/ory/charts/hydra/templates/deployment.yaml#L67), that is responsible for generating data in the database on boot and handling the upgrades.
`hydra migrate sql --config /etc/config/config.yaml`

**Reasons**

This is a critical feature for the Ory component lifecycle

**Attachments**

PR #289, opened 2021-10-20
Ory upgrade and config migration

werdes72:
See: https://github.com/kyma-incubator/reconciler/issues/263


cnvergence (dismissed):

kwiatekus (approved):
Tested in local reconciler instance

PR #312, opened 2021-10-26
Remove old nats-operator resources if any remain

VladislavPaskar:
Eventing NATS backend is being changed from operator to helm chart installation with jetstream. That's why the old operator resources need to be deleted, if any of them still remain in the cluster. It can only occur during the upgrade flow.


pkg/reconciler/kubernetes/kubeclient/kubeclient.go
```diff
@@ -37,6 +37,19 @@ type Metadata struct {
 	Kind      string
 }
 
+//go:generate mockery --name Client
+type Client interface {
```
clebs, line 41: This interface seems wrong to me 🤔 
Now we have 2 overlapping interfaces on essentially the same package (this and https://github.com/kyma-incubator/reconciler/blob/455f43a4dec7df24f98d6932a8bc12288da45730/pkg/reconciler/kubernetes/client.go).

If there are more needs from the k8s client we have for the reconciler we should reconsider the design of this package and of the existing interface, but having 2 interfaces that have the same name, almost identical import path and overlap in functions will lead to confusion, misuse and bugs.

VladislavPaskar: The main idea behind this interface was the mocking. In order to test our implementation, we need to mock the pkg/reconciler/kubernetes/kubeclient/kubeclient.go. The mockery tool which we use for mocking purposes requires an interface for that. And in general why not have an interface for the kubeclient, perhaps we will need to have another implementation of that in the future. 
As I see from the original source of the code https://github.com/billiford/go-clouddriver/blob/master/pkg/kubernetes/client.go, the introduced an interface too. 

About the confusing naming I have to agree. But the we already got two clients in the same package anyway. My suggestion would be to just rename the interface in order to avoid the misuses and misinterpretations. What would be your naming suggestion ? I could only come up with Client or ClientStruct until we come we some design changes in this package.

clebs: As you mention, the kubernetes package already has an interface defining the way to access it in `/pkg/kubernetes/client.go`.
If you need more functionality from that, you can either add it there or use its `Clientset()` function to get the underlying kubernetes full Clientset.

As for the interface only for mocking, you can define that in your own package test file and generate the mock from it. Golang has implicit interfaces, so you do not need to create your mocking interface in `/pkg/kubernetes/kubeclient/`.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = r.removeNatsOperatorCRDs(kubeClient, logger)
+	return err
+}
+
+// delete the leftover CRDs, which were outside of charts
+func (r *removeNatsOperatorStep) removeNatsOperatorCRDs(kubeClient kubeclient.Client, log *zap.SugaredLogger) error {
```
clebs, line 109: I see the CRDs that are removed here on the `cluster-essentials` chart:
https://github.com/kyma-project/kyma/blob/main/resources/cluster-essentials/files/natsserviceroles.nats.crd.yaml
https://github.com/kyma-project/kyma/blob/main/resources/cluster-essentials/files/natsclusters.nats.crd.yaml

Therefore I am not sure if they need to be removed here.

VladislavPaskar: These CRDs are outdated and will be removed as a part of this issue as well: https://github.com/kyma-project/kyma/pull/12579

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func newRemoveNatsOperatorStep() *removeNatsOperatorStep {
+	return &removeNatsOperatorStep{
+		kubeClientProvider: defaultKubeClientProvider,
+	}
+}
+
+func defaultKubeClientProvider(context *service.ActionContext, logger *zap.SugaredLogger) (*kubeclient.KubeClient, error) {
+	kubeClient, err := kubeclient.NewKubeClient(context.Task.Kubeconfig, logger)
```
tobiscr, line 42: [You don't need this fct.](https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/service/action.go#L15). The `kubeclient.KubeClient` won't become accessible soon ([PR is open for this change](https://github.com/kyma-incubator/reconciler/pull/384/files#diff-7db9da48a12f7d364dd6a81c34f9d6d1fa160e2bd9ced27e0778e5b78ec67f1dR3))

tobiscr (changes requested):

PR #328, opened 2021-10-29
Fix statefulset check

rakesh-garimella:
The reconciler currently does not wait until the statefulset is coming up. Fix the the check so that it waits for statefulset to come up. Similar fix for Daemonset check was provided here: https://github.com/kyma-incubator/reconciler/pull/259


pkg/reconciler/kubernetes/progress/progress.go
```diff
@@ -202,12 +202,7 @@ func (pt *Tracker) statefulsetInState(inState State, object *resource) (bool, er
 		if err != nil {
 			return false, err
 		}
-		for _, condition := range statefulSet.Status.Conditions {
-			if condition.Status != v1.ConditionTrue {
-				return false, nil
-			}
-		}
-		return true, err
+		return statefulSet.Status.ReadyReplicas == statefulSet.Status.Replicas, nil
```
tobiscr, line 205: But isn't a stateful set already treated as ready when 1 of the is ready (like `statefulSet.Status.ReadyReplicas > 0`) ?

rakesh-garimella: I would say its kind of misleading from the cli point of view as it would say `kyma is installed` although some pods are still starting. up

suleymanakbas91 (approved):

Issue #285, opened 2021-10-19
Kyma CLI / Reconciler CLI allow the definition of a URL for a component


tobiscr:
**Description**

The Kyma CLI and the Reconciler CLI have to be able to support URLs for components. This is required to install CMP components (which are located outside of the Kyma repository) by the reconciler.

Following options have to be support by both CLIs:

* Define optionally a URL as part of a components definition via CLI parameter (e.g.  `--components cmp@compass-ns@schema//host/path`)
* Define a URL as part of the [components file](https://github.com/kyma-project/kyma/blob/main/installation/resources/components.yaml) (e.g. by adding  a `url` field which can be optionally specified for a component entry)

ACs:
* URL can be passed to Kyma CLI / Reconciler CLI as part of the components-definition provided via CLI parameter
* URL can be defined in the components-file and is considered by the Kyma CLI / Reconciler CLI and passed properly when  the reconciler-API is triggered

**Reasons**

External located components have to be supported also via Kyma CLI and Reconciler CLI to enable the CMP team to install their components via the reconciler.

**Attachments**

PR #335, opened 2021-11-02
Kyma cli / reconciler cli allow the definition of a url for a component

rJankowski93:
https://github.com/kyma-incubator/reconciler/issues/285


cmd/mothership/local/options.go
```diff
@@ -80,13 +72,19 @@ func componentsFromStrings(list []string, values []string) ([]*keb.Component, er
 	}
 
 	for _, item := range list {
-		s := strings.Split(item, "@")
-		name := s[0]
+		if strings.HasPrefix(item, "{") {
```
pPrecel, line 75: why are you checking if this string contains `{`? May it be a JSON? If yes why are you not removing the last JSON symbol?

rJankowski93: Variable `item` can looks like `{reafter,rafterNs,rafterUrl}`
But someone can pass only one value e.g. `rafter` 
This if check if we get three things(name, namespace and url) or one thing(name). 
In previous version we usevchar `@` but it can be problem in url in future. 

In line 76 we remove first char `{` and last char `}`.

clebs, line 75: Shouldn't this always be true?

pPrecel: yeah... right, I missed it 🤦

cmd/mothership/local/options.go
```diff
@@ -52,18 +52,10 @@ func componentsFromFile(path string) ([]string, []string, error) {
 
 	for _, c := range compList.Prerequisites {
 		preComps = append(preComps, c.Name)
-		if c.Namespace != "" {
-			defaultComps = append(defaultComps, c.Name+"@"+c.Namespace)
-		} else {
-			defaultComps = append(defaultComps, c.Name)
-		}
+		defaultComps = append(defaultComps, fmt.Sprintf("{%s,%s,%s}", c.Name, c.Namespace, c.URL))
```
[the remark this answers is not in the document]
rJankowski93: All empty no make sens, because if we don't get the name of component we don't recognise component. 

Empty namespace and/or url is ok. In this case reconciler get default value. 
[test](https://github.com/kyma-incubator/reconciler/blob/acb15111c7922b6e11dea184b5e1db2475174de1/cmd/mothership/local/local_test.go#L10)

cmd/mothership/local/options.go
rJankowski93: I added this because I wanted keep compatibility.

pPrecel (approved):
LGTM

rJankowski93:
LGTM

PR #336, opened 2021-11-02
Refactor Eventing reconciler to allow multi-step actions

marcobebway:
**Description**

Refactor Eventing reconciler to allow multi-step actions.


pkg/reconciler/instances/eventing/preaction/preaction.go
```diff
@@ -0,0 +1,42 @@
+package preaction
```
clebs, line 1: The code in this file and for `postaction` look almost identical, is this duplication meaningful?
Might be better to have one step looping logic and concentrate the differences on the specific steps + parametric initialisation.

marcobebway: You are right. Changed.

clebs (approved):

clebs (approved):

Issue #342, opened 2021-11-03
ORY pre-delete action


strekm:
**Description**

Implement pre-delete action for ORY reconciler which will delete `ory-hydra-credentials` created in pre-install step.

**Reasons**

Kyma should be deleted properly. 

**Attachments**

strekm:
kyma cli will be bumped with https://github.com/kyma-incubator/reconciler/issues/271

PR #352, opened 2021-11-04
Add preDeleteAction to ory reconciler

raypinto:
See https://github.com/kyma-incubator/reconciler/issues/342

### DIFF
--- a/pkg/reconciler/instances/ory/action.go
+++ b/pkg/reconciler/instances/ory/action.go
@@ -131,12 +131,15 @@ func (a *preDeleteAction) Run(context *service.ActionContext) error {
 		return errors.Wrap(err, "failed to retrieve native Kubernetes GO client")
 	}
 
-	secretExists, err := a.dbSecretExists(context.Context, client, dbNamespacedName, logger)
+	secretExists, err := a.dbSecretExists(context.Context, client, dbNamespacedName)
 	if err != nil {
-		return errors.Wrap(err, "failed to get DB secret")
+		return errors.Wrapf(err, "failed to get DB secret %s", dbNamespacedName.Name)
 	}
 	if secretExists {
-		a.deleteSecret(context.Context, client, dbNamespacedName, logger)
+		err = a.deleteSecret(context.Context, client, dbNamespacedName, logger)
+		if err != nil {
+			return errors.Wrapf(err, "failed to delete DB secret %s", dbNamespacedName.Name)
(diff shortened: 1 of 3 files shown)


pkg/reconciler/instances/ory/ory.go
```diff
@@ -18,10 +18,13 @@ func init() {
 	}
 
 	reconciler.
-		WithPreReconcileAction(&preAction{
+		WithPreReconcileAction(&preInstallAction{
```
Tomasz-Smelcerz-SAP, line 21: The name change implies "installation" operation. Will the same code also handle upgrades? If so, then the new name isn't really good.

pkg/reconciler/instances/ory/action.go
```diff
@@ -163,6 +191,27 @@ func (a *postAction) patchSecret(ctx context.Context, client kubernetes.Interfac
...
+		}
+		return false, errors.Wrap(err, "Could not get DB secret")
+	}
+	return true, nil
+}
+
+func (a *preDeleteAction) deleteSecret(ctx context.Context, client kubernetes.Interface, name types.NamespacedName, logger *zap.SugaredLogger) error {
+	err := client.CoreV1().Secrets(name.Namespace).Delete(ctx, name.Name, metav1.DeleteOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to delete the secret")
```
Tomasz-Smelcerz-SAP, line 208: I don't know the API, but perhaps it's worth to check if `if kerrors.IsNotFound(err)` and if so, return true here.
It is one of the common network programming errors: You assume that between the check if the secret exists (time == t[0]) until the moment you actually request a delete (time == t[1] == t[0]+delta), nothing changes.
What if your delta is 10 seconds? Easily possible in network scenarios. Ordinary user can delete the secret by hand within that time.

pkg/reconciler/instances/ory/ory.go
raypinto: That's a good point. It is used for upgrades as well. I will rename it to preReconcileAction and postReconcileAction respectively.

pkg/reconciler/instances/ory/action.go
raypinto: It's an edge case. But a valid one. Will do.

Tomasz-Smelcerz-SAP (approved):

PR #353, opened 2021-11-04
Configuration Improvements

ralikio:
Changes:
- Initial component removal
- Setting default configuration
- Conditional helm charts download


pkg/reconciler/instances/connectivityproxy/action.go
```diff
@@ -35,7 +50,7 @@ func (a *CustomAction) Run(context *service.ActionContext) error {
 		}
 	}
 
-	if binding != nil && app == nil {
+	if (binding != nil && app == nil) || (binding != nil && app != nil) {
```
franpog859, line 53: Should we install it even if there is no binding? I thought that it should be installed only when user creates a binding for the Connectivity Service

pkg/reconciler/instances/connectivityproxy/command_test.go
```diff
@@ -145,8 +145,8 @@ func TestCommandInstall(t *testing.T) {
 
 		err := commands.Install(actionContext, secret)
 		require.Equal(t, map[string]interface{}{
-			"key-1": []byte("value-1"),
-			"key-2": []byte("value-2"),
+			"binding.key-1": []byte("value-1"),
+			"binding.key-2": []byte("value-2"),
```
franpog859, line 149: It'd be superb to also add the JSON marshalled values to see if the implementation works correctly in these cases

pkg/reconciler/instances/connectivityproxy/connectivityproxy.go
```diff
@@ -15,39 +16,44 @@ const (
...
+	}
+	reconcilerInstance.
+		WithDeleteAction(&action).
+		WithReconcileAction(&action)
 }
 
-func registrySecretCopy(configs map[string]interface{}, inClusterClientSet, targetClientSet k8s.Interface) *SecretCopy {
+func registrySecretCopy(task *reconciler.Task, inClusterClientSet, targetClientSet k8s.Interface) *SecretCopy {
+	configs := task.Configuration
+	toNamespace := task.Namespace
```
franpog859, line 53: Is the `registryConfigPrefix` somehow preserved or it should be dropped there?

franpog859 (commented):
What does mean the "Initial component removal" from the PR description?

pkg/reconciler/instances/connectivityproxy/action.go
franpog859: Oh, my bad :d It's fine. I got lost in the logical operands

pkg/reconciler/chart/helm.go
```diff
@@ -37,7 +38,7 @@ func NewHelmClient(chartDir string, logger *zap.SugaredLogger) (*HelmClient, err
 func (c *HelmClient) Render(component *Component) (string, error) {
 	var helmChart *chart.Chart
 	var err error
-	if component.url != "" {
+	if component.url != "" && !strings.HasSuffix(component.url, ".git") {
```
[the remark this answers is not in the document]
ralikio: Both downloading helm charts and git repository operates on `component.url` field. Above change makes it possible for both to coexists.

pkg/reconciler/instances/connectivityproxy/connectivityproxy.go
ralikio: It should be preserved.

pkg/reconciler/chart/helm.go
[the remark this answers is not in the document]
franpog859: Good!

pkg/reconciler/instances/connectivityproxy/connectivityproxy.go
franpog859: Good! :D

internal/components/componentlist.go
```diff
@@ -16,10 +16,11 @@ type ComponentList struct {
...
 type Component struct {
-	Name          string
-	Namespace     string
-	URL           string
-	Configuration map[string]interface{}
+	Name              string
+	Namespace         string
+	URL               string
+	Configuration     map[string]interface{}
+	FlatConfiguration map[string]interface{}
```
tobiscr, line 23: Please be aware coming refactoring in PR #404 :

The workspace factory is currently using a struct from the `internal`  pkg as part of its interface definition which isn't a good practise. We are changing this in #404 and will use the component from the `chart` package.

pkg/reconciler/chart/provider.go
```diff
@@ -145,15 +145,16 @@ func (p *DefaultProvider) newWorkspace(component *Component) (*workspace.Workspa
...
 		c := components.Component{
-			Name:          component.name,
-			URL:           component.url,
-			Configuration: configuration,
+			Name:              component.name,
+			URL:               component.url,
+			Configuration:     configuration,
+			FlatConfiguration: component.configuration,
 		}
 		ws, err = p.wsFactory.Get(component.version, &c)
```
tobiscr, line 159: In #404 the GET will expect the component of the `chart`package. This makes the creation of the `components.Component` struct obsolete.

pkg/reconciler/git/clone.go
```diff
@@ -97,8 +97,9 @@ func (r *Cloner) CloneAndCheckout(dstPath, rev string) error {
 }
 
 func (r *Cloner) buildAuth() (transport.AuthMethod, error) {
-	if r.repo.TokenNamespace == "" {
-		return nil, nil
+	tokenNamespace := "default"
```
tobiscr, line 100: Still not convinced that it's a good approach to let the reconciler do some Secret-lookups to retrieve Git credentials instead providing them directly as configuration-parameter. This couples the GIT library of the reconciler to company internal deployment agreements.

@piotrmiskiewicz  / @PK85 : can't we provide such credentials as part of the component configuration provided by KEB?

pkg/reconciler/git/clone.go
```diff
@@ -111,14 +112,14 @@ func (r *Cloner) buildAuth() (transport.AuthMethod, error) {
...
 
 	secret, err := r.inClusterClientSet.CoreV1().
-		Secrets(r.repo.TokenNamespace).
+		Secrets(tokenNamespace).
 		Get(context.Background(), secretKey, v1.GetOptions{})
 
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil && !apierrors.IsNotFound(err) && !apierrors.IsForbidden(err) {
 		return nil, err
 	}
```
tobiscr, line 120: This will hide failures in our SKR deployment. If a token-Namespace is defined but the corresponding secret is missing or cannot be read caused by insufficient permissions we should also fail as it's an invalid state, shoudn't we?

Currently, I would assume the GIt request would just return "forbidden" (because the authentication wasn't provided). Could become a challenge to figure whether the credential is wrong  or just not provided by because the secret wasn't readable.

pkg/reconciler/workspace/factory.go
```diff
@@ -138,18 +139,17 @@ func (f *DefaultFactory) clone(version, dstDir string, component ...*components.
 
 	repo := f.repository
 	if len(component) > 0 {
```
tobiscr, line 141: Please be aware that part will disappear with #404

tobiscr (changes requested):

pkg/reconciler/instances/connectivityproxy/command_test.go
ralikio: Outdated.

pkg/reconciler/git/clone.go
```diff
@@ -111,14 +112,14 @@ func (r *Cloner) buildAuth() (transport.AuthMethod, error) {
...
 
 	secret, err := r.inClusterClientSet.CoreV1().
-		Secrets(r.repo.TokenNamespace).
+		Secrets(tokenNamespace).
 		Get(context.Background(), secretKey, v1.GetOptions{})
 
 	if err != nil && !apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
ralikio: Outdated.

pkg/reconciler/git/clone.go
ralikio: I do not recall any specific mechanics for that. I suggest to handle it outside of this PR.

pkg/reconciler/git/clone.go
ralikio: I have changed the logic so that token namespace defaults to a "default" namespace. I am lacking the possibility to log anything in clone.go. I suggest adding a logger and informing about that specific case.

tobiscr: sure, if we need a logger let's add one

pkg/reconciler/git/clone.go
```diff
@@ -111,14 +115,15 @@ func (r *Cloner) buildAuth() (transport.AuthMethod, error) {
 	}
 
 	secret, err := r.inClusterClientSet.CoreV1().
-		Secrets(r.repo.TokenNamespace).
+		Secrets(tokenNamespace).
 		Get(context.Background(), secretKey, v1.GetOptions{})
 
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil && !apierrors.IsNotFound(err) && !apierrors.IsForbidden(err) {
+		r.logger.Info("Token not found of forbidden")
```
franpog859, line 122: I think it's an error but NOT NotFound nor Forbidden 🤔 Is it correct?

tobiscr (approved):

tobiscr (approved):

franpog859 (approved):

Issue #343, opened 2021-11-03
ORY custom post reconcile action regenerates JWKS and patches secret even if not required


kwiatekus:
**Description**

ORY reconciliation actions causes unnecessary jwks regeneration and deployments restart.
This is error-prone in the scope of single cluster and imposes computational resources problem for the reconciler.

**Expected result**

JWKS should be regenerated based on detection mechanism ( i.e when existing one on the runtime is missing or corrupted )

**Actual result**

JWKS is regenerated and Secret is patched with every reconciliation causing ory deployments to be redeployed every time reconciliation is performed  

**Steps to reproduce**

Build reconciler binaries
```
make build-darwin
```
Run ory reconciliation locally against any kyma runtime:
```
./bin/mothership-darwin local --components ory  --kubeconfig {path to your cluster's kubeconfig} --value global.domainName='{your cluster's domain}'
```

strekm:
https://github.com/kyma-incubator/reconciler/pull/356

strekm:
still we need to modify ORY chart

cnvergence:
Decided to just create once this secret, so we will need to remove some jobs and change Ory Helm Chart regarding this secret.
For rotating this secret in Kyma 1.x, there was a cronjob responsible for that, we can also implement that in the reconciler.

strekm:
https://github.com/kyma-project/kyma/pull/12550

strekm:
@mjakobczyk & @piotrkpc doing review

PR #356, opened 2021-11-04
Patch the JWKS secret only when necessary

cnvergence:
**Related issue(s)**
 
See: https://github.com/kyma-incubator/reconciler/issues/343

### DIFF
--- a/pkg/reconciler/instances/ory/action.go
+++ b/pkg/reconciler/instances/ory/action.go
@@ -63,7 +63,7 @@ func (a *preReconcileAction) Run(context *service.ActionContext) error {
 		return errors.Wrap(err, "failed to retrieve native Kubernetes GO client")
 	}
 
-	secretObject, err := a.getDBConfigSecret(context.Context, client, dbNamespacedName)
+	secretObject, err := getSecret(context.Context, client, dbNamespacedName)
 	if err != nil {
 		if !kerrors.IsNotFound(err) {
 			return errors.Wrap(err, "Could not get DB secret")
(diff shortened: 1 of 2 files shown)


pkg/reconciler/instances/ory/action.go
```diff
@@ -107,23 +108,36 @@ func (a *postAction) Run(context *service.ActionContext) error {
...
-	patchData, err := jwks.Get(jwksAlg, jwksBits)
+	secretObject, err := getSecret(context.Context, client, jwksNamespacedName)
 	if err != nil {
-		return errors.Wrap(err, "failed to generate JWKS secret")
+		return errors.Wrap(err, "Could not get JWKS secret")
 	}
-	if err := a.patchSecret(context.Context, client, jwksNamespacedName, patchData, logger); err != nil {
-		return errors.Wrap(err, "failed to patch Ory secret")
+
+	if isEmpty(secretObject) {
```
strekm, line 116: why do we need to check if secret is empty?

cnvergence: Because we create it in the Helm chart: https://github.com/kyma-project/kyma/blob/main/resources/ory/charts/oathkeeper/templates/secret.yaml

cnvergence: My initial idea to create the secret beforehand would result in much more work needed to implement that.
Current Ory Oathkeeper helm chart requires this secret to be mounted as a volume.

cnvergence: So we would need to also change the whole structure of Ory Reconciler.

mjakobczyk (approved):

PR #362, opened 2021-11-08
Create State helper generator

m00g3n:
Create `State` helper generator to sync oapi model with the rest of the code. 
See also: https://github.com/kyma-project/control-plane/issues/1060


configs/reconciler.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 db:
-  driver: postgres
+  driver: sqlite
```
tobiscr, line 3: Not sure if this will work.. on MAC I'm getting failiures that SQLite support isn't available in the default binaries

tobiscr (changes requested):

configs/reconciler.yaml
m00g3n: Shouldn't the `reconciler.db` be removed from being tracked? 
It works if you change the `sqlite.file` entry.

tobiscr (approved):

PR #370, opened 2021-11-09
Generate value of SecretsCookie in Ory Hydra secret

cnvergence:
SecretsCookie should be generated and not empty on Install
```
The configuration contains values or keys which are invalid:
secrets.cookie: <nil>
                ^-- expected array, but got null

time=2021-11-09T16:33:42Z level=fatal msg=Unable to instantiate configuration. audience=application error=map[message:I[#/secrets/cookie] S[#/properties/secrets/properties/cookie/type] expected array, but got null] service_name=ORY Hydra service_version=v1.10.7
```


werdes72 (approved):
/woof

Issue #365, opened 2021-11-09
istio reconciler should support downgrades


strekm:
**Description**

At this moment istio reconciler is not supporting downgrades. This can lead to situation that on managed kyma customer can bump istio to newer version before kyma supports it. In such situation istio reconciler should downgrade istio to istio supported in kyma. 

AC:
- [x] istio downgrade is supported within one minor version

**Reasons**

Reconciler should always apply desired state, in this case managed kyma istio.

**Attachments**

raypinto:
Cli is bumped. Tested with the kyma-cli and it works as expected!

PR #380, opened 2021-11-10
Istio downgrade is now permitted upto one minor

raypinto:
See https://github.com/kyma-incubator/reconciler/issues/365


veichtj (dismissed):

pkg/reconciler/instances/istio/action.go
```diff
@@ -225,8 +235,10 @@ func canUpdate(ver actions.IstioVersion, logger *zap.SugaredLogger) bool {
 	dataPlaneVsTarget := targetHelperVersion.compare(&dataPlaneHelperVersion)
 
 	if pilotVsTarget == -1 || dataPlaneVsTarget == -1 {
-		logger.Errorf("Downgrade detected from pilot: %s and data plane: %s to version: %s - finishing...", ver.PilotVersion, ver.DataPlaneVersion, ver.TargetVersion)
-		return false
+		if !targetHelperVersion.isDowngradePermitted(&pilotHelperVersion) || !targetHelperVersion.isDowngradePermitted(&dataPlaneHelperVersion) {
+			logger.Errorf("Downgrade detected from pilot: %s and data plane: %s to version: %s - finishing...", ver.PilotVersion, ver.DataPlaneVersion, ver.TargetVersion)
+			return false
+		}
 	}
```
veichtj, line 242: We are mixing here the upgrade and the downgrade case, what might be confusing for one who reads that. The `canUpdate` function says now true if either update or downgrade is allowed. 
What about creating a new function like "canReconcile" which either returns true if update or downgrade is possible?

pkg/reconciler/instances/istio/action.go
```diff
@@ -201,6 +201,16 @@ func canUninstall(istioVersion actions.IstioVersion) bool {
...
+func (h *helperVersion) isDowngradePermitted(second *helperVersion) bool {
+	// we permit one minor downgrade, please see: https://github.com/kyma-incubator/reconciler/issues/365
+	if h.major == second.major {
+		if second.minor-h.minor == 0 || second.minor-h.minor == 1 {
+			return true
+		}
+	}
+	return false
+}
+
```
veichtj, line 213: Maybe also add here a function isUpdgradePermitted to have the logic compressed in the `helperVersion`

veichtj (commented):
You could also think about adding a "light" integration test in `istio_test.go` like there is already an example to tackle also the update / downgrade case

pkg/reconciler/instances/istio/action.go
raypinto: As discussed, "canReconcile" could indicate an install as well. Hence I've added an additional log to clearly specify the valid downgrade scenario - which is also an "update".

pkg/reconciler/instances/istio/action.go
```diff
@@ -225,8 +235,11 @@ func canUpdate(ver actions.IstioVersion, logger *zap.SugaredLogger) bool {
 	dataPlaneVsTarget := targetHelperVersion.compare(&dataPlaneHelperVersion)
 
 	if pilotVsTarget == -1 || dataPlaneVsTarget == -1 {
-		logger.Errorf("Downgrade detected from pilot: %s and data plane: %s to version: %s - finishing...", ver.PilotVersion, ver.DataPlaneVersion, ver.TargetVersion)
-		return false
+		if !targetHelperVersion.isDowngradePermitted(&pilotHelperVersion) || !targetHelperVersion.isDowngradePermitted(&dataPlaneHelperVersion) {
+			logger.Errorf("Downgrade detected from pilot: %s and data plane: %s to version: %s - finishing...", ver.PilotVersion, ver.DataPlaneVersion, ver.TargetVersion)
+			return false
+		}
+		logger.Infof("Valid Downgrade detected from pilot: %s and data plane: %s to version: %s", ver.PilotVersion, ver.DataPlaneVersion, ver.TargetVersion)
```
Tomasz-Smelcerz-SAP, line 242: I would return "true" here immediately, in that way you don't have to reason what the logic in lines 245-249 does.

pkg/reconciler/instances/istio/action.go
```diff
@@ -201,6 +201,16 @@ func canUninstall(istioVersion actions.IstioVersion) bool {
 	return isInstalled(istioVersion) && istioVersion.ClientVersion != ""
 }
 
+func (h *helperVersion) isDowngradePermitted(second *helperVersion) bool {
+	// we permit one minor downgrade, please see: https://github.com/kyma-incubator/reconciler/issues/365
```
mjakobczyk, line 205: I find such reasoning as unnecessary explanations. In addition, we should not provide comments next to the code (we have documentation for it and this one is not aligned with clean code rules).

pkg/reconciler/instances/istio/istio_test.go
```diff
@@ -48,14 +51,124 @@ const istioctlMockCompleteVersion = `{
...
+		// when
+		err := action.Run(actionContext)
+
+		// then
+		require.NoError(t, err)
+		commanderMock.AssertCalled(t, "Version", mock.Anything, mock.Anything)
+		commanderMock.AssertCalled(t, "Upgrade", mock.Anything, mock.Anything, mock.Anything)
+	})
+
+	t.Run("Istio update should NOT permit more than one minor downgrade", func(t *testing.T) {
```
mjakobczyk, line 142: Why do you uppercase `NOT` in this context?

pkg/reconciler/instances/istio/istio_test.go
```diff
@@ -48,14 +51,124 @@ const istioctlMockCompleteVersion = `{
...
+	  }`
+)
 
-		wsf, _ := workspace.NewFactory(nil, "./test_files", log.NewLogger(true))
+func Test_RunUpdateAction(t *testing.T) {
+	wsf, _ := workspace.NewFactory(nil, "./test_files", log.NewLogger(true))
+	model := reconciler.Task{
+		Component: "istio-configuration",
+		Namespace: "istio-system",
+		Version:   "1.11.2",
```
mjakobczyk, line 116: Can you put here some fake Istio version? We don't have to be that specific when we put it into unit tests. It can be something like: `1.0.0` or `0.1.0`.

pkg/reconciler/instances/istio/action.go
```diff
@@ -201,6 +201,16 @@ func canUninstall(istioVersion actions.IstioVersion) bool {
 	return isInstalled(istioVersion) && istioVersion.ClientVersion != ""
 }
 
+func (h *helperVersion) isDowngradePermitted(second *helperVersion) bool {
```
mjakobczyk, line 204: How about introducing a few simple unit tests for this function only? For testing purposes I changed the implementation to the following case:
```go
if second.minor-h.minor == 0 || second.minor-h.minor == 1 || second.minor-h.minor == 2
```
and tests were still green, whereas they should not. Please provide unit tests for this contract.

pkg/reconciler/instances/istio/action.go
```diff
@@ -201,6 +201,16 @@ func canUninstall(istioVersion actions.IstioVersion) bool {
...
+	// we permit one minor downgrade, please see: https://github.com/kyma-incubator/reconciler/issues/365
+	if h.major == second.major {
+		if second.minor-h.minor == 0 || second.minor-h.minor == 1 {
```
mjakobczyk, line 207: This can be sampled:
```suggestion
		if second.minor == h.minor || second.minor - h.minor == 1 {
```

pkg/reconciler/instances/istio/action.go
mjakobczyk, line 204: I don't like the fact you bind some `downgrade` scenario to the `helperVersion` struct. How about such approach:
- for `helperVersion` you can define a find `differsAtMostByOneMinor()` that will check exactly what the function says,
- for checking the permissible downgrade you can introduce just a local private function or the one that will be attached to the action and inside you will just call the function defined in the previous step, such as:

```go
func (a *ReconcileAction) isDowngradePermitted(current, target helperVersion) bool {
  // implementation details...
}
```

EDIT: I just realised we do already have such function in place and this one is called: `maxOneMinorBehind()` so please reuse it.

mjakobczyk (changes requested):

pkg/reconciler/instances/istio/istio_test.go
raypinto: I think it explicitly states why this test is in place. And this pattern is consistent also across our other components. For example: https://github.com/kyma-incubator/api-gateway/blob/7e5551484fef46333cf66058fb0f6a68ec00cf8f/internal/validation/validate_test.go#L164

pkg/reconciler/instances/istio/istio_test.go
veichtj: It can be yes, but should we shouldn't we prefer having "real" test data configuration, as you might not detect bugs when you use fake data?

pkg/reconciler/instances/istio/action.go
raypinto: I honestly don't see a problem with documentation and I'm always used to commenting when necessary. If we follow the pattern in the team to not comment, I will take it off.

pkg/reconciler/instances/istio/action.go
raypinto: That's right! I will add additional tests for this function.

pkg/reconciler/instances/istio/action.go
raypinto: Will do.

pkg/reconciler/instances/istio/action.go
raypinto: Will do!

mjakobczyk (approved):

veichtj (approved):

PR #384, opened 2021-11-10
Refactor kubernetes reconciler pkg to minimize exposing internals

clebs:
Improve the reconciler kubernetes API to hide the internals and avoid teams misusing it in the custom component reconcilers.


tobiscr (approved):

pkg/reconciler/kubernetes/internal/warnings.go
```diff
@@ -1,4 +1,4 @@
-package kubeclient
+package internal
```
dbadura, line 1: Why not to move those things to package `kubeclient` inside of `internal`?

clebs: Would also be fine, I just renamed the `kubeclient` package as is and did not see the need for a single sub-package.
Package `kubernetes` defines the `API` and internal just has the implementation.

veichtj (approved):

tobiscr (dismissed):

Maladie (dismissed):

pkg/reconciler/instances/rma/lazy_client.go
```diff
@@ -25,7 +25,7 @@ type LazyKubeClient struct {
 
 func (c *LazyKubeClient) init() error {
 	c.initClient.Do(func() {
-		c.client, c.clientErr = kubeclient.NewInClusterClient(c.log)
+		c.client, c.clientErr = reconK8s.NewInClusterClientSet(c.log)
```
ebensom, line 28: Wouldn't it be better to initialize only a new `InClusterClient` here, and return a new `ClientSet` from reconciler kubeclient at every call to LazyKubeClient.ClientSet() instead of returning the same clientset instance?

In this case different workers would use different K8S client with separate client-side throttling (rate-limiting) bucket, HTTP client, etc. instead of a static, "global" shared clientset

clebs: Ideally I wanted to do it like that, but the in-cluster client generally has less things than the local one (e.g. no kubeconfig). So it would break the `Client` interface that we have.

It is a good improvement I would like to add but I already am now constantly having to fix issues because of teams misusing the package so it would be good to at least get this merged ASAP and then improve upon it.

tobiscr (approved):

veichtj (approved):

akgalwas (approved):

PR #389, opened 2021-11-11
Service Catalog migration reconciler

wozniakjan:
It was decided that we should reimplement the logic from https://github.com/SAP/sap-btp-service-operator-migration as a reconciler. It's not trivial to import the code because the whole application is structured as CLI.

This PR imports what is importable and reimplements certain heavily CLI architectured parts of the application.

The most notable changes are:
- using proper logger instead of `fmt.Println`
- forward errors out of reconciliation loop instead of panic using `cobra.CheckErr`

there will be a follow-up PR containing similar logic for including https://github.com/kyma-incubator/sc-removal


go.mod
```diff
@@ -41,3 +45,6 @@ require (
 	k8s.io/kubectl v0.22.1
 	sigs.k8s.io/yaml v1.2.0
 )
+
+// NOTE: remove with sc-migration reconciler
+replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.9.0
```
wozniakjan, line 50: there is a conflict in dependencies, [sap-btp-service-operator](https://github.com/SAP/sap-btp-service-operator/blob/ae34a509c2e1f1ca9465ea227a752f0f57b4d130/go.mod#L18) requires v0.7.0 which imports incompatible prometheus client

```
$ make build-linux
CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./bin/reconciler-linux -ldflags '-s -w' ./cmd/reconciler
# sigs.k8s.io/controller-runtime/pkg/metrics
../../../../pkg/mod/sigs.k8s.io/controller-runtime@v0.7.0/pkg/metrics/client_go_adapter.go:134:3: cannot use &latencyAdapter{...} (type *latencyAdapter) as type metrics.LatencyMetric in field value:
        *latencyAdapter does not implement metrics.LatencyMetric (wrong type for Observe method)
                have Observe(string, url.URL, time.Duration)
                want Observe(context.Context, string, url.URL, time.Duration)
../../../../pkg/mod/sigs.k8s.io/controller-runtime@v0.7.0/pkg/metrics/client_go_adapter.go:135:3: cannot use &resultAdapter{...} (type *resultAdapter) as type metrics.ResultMetric in field value:
        *resultAdapter does not implement metrics.ResultMetric (wrong type for Increment method)
                have Increment(string, string, string)
                want Increment(context.Context, string, string, string)
make: *** [Makefile:30: build-linux] Error 2
```

but given the plan is that after we perform the migration, this reconciler will be purged, I would like to propose having this temporary replace directive as a path of least resistance.

tobiscr (approved):

pkg/reconciler/instances/scmigration/scmigration.go
```diff
@@ -0,0 +1,22 @@
...
+func init() {
+	log := logger.NewLogger(false)
+
+	log.Debugf("Initializing component reconciler '%s'", ReconcilerName)
+	reconciler, err := service.NewComponentReconciler(ReconcilerName)
+	if err != nil {
+		log.Fatalf("Could not create '%s' component reconciler: %s", ReconcilerName, err)
+	}
+
+	reconciler.WithDependencies("btp-operator")
```
tobiscr, line 20: Please drop this call as we are going to remove it from reconciler soon (we will add a deprectation mark to it soon).

Please implement your reconciler in way, that it is able to identify such dependencies by its own or can be fixed by retrying it until the dependency is in place.

tobiscr (dismissed):

wozniakjan:
todo: support execution on already migrated clusters (e.g. no svcat CRDs present)

jeremyharisch (approved):

PR #401, opened 2021-11-12
Allign patching istio-sidecar-injector with Kyma 1.x behaviour

cnvergence:
In Kyma 1.x, istio-sidecar-injector was patched also on the upgrades [link](https://github.com/kyma-project/kyma/blob/1.24.8/resources/istio/files/istio-upgrade.sh#L34).
This patch is needed for Gardener clusters and still can not be configured with Istio Helm values.

### DIFF
--- a/pkg/reconciler/instances/istio/action.go
+++ b/pkg/reconciler/instances/istio/action.go
@@ -113,6 +113,11 @@ func (a *ReconcileAction) Run(context *service.ActionContext) error {
 			return errors.Wrap(err, "Could not update Istio")
 		}
 
+		err = a.performer.PatchMutatingWebhook(context.KubeClient, context.Logger)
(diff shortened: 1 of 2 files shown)


pkg/reconciler/instances/istio/action.go
```diff
@@ -113,6 +113,11 @@ func (a *ReconcileAction) Run(context *service.ActionContext) error {
 			return errors.Wrap(err, "Could not update Istio")
 		}
 
+		err = a.performer.PatchMutatingWebhook(context.KubeClient, context.Logger)
```
Tomasz-Smelcerz-SAP, line 116: Something's wrong with formatting.

Tomasz-Smelcerz-SAP (approved):

PR #403, opened 2021-11-12
Make replace work for ClusterIP services

kwiatekus:
https://github.com/kyma-project/kyma/issues/12597


tobiscr:
@kwiatekus and I were implementing in pair programming session, therefore I'm not a good candidate for a review. Would be good to get a review by @clebs  or @suleymanakbas91.

pbochynski (commented):
If I understand correctly, the implementation the interceptor is retrieving the service using kubernetes client and after all interceptors are applied, we fetch the object again. It looks suboptimal. Probably it would be better if interceptors could get manifest and the target object (if exists), but it requires more refactoring. The benefit would be simpler implementation of the interceptor and better performance (less calls). Just for your consideration.

tobiscr:
thanks for the feedback @pbochynski . The interceptors should be called after we are getting a resource (inside of the replace: https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/kubernetes/kubeclient/replace.go#L102)

kwiatekus:
> thanks for the feedback @pbochynski . The interceptors should be called after we are getting a resource (inside of the replace: https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/kubernetes/kubeclient/replace.go#L102)

I created a follow up issue for the optimisation 
https://github.com/kyma-incubator/reconciler/issues/412

Its not doable in 5 min. We would probably aim to merge the current solution now to see if there are next problems behing the corner

m00g3n (approved):
LGTM

PR #427, opened 2021-11-17
Modify reconciliation info endpoint

dbadura:
Proposed changes:
- modify OpenAPI
- remove cluster metadata from response
- adjust response to be aligned to endpoint functionality

See: https://github.com/kyma-project/control-plane/issues/1083


tobiscr (dismissed):

cmd/mothership/mothership/start/helpers.go
```diff
@@ -19,40 +18,3 @@ func components(cfg model.ClusterConfigurationEntity) []keb.Component {
 	}
 	return components
 }
-
-func clusterMetadata(runtimeID string, lastState *cluster.State) *keb.Cluster {
```
kwiatekus, line 23: We should not drop the whole `cluster` part of the response.. but just the kubeconfig.
See https://github.com/kyma-incubator/reconciler/issues/433, where @tobiscr expects non empty `components` list as part of that object which is here being removed

kwiatekus (changes requested):

openapi/external_api.yaml
```diff
@@ -333,11 +362,24 @@ components:
...
           format: uri
 
-    HTTPReconciliationOperations:
+    HTTPReconciliationInfo:
       type: object
+      required: [ lock, runtimeID, schedulingID, created, updated, status,operations ]
       properties:
-        cluster:
-          $ref: "#/components/schemas/cluster"
+        lock:
```
pPrecel, line 369: please remove this and add the `finished` field

openapi/external_api.yaml
```diff
@@ -194,10 +194,29 @@ paths:
...
+            type: string
+            format: uuid
+        - name: version
+          required: true
+          in: path
+          schema:
+            type: string
+      responses:
+        "200":
+          $ref: "#/components/responses/configurationOkResponse"
```
pPrecel, line 218: One small suggestion:

```suggestion
          $ref: "#/components/responses/ConfigurationOkResponse"
```

openapi/external_api.yaml
dbadura: Replaced

openapi/external_api.yaml
dbadura: I will apply it in https://github.com/kyma-incubator/reconciler/pull/437

cmd/mothership/mothership/start/helpers.go
dbadura: The components config will be implemented in this PR in another endpoint:
https://github.com/kyma-incubator/reconciler/pull/437

kwiatekus (approved):

PR #431, opened 2021-11-17
Add preaction to delete the Nats-operator resources if any remain

VladislavPaskar:
Add preaction to delete the Nats-operator resources if any remain


pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -0,0 +1,109 @@
...
+		ProgressInterval: progressTrackerInterval,
+		ProgressTimeout:  progressTrackerTimeout,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return kubeClient, nil
+}
+
+func (r *removeNatsOperatorStep) Execute(context *service.ActionContext, logger *zap.SugaredLogger) error {
```
marcobebway, line 49: Currently, the execute logic would always render 1.X Charts/CRDs and tries to delete them, which would make sense in case of upgrading a 1.X cluster to 2.X. But if a 2.X cluster is being reconciled, the same code would try to delete 1.X Charts/CRDs again, and in my opinion that is not necessary.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/kubernetes/mocks"
+	"github.com/stretchr/testify/require"
+
+	"github.com/kyma-incubator/reconciler/pkg/logger"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/chart"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/service"
+)
+
+func TestDeletingNatsOperatorResources(t *testing.T) {
```
marcobebway, line 21: Could you please add a test-case to simulate a 2.X cluster which does not have 1.X resources?

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	action, actionContext, mockProvider, k8sClient := setup()
+
+	// execute the step
+	err := action.Execute(actionContext, actionContext.Logger)
+	require.NoError(t, err)
+
+	// ensure the right calls were invoked
+	mockProvider.AssertCalled(t, "RenderManifest", mock.Anything)
```
marcobebway, line 71: Please instead of `mock.Anything`, use the correct argument which would evaluate to a `*chart.Component`.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -0,0 +1,76 @@
...
+	k8sClient.AssertCalled(t, "Delete", actionContext.Context, mock.Anything, namespace)
```
marcobebway, line 72: Please use the following:

```golang
k8sClient.AssertCalled(t, "Delete", actionContext.Context, "testManifest", namespace)
```

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -0,0 +1,76 @@
...
+	k8sClient.AssertCalled(t, "DeleteResourceByKindAndNameAndNamespace", "customresourcedefinitions", natsOperatorCRDsToDelete[0], "kyma-system")
+	k8sClient.AssertCalled(t, "DeleteResourceByKindAndNameAndNamespace", "customresourcedefinitions", natsOperatorCRDsToDelete[1], "kyma-system")
```
marcobebway, line 74: Please use the `namespace` const instead of hardcoding `kyma-system`.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func TestDeletingNatsOperatorResources(t *testing.T) {
+	setup := func() (removeNatsOperatorStep, *service.ActionContext, *pmock.Provider, *mocks.Client) {
+		mainContext := context.TODO()
+		k8sClient := mocks.Client{}
+		log := logger.NewLogger(false)
+
+		mockProvider := pmock.Provider{}
+		mockManifest := chart.Manifest{
+			Manifest: "testManifest",
```
marcobebway, line 29: Please use a const for the manifest:

```golang
const manifest = "testManifest"
```

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (r *removeNatsOperatorStep) removeNatsOperatorResources(context *service.ActionContext, kubeClient kubernetes.Client, logger *zap.SugaredLogger) error {
+	// get charts from the version 1.2.x, where the NATS-operator resources still exist
+	comp := chart.NewComponentBuilder(natsOperatorLastVersion, natsSubChartPath).
+		WithConfiguration(map[string]interface{}{
+			oldConfigValue: newConfigValue, // replace the missing global value, as we are rendering on the subchart level
+		}).
+		WithNamespace(namespace).
+		Build()
```
marcobebway, line 74: Please extract this code into a function so we can use it in the test assertions.

pkg/reconciler/kubernetes/adapter.go
```diff
@@ -213,6 +213,18 @@ func (g *kubeClientAdapter) newNamespaceUnstruct(namespace string) (*unstructure
...
+		kind, name, namespace, metav1.DeleteOptions{})
+	deletedResource := toResource(metadata)
+	if err != nil && !k8serr.IsNotFound(err) {
+		g.logger.Errorf("Failed to delete Kubernetes unstructured resource kind='%s', name='%s', namespace='%s': %s",
+			kind, name, namespace, err)
+		return deletedResource, err
+	}
+	return deletedResource, nil
+}
+
```
marcobebway, line 227: If this code is not needed by any other component reconciler, then maybe we can bring it into the eventing reconciler only.

pkg/reconciler/kubernetes/adapter.go
```diff
@@ -213,6 +213,18 @@ func (g *kubeClientAdapter) newNamespaceUnstruct(namespace string) (*unstructure
...
+	metadata, err := g.kubeClient.DeleteResourceByKindAndNameAndNamespace(
+		kind, name, namespace, metav1.DeleteOptions{})
+	deletedResource := toResource(metadata)
+	if err != nil && !k8serr.IsNotFound(err) {
+		g.logger.Errorf("Failed to delete Kubernetes unstructured resource kind='%s', name='%s', namespace='%s': %s",
+			kind, name, namespace, err)
+		return deletedResource, err
+	}
+	return deletedResource, nil
+}
```
marcobebway, line 226: If this code is not needed by other component reconcilers, then we can move it inside the Eventing reconciler only.

pkg/reconciler/kubernetes/adapter.go
```diff
@@ -235,6 +247,12 @@ func (g *kubeClientAdapter) Delete(ctx context.Context, manifest, namespace stri
 	for i := len(unstructs) - 1; i >= 0; i-- {
 		unstruct := unstructs[i]
 
+		// execute the delete request only if the resource exists
+		_, err := g.kubeClient.Get(unstruct.GetKind(), unstruct.GetName(), namespace)
+		if k8serr.IsNotFound(err) {
+			continue
+		}
```
marcobebway, line 254: In my opinion this is not needed, because we already have `if !k8serr.IsNotFound(err)` after calling `DeleteResourceByKindAndNameAndNamespace`.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -0,0 +1,109 @@
...
+		resource, err := kubeClient.DeleteResourceByKindAndNameAndNamespace("customresourcedefinitions", crdName, namespace)
+		if err != nil && !errors.IsNotFound(err) {
+			logger.Errorf("Failed to delete the nats-operator CRDs, name='%s', namespace='%s': %s", crdName, namespace, err)
+			return err
+		}
+		resources = append(resources, resource)
+	}
+	if len(resources) > 0 {
+		logger.Info("NATS-operator CRDs are deleted")
+	}
```
marcobebway, line 107: I recommend to log exactly what was deleted inside the loop, and remove this if block.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// set the right eventing name, which went lost after rendering
+	manifest.Manifest = strings.ReplaceAll(manifest.Manifest, natsSubChartPath, eventingNats)
+
+	logger.Info("Removing nats-operator chart resources")
+	// remove all the existing nats-operator resources, installed via charts
+	resources, err := kubeClient.Delete(context.Context, manifest.Manifest, namespace)
+	if len(resources) > 0 {
+		logger.Info("Nats-operator chart resources are deleted")
```
marcobebway, line 88: This log might not always be accurate, see how's the `kubeClient.Delete` func is implemented, I see some return paths which have `return deletedResources, err` so maybe we have some resources which were successfully deleted and other resources which failed, which makes `len(resources) > 0 equals true`, but still we also get an error.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func newRemoveNatsOperatorStep() *removeNatsOperatorStep {
+	return &removeNatsOperatorStep{
+		kubeClientProvider: defaultKubeClientProvider,
+	}
+}
+
+func defaultKubeClientProvider(context *service.ActionContext, logger *zap.SugaredLogger) (kubernetes.Client, error) {
+	kubeClient, err := kubernetes.NewKubernetesClient(context.Task.Kubeconfig, logger, &kubernetes.Config{
```
tobiscr, line 39: QQ: The `ctx.KubeClient`isn't working out here?

pkg/reconciler/kubernetes/adapter.go
```diff
@@ -213,6 +213,18 @@ func (g *kubeClientAdapter) newNamespaceUnstruct(namespace string) (*unstructure
 	return nsUnstructs[0], nil
 }
 
+func (g *kubeClientAdapter) DeleteResourceByKindAndNameAndNamespace(kind, name, namespace string) (*Resource, error) {
```
tobiscr, line 216: If all parameters are required, maybe we call it just `DeleteResource` (quite long fct name otherwise :) ) ?

pkg/reconciler/kubernetes/adapter.go
VladislavPaskar: Actually the `if !k8serr.IsNotFound(err)` is being called after the DELETE request. The main idea of this added if was to use the "cheap" get call from the cache and ensure the resource actually exists before executing the "not cheap" Delete call

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
VladislavPaskar: This still works, but I used the factory method here because of the comments, that this kubeClient is gonna be hidden in the next PRs

pkg/reconciler/kubernetes/adapter.go
VladislavPaskar: Done

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
VladislavPaskar: I added a TODO to the code with the reference to the feature, which makes your comment implementable, for now I think we keep it this way.

pkg/reconciler/kubernetes/adapter.go
marcobebway: > use the "cheap" get call from the cache

Where do you see that get is using a cache?

VladislavPaskar: Sorry, my mistake, there is no cache operation there, so only get operation before delete

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -0,0 +1,94 @@
...
+	require.NoError(t, err)
+
+	// ensure the right calls were invoked
+	mockProvider.AssertCalled(t, "RenderManifest", mockedComponentBuilder)
+	k8sClient.AssertCalled(t, "Delete", actionContext.Context, manifestString, namespace)
+	k8sClient.AssertCalled(t, "DeleteResource", crdPlural, natsOperatorCRDsToDelete[0], namespace)
+	k8sClient.AssertCalled(t, "DeleteResource", crdPlural, natsOperatorCRDsToDelete[1], namespace)
+}
+
+// todo execute this test, when the check for kyma2x version is available, see the the todo comment from removenatsoperatorstep:Execute()
```
tobiscr, line 40: Please open also an issue for this so that we won't forget it

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
tobiscr: right - makes sense

tobiscr (approved):

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
VladislavPaskar: https://github.com/kyma-incubator/reconciler/issues/453

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -0,0 +1,129 @@
...
+	logger.Info("Removing nats-operator CRDs")
+	for _, crdName := range natsOperatorCRDsToDelete {
+		_, err := kubeClient.DeleteResource(crdPlural, crdName, namespace)
+		if err != nil && !errors.IsNotFound(err) {
+			logger.Errorf("Failed to delete the nats-operator CRDs, name='%s', namespace='%s': %s", crdName, namespace, err)
+			return err
+		}
+		if err == nil {
+			logger.Debugf("Deleted %s CRD from %s namespace", crdName, namespace)
+		}
```
marcobebway, line 116: Improving Evening reconciler logs should be done as part of https://github.com/kyma-incubator/reconciler/issues/460.

marcobebway (approved):

tobiscr (approved):

Issue #277, opened 2021-10-18
Add cluster status ' skipped'


tobiscr:
**Description**

Cluster configurations which will never be reconciled (because they got faster outdated as they could be reconciled) should receive the cluster status `skipped` instead staying endless in `reconcile_pending` status.

The change to the `reconcile_skipped` status should happen by the cluster-inventory (inside one DB-transaction):

1. query for cluster-status which are set to `reconcile_pending` for the particular cluster (SQL pseudo statement: `select * from inventory_cluster_config_statuses where status='reconcile_pending' and cluster_version=123`)
2. create for each `cluster_config` a new status which is `reconcile_pending`

AC:
* A new cluster version provided by KEB will update the status of any other cluster-configurations which are in status `reconcile_pending` to `reconcile_skipped`

**Reasons**

Simplify analysis for human users.

**Attachments**

janmedrek:
Functionalities of the Reconciler were migrated to the new architecture, this issue will not be worked on.

PR #435, opened 2021-11-18
Add cluster status skipped

rJankowski93:
https://github.com/kyma-incubator/reconciler/issues/277


pkg/cluster/inventory.go
```diff
@@ -187,6 +188,44 @@ func (i *DefaultInventory) createStatus(configEntity *model.ClusterConfiguration
...
 	}
 
+	// create new reconcile pending statuses
+	if status == model.ClusterStatusReconcilePending {
+		q, err := db.NewQuery(i.Conn, &model.ClusterStatusEntity{}, i.Logger)
+		if err != nil {
+			return nil, err
+		}
+
+		notReconciledStatues, err := q.Select().
```
tobiscr, line 198: We should not always query the whole history.  Would propose following flow:

* Implement DB migration logic (will require #442 ) and migrate all ol pending-status which were not applied to `skipped`
* Check just the previous cluster-state status, and if this is `pending`, just change this one to `skipped`

tobiscr (changes requested):

nachtmaar:
@tobiscr Can we close this PR ?

tobiscr:
Caused by strategy change to change the reconciler architecture to K8s native app, this feature isn't required anymore

Issue #374, opened 2021-11-10
Enhance automated tests to verify that multiple mothership-reconcilers will not cause race-conditions


tobiscr:
**Description**

Our unit tests have to be enhanced to ensure that multiple mothership reconcilers instances won't cause potential race reconciliations when operating in parallel with it.

Following cases can happen when multiple mothership reconcilers are running in parallel (e.g. by running them with >1 replicas):

_Case 1: reconciliations for the same cluster are created in parallel_
1. Enhance unit-test of reconciliation repository: ensure that an parallel creation of a reconciliation entry for a cluster is only successful once
2. Implement race-condition integration test:
     1. Test starts several scheduler instances in parallel which are watching the inventory in different intervals (e.g. 500ms, 550ms, 600ms etc.)
     2. Test creates several cluster instances (with some delay between the creation) in the inventory
     3. Test waits a few seconds and verifies that each cluster was only added once to the reconciliation repository

_Case 2: Worker pools are trying to pick the same operation in parallel_ 
1. Enhance unit-test of reconciliation repository: ensure that an parallel update of the same operation is only successful once
2. Implement race-condition integration test:
     1. Start multiple worker pools in parallel which are watching the inventory with different intervals
     2. Add a few reconciliation+operation entries to the reconciliation repository (with some delay between the creation)
    3. Test waits a few seconds: each operation should only be picked up by one worker, all other workers have to fail

_Case 3: Bookkeepers try to mark an operation as orphan in parallel_ 
1. Enhance unit-test of reconciliation repository: ensure that an parallel finish of an reconciliation is only successful once
2. Enhance unit-test of bookkeeper to verify behaviour in case of race-condition errors (when updating an operation which was already updated by another bookkeeper): 
    * test method `markOrphanOperations` in an multithreaded scenario and verify that an operation gets only updated once to `orphan` state
    * test method `finishReconciliation` within an multithreaded scenario and ensure that an a reconcilation gets only finished (incl. corresponding cluster update) once

**Reasons**

Detect race conditions caused by parallel running mothership reconcilers.

**Attachments**

PR #436, opened 2021-11-18
Enhance automated tests due to race-conditions incl. DB Fix

jeremyharisch:
**Description**

This PR enhances multiple unit-test due to race conditions. Furthermore, it fixes a bug for the inMemory and persistent database implementation.

**Changes**

New Test-Cases:
- reconciliation repository:
    -  Create multiple instances of single reconciliations in parallel; Only one should be successful
    - Update single reconciliation in multiple parallel threads; Only one should be successful
    - Mark single reconciliation as finished in multiple parallel threads; Only one should be successful
- bookkeeper:
    - Mark two operations as orphan in multiple parallel threads; Each operation should only be updated once as an orphan
    - Finish two operations as orphan in multiple parallel threads; Each operation should only be updated once as an done
- Workerpool:
    - Multiple WorkerPools watching same reconciliation repository; Each operation should only be assigned to one worker / should only be invoked once
- Scheduler:
    - Multiple scheduler watching same inventory; For each cluster reconciliations should be only created once
 
Refactor:
- Introduction of `NewCluster ` function and `TestCluster` type, to reduce code duplication in unit tests; adapt testcase due to this
- Wrap errors in `factory.go`, `sqlite.go` for better error log output
- Introduction of `IsAlreadyInStateError`
- Refactor of bookkeeper: Introduction of `BookkeepingTask` to establish `Chain of Responsibility` pattern to have a better testable code (including written unit tests)

Fix:
- Increase test timeout to 20min, since Pipelines take longer than 15min, local execution takes around 3,5min
- Change host in unittest config to `127.0.0.1`, due to IPv6 issues
- In update call (persistent and inmemory): Check if new wanted state is same as current state; If yes update should fail
- `Invoke` of remoteInvoker called `i.updateOperationState(params, model.OperationStateClientError, errorReason)` twice; but it should only be called once
- Adapt Remote Invoker test due to DB  and remote invoker fix

**Related issue**
https://github.com/kyma-incubator/reconciler/issues/374


pkg/scheduler/service/bookkeeper_test.go
```diff
@@ -85,3 +92,116 @@ func TestBookkeeper(t *testing.T) {
...
+				require.NoError(t, err)
+				for _, opEntity := range opEntities {
+					err := reconRepo.UpdateOperationState(opEntity.SchedulingID, opEntity.CorrelationID, model.OperationStateDone)
+					require.NoError(t, err)
+				}
+			}
+
+			//initialize logger including error channel
+			errChannel := make(chan error, 100)
+			bookkeeperLogger := zaptest.NewLogger(t, zaptest.WrapOptions(zap.Hooks(func(e zapcore.Entry) error {
```
tobiscr, line 160: General a good idea to inject some interception logic into another object - but failure prone if someone changes latter the log behaviour.

If a function has to be tested but it's not possible to test it, it indicates a bad testable code base which requires a refactoring.

Would recommend following:

* the Bookkeeper has already too many responsibilities bundled in one place: let's introduce a responsibility pattern and extract the Orphan-Ops detection and Finish-reconciliation logic dedicated chain-steps.
* Each step can be implemented in way that it becomes testable.

pkg/scheduler/reconciliation/reconciliation_test.go
```diff
@@ -729,3 +729,93 @@ func dbConnection(t *testing.T) db.Connection {
...
+			go func() {
+				time.Sleep(time.Until(startAt))
+				err = repo.UpdateOperationState(recon.SchedulingID, allOperations[0].CorrelationID, model.OperationStateError, "")
+				if err != nil {
+					errChannel <- err
+				}
+			}()
+		}
+		time.Sleep(5 * time.Second)
+		require.Equal(t, 49, len(errChannel))
```
tobiscr, line 790: Would be good to verify that the operation was really updated also in case of race-conditions.

pkg/scheduler/reconciliation/reconciliation_test.go
```diff
@@ -729,3 +729,93 @@ func dbConnection(t *testing.T) db.Connection {
...
+			go func() {
+				time.Sleep(time.Until(startAt))
+				_, err := repo.CreateReconciliation(mockClusterState, nil)
+				if err != nil {
+					errChannel <- err
+				}
+			}()
+		}
+		time.Sleep(5 * time.Second)
+		require.Equal(t, 49, len(errChannel))
```
tobiscr, line 759: We should also check that the reconciliation was really created in case of race-conditions.

pkg/scheduler/reconciliation/reconciliation_test.go
```diff
@@ -729,3 +729,93 @@ func dbConnection(t *testing.T) db.Connection {
...
+			go func() {
+				time.Sleep(time.Until(startAt))
+				err = repo.FinishReconciliation(recon.SchedulingID, mockClusterState.Status)
+				if err != nil {
+					errChannel <- err
+				}
+			}()
+		}
+		time.Sleep(5 * time.Second)
+		require.Equal(t, 49, len(errChannel))
```
tobiscr, line 819: Please also quickly verify that the reconciliation was updated to be finished.

We forgot one thing to verify: the the public functions of a transition (see [transition.go](https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/service/transition.go))should also be checked to handle race-conditions properly (sorry, forgot that part)

tobiscr (changes requested):

configs/reconciler-unittest.yaml
```diff
@@ -6,7 +6,7 @@ db:
   blockQueries: true
   logQueries: true
   postgres:
-    host: "localhost"
+    host: "127.0.0.1"
```
ruanxin, line 9: just by curiosity, normally these two values should be synonym, DNS will handle the look up from localhost to IP 127.0.0.1 correctly, do we have a specific reason to change?

jeremyharisch: It seems like, we have an IPv6 problem here, that some times localhost will not be properly translated to 127.0.0.1

pkg/scheduler/reconciliation/reconciliation_test.go
```diff
@@ -729,3 +729,93 @@ func dbConnection(t *testing.T) db.Connection {
...
+		mockClusterState, _ := createClusterStates(t, inventory)
+
+		defer func() {
+			require.NoError(t, inventory.Delete(mockClusterState.Cluster.RuntimeID))
+		}()
+
+		startAt := time.Now().Add(1 * time.Second)
+		for i := 0; i < 50; i++ {
+			go func() {
+				time.Sleep(time.Until(startAt))
```
ruanxin, line 751: If I understood correctly, you want to prevent the follow up code running, and wait for the next thread started. But `Sleep` will happen at each round within this for loop, which means it pause 3 time.

In single processors, there is no way you can make sure the created multiple threads started at the exactly same time, it's not how CPU works, CPU running sequentially, thread just win some CPU time slice when it's run, 
like this: https://en.wikipedia.org/wiki/Multithreading_(computer_architecture)#/media/File:Multithreaded_process.svg

If you really want to achieve parallel, since now days we all using multi core CPU, probably with some multiprocessors programming technique, you can configure thread running at each core at same time, but it's not controlled by our programmer with go code in the application level, it's down to OS level already.
https://stackoverflow.com/questions/20993139/how-does-goroutines-behave-on-a-multi-core-processor

pkg/scheduler/reconciliation/reconciliation_test.go
```diff
@@ -729,3 +729,93 @@ func dbConnection(t *testing.T) db.Connection {
...
+				_, err := repo.CreateReconciliation(mockClusterState, nil)
+				if err != nil {
+					errChannel <- err
+				}
+			}()
+		}
+		time.Sleep(5 * time.Second)
```
ruanxin, line 758: We can also use wait group here to eliminate the uncertainty of running time (the time mainly in spinning up database for you test, it depends on many hardware conditions).

pkg/scheduler/service/bookkeeper_test.go
```diff
@@ -85,3 +92,116 @@ func TestBookkeeper(t *testing.T) {
...
+			var wg sync.WaitGroup
+
+			//create mock database connection
+			dbConn := db.NewTestConnection(t)
+			//prepare inventory
+			inventory, err := cluster.NewInventory(dbConn, true, cluster.MetricsCollectorMock{})
+			require.NoError(t, err)
+
+			//add cluster to inventory
+			clusterState, err := inventory.CreateOrUpdate(1, &keb.Cluster{
```
ruanxin, line 119: try to create a function or factory class to initialize the test dummy data, so that we can remove this big chunk of code in test body.

Actually, in `pkg/cluster/inventory_test.go:302`, you can already find a function which can create dummy cluster, the question is how we reference it.

pkg/scheduler/service/bookkeeper_test.go
```diff
@@ -85,3 +92,116 @@ func TestBookkeeper(t *testing.T) {
...
+			require.NoError(t, err)
+			reconResult, err := bk.newReconciliationResult(recons[0])
+			require.NoError(t, err)
+			reconResult.orphanTimeout = 0 * time.Microsecond
+
+			//call markOrphanOperations in parallel threads
+			startAt := time.Now().Add(2 * time.Second)
+			for i := 0; i < 50; i++ {
+				wg.Add(1)
+				go func() {
```
ruanxin, line 184: Normally for anonymous function, if the logic is simple and just a few lines, we use it for sake of convenience.

In this case, we should create a function with a proper name for it, and use it here, it's self-explanatory and reduces the test body.

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -87,3 +99,126 @@ func TestWorkerPool(t *testing.T) {
...
+
+	t.Run("Multiple WorkerPools watching same reconciliation repository", func(t *testing.T) {
+		//make sure persistent db is empty
+		err := os.RemoveAll("./test/")
+		require.NoError(t, err)
+
+		//initialize WaitGroup
+		var wg sync.WaitGroup
+		//prepare keb clusters
+		kebClusters := []*keb.Cluster{
```
ruanxin, line 109: Let's reduce this big block, we can create 3 dummy cluster one by one using `pkg/cluster/inventory_test.go:302`

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -17,16 +20,25 @@ import (
...
+	}
+	i.mux.Lock()
 	i.params = append(i.params, params)
+	i.mux.Unlock()
 	return nil
 }
 
 func TestWorkerPool(t *testing.T) {
-	test.IntegrationTest(t) //required because a valid Kubeconfig is required to create test cluster entry
+	//test.IntegrationTest(t) //required because a valid Kubeconfig is required to create test cluster entry
```
ruanxin, line 40: As far as I understood, this function is a helper to control if needed to run it as an integration test. Although the way how we handle the distinguish between unit and integration test can be challenged, to remove it is not a good idea at the moment.

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -17,16 +20,25 @@ import (
 )
 
 type testInvoker struct {
-	params []*invoker.Params
+	params     []*invoker.Params
+	reconRepo  reconciliation.Repository
```
ruanxin, line 23: a struct is a collection of data, but the data should meaningfully belong to this collection. Reconciliation repository should not belong to the invoker, and let invoker carry Mutex and Channel is also not a good solution here.

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -56,19 +68,19 @@ func TestWorkerPool(t *testing.T) {
 		require.NoError(t, inventory.Delete(clusterState.Cluster.RuntimeID))
 	}()
 
+	//create test invoker to be able to verify invoker calls
+	testInvoker := &testInvoker{}
+
 	//create reconciliation for cluster
-	reconRepo := reconciliation.NewInMemoryReconciliationRepository()
-	reconEntity, err := reconRepo.CreateReconciliation(clusterState, nil)
+	testInvoker.reconRepo = reconciliation.NewInMemoryReconciliationRepository()
```
ruanxin, line 74: If you managed to remove reconRepo out of testInvoker, we don't have to change this test completely. Normally for the existing test, we should be really careful when changing it. If the related logic is not changed, the test should be prevented to change.

Unit tests is our first line of defense to save us to sleep well, keep it stable:)

configs/reconciler-unittest.yaml
ruanxin: aha, good to know, thanks!

pkg/scheduler/invoker/remoteinvoker_test.go
```diff
@@ -157,7 +157,9 @@ func TestRemoteInvoker(t *testing.T) {
 func invokeRemoteInvoker(reconRepo reconciliation.Repository, op *model.OperationEntity, cfg *config.Config) error {
 	//reset operation state
 	if err := reconRepo.UpdateOperationState(op.SchedulingID, op.CorrelationID, model.OperationStateNew); err != nil {
-		return err
+		if err.Error() != "cannot update state of operation 'CRDs' because it is already in state new" {
```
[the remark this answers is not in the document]
jeremyharisch: Done 👍🏻

pkg/scheduler/reconciliation/reconciliation_test.go
jeremyharisch: Done

pkg/scheduler/service/bookkeeper_test.go
jeremyharisch: Will refactor this

pkg/scheduler/service/bookkeeper_test.go
jeremyharisch: Lets talk about it ahgain, I would not change it

pkg/scheduler/worker/workerpool_test.go
jeremyharisch: Lets talk about it. The invoker needs the repo to update its status, as already mentioned last week in a small meeting

pkg/scheduler/worker/workerpool_test.go
jeremyharisch: Was removed by fault, will add it again

pkg/scheduler/worker/workerpool_test.go
jeremyharisch: Unfortunately, we need to reconRepo in the Invoker ti update the Status of the operations. The old Unit test did not cover such scenario and thus needed to be adapted

pkg/scheduler/service/bookkeeper_test.go
jeremyharisch: Done

pkg/scheduler/reconciliation/inmemory.go
```diff
@@ -213,6 +213,10 @@ func (r *InMemoryReconciliationRepository) UpdateOperationState(schedulingID, co
 		return &repository.EntityNotFoundError{}
 	}
 
+	if err := operationAlreadyInState(op, state); err != nil {
```
tobiscr, line 216: It turned out that we have also cases in which an update of an operation is required even if it's already in state: the heartbeat messages are continuously updating an operation (even when its state hasn't changed).

Maybe we add a flag to the `UpdateOperationState` to allow updates even if the state hasn't changed. The only case in which the update ins't allowed would be when a worker tries to pick an operation. Otherwise, updating exists states would be acceptable.

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -358,6 +358,10 @@ func (r *PersistentReconciliationRepository) UpdateOperationState(schedulingID,
 			return err
 		}
 
+		if err := operationAlreadyInState(op, state); err != nil {
```
tobiscr, line 361: see comment above

pkg/scheduler/service/bookkeepingtask.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+	"time"
+)
+
+type BookkeepingTask interface {
+	Apply(*ReconciliationResult) []error
+}
+
+type orphanOperation struct {
```
tobiscr, line 15: To make it's purpose more self-explaining, it could be worth to rename it to `markOrphanOperations` or similar.

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -87,3 +84,77 @@ func TestWorkerPool(t *testing.T) {
...
+				err := workerPool.Run(ctx)
+				if err != nil {
+					errChannel <- err
+				}
+			}(errChannel, workerPool)
+		}
+		wg.Wait()
+
+		//verify that invoker was properly called
+		require.Equal(t, 12, len(testInvoker.errChannel))
```
tobiscr, line 153: Would be good to add a comment why we check for 12 to make it more obvious for colleagues who don't know all details of the execution flow

tobiscr (changes requested):

pkg/db/factory.go
```diff
@@ -17,7 +18,7 @@ func NewConnectionFactory(configFile string, migrate bool, debug bool) (Connecti
 
 	encKey, err := readEncryptionKey()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error reading encryption key")
```
tobiscr, line 21: Would recommend to enhance directly the error in `readEncryptionKey` instead wrapping it here

tobiscr (changes requested):

pkg/db/sqlite.go
```diff
@@ -117,18 +117,18 @@ func (scf *sqliteConnectionFactory) Init(_ bool) error {
 		//read DDL (test-table structure)
 		ddl, err := ioutil.ReadFile(scf.schemaFile)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "error reading file")
```
tobiscr, line 120:
```suggestion
			return errors.Wrapf(err, "error reading file DDL schema file '%s'", scf.schemaFile)
```

pkg/scheduler/reconciliation/reconciliation_test.go
```diff
@@ -729,3 +681,131 @@ func dbConnection(t *testing.T) db.Connection {
...
+		//initialize WaitGroup
+		var wg sync.WaitGroup
+		//reset db connection
+		dbConn = nil
+
+		repo := newPersistentRepository(t)
+		inventory, err := cluster.NewInventory(db.NewTestConnection(t), true, cluster.MetricsCollectorMock{})
+		require.NoError(t, err)
+
+		errChannel := make(chan error, 100)
```
tobiscr, line 697: Make worth to use for  the lengh of this channel and for the loop-counter a variable to point out that both belong together

pkg/scheduler/reconciliation/reconciliation_test.go
```diff
@@ -729,3 +681,131 @@ func dbConnection(t *testing.T) db.Connection {
...
```
tobiscr, line 734: See above.. would recommend to use a variable for the length and for the loop to show that both belong together

pkg/scheduler/reconciliation/reconciliation_test.go
```diff
@@ -729,3 +681,131 @@ func dbConnection(t *testing.T) db.Connection {
...
```
tobiscr, line 780: see above regarding the variable

pkg/scheduler/reconciliation/reconciliation_test.go
```diff
@@ -729,3 +681,131 @@ func dbConnection(t *testing.T) db.Connection {
 	}
 	return dbConn
 }
+
+func TestReconciliationParallel(t *testing.T) {
+
+	t.Run("Create multiple instances of single reconciliations in parallel", func(t *testing.T) {
```
tobiscr, line 687: As this test case  and 'Update single operation state in multiple parallel threads'  and 'Mark single reconciliation as finished in multiple parallel threads' are having the same body, it would be better to use one method for it and pass the changing code as closure variable to it.

tobiscr (changes requested):

pkg/scheduler/reconciliation/reconciliation_test.go
jeremyharisch: 👍🏻

pkg/db/sqlite.go
jeremyharisch: Done

pkg/db/factory.go
jeremyharisch: Was just there for debugging. Wrapping is removed since error of called function has already a good user feedback

pkg/scheduler/reconciliation/reconciliation_test.go
jeremyharisch: Done

pkg/scheduler/reconciliation/inmemory.go
jeremyharisch: Done

pkg/scheduler/reconciliation/persistent.go
jeremyharisch: Done

PR #449, opened 2021-11-22
Support parallel pre-components for installation / deletion

tobiscr:


configs/reconciler.yaml
```diff
@@ -27,6 +27,4 @@ mothership:
       base:
         url: "http://localhost:8081/v1/run"
     preComponents:
-      - cluster-essentials
-      - istio-configuration
-      - certificates
+      - [cluster-essentials, istio-configuration, certificates]
```
suleymanakbas91, line 30: Certificates component is using a [Gateway CRD](https://github.com/kyma-project/kyma/blob/main/resources/certificates/templates/gateway.yaml  ) which is installed by istio. So, they cannot be installed at the same time.

suleymanakbas91 (changes requested):

configs/reconciler.yaml
suleymanakbas91: I also see there is a [service monitor](https://github.com/kyma-project/kyma/blob/main/resources/istio-configuration/templates/monitoring/service-monitor.yaml) in istio that is installed by cluster-essentials..

tobiscr: Feedback from @pbochynski  and @strekm was to running all in parallel works out. Also, my local tests were looking promising.

suleymanakbas91 (approved):

Issue #444, opened 2021-11-20
Error of missing GIT-revision has to include the repo-URL


tobiscr:
**Description**

If the download logic is cloning a component from a GIT repository and the repository doens not include the given version (=can be a revision, branch- or tag-name), the error message just points out that the version doens not exist. The error message should also include the repo-URL to make it clear, which repo was used.

Current error message looks like:

     Failed to get manifest for component 'connectivity-proxy' in Kyma version '2.0.0-rc6': failed to resolve GIT revision '2.0.0-rc6': reference not found

**Expected result**

Error message should also include the repo URL, like (just as example)

     All attempts fail:\n#1: Failed to get manifest for component 'connectivity-proxy' in Kyma version '2.0.0-rc6' using repository 'https://repourl.git': failed to resolve GIT revision '2.0.0-rc6': reference not found\n#

**Actual result**

Error message just shows that revision is missing, but doens't include a hint which repo was used.

**Steps to reproduce**

**Troubleshooting**

PR #452, opened 2021-11-23
Error of missing GIT-revision has to include the repo-URL

rJankowski93:
https://github.com/kyma-incubator/reconciler/issues/444


tobiscr (approved):

pkg/reconciler/service/install.go
```diff
@@ -89,6 +89,10 @@ func (r *Install) renderManifest(chartProvider chart.Provider, model *reconciler
 	if err != nil {
 		msg := fmt.Sprintf("Failed to get manifest for component '%s' in Kyma version '%s'",
 			model.Component, model.Version)
+		if model.URL != "" {
```
tobiscr, line 92: I think we should also enhance the log-messagse in `git/clone.go` when it'S complaninig about the not found revision:  https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/git/clone.go#L79 (hope hte URL is accessbiel over the `repo` struct).

tobiscr (changes requested):

pkg/reconciler/service/install.go
rJankowski93: you're right. I added this msg in clone.go

tobiscr (approved):

PR #455, opened 2021-11-23
Default Main Version for Components with URL

ralikio:
Change default branch for external components with external url to "main".

### DIFF
--- a/pkg/scheduler/invoker/invoker.go
+++ b/pkg/scheduler/invoker/invoker.go
@@ -10,6 +10,11 @@ import (
 	"github.com/kyma-incubator/reconciler/pkg/reconciler"
 )
 
+const (
+	defaultBranch = "main"
+	tokenKey      = "repo.token.namespace"
+)
+
 type Invoker interface {
 	Invoke(ctx context.Context, params *Params) error
 }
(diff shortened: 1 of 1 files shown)


pkg/scheduler/invoker/invoker.go
```diff
@@ -36,7 +40,9 @@ func (p *Params) newRemoteTask(callbackURL string) *reconciler.Task {
 
 func (p *Params) newTask() *reconciler.Task {
 	version := p.ClusterState.Configuration.KymaVersion
-	if p.ComponentToReconcile.Version != "" {
+	if p.ComponentToReconcile.URL != "" && p.ComponentToReconcile.Version == "" {
+		version = defaultBranch
+	} else if p.ComponentToReconcile.Version != "" {
 		version = p.ComponentToReconcile.Version
```
tobiscr, line 46: We should actually not set a default here (at least my opinion): we cannot know which naming convention is used in an external GIT repository (maybe they use `master` as their mainstream branch etc.).

I would recommend to make the version mandatory for external components hosted in a GIT repository and not to use any default here.

In this case: the version should be provided as part of the component configuration.

tobiscr: Please set the version only for GIT urls.

tobiscr (changes requested):

tobiscr (approved):

Issue #443, opened 2021-11-19
Download GIT repositories have to consider the revision-ID when caching sources


tobiscr:
**Description**

When cloning an external component in the workspace factory, the revision-ID should be considered by the caching mechanism (currently the given version, branch tag etc.) is used which is non-unique for GIT (code in a  branch can change over time without causing URL changes etc.).

AC:
* Use revision ID in the workspace-directory name to ensure changes in a branch will trigger another clone of the repository
* Implement unit-test which clones a repo and uses the same branch-name as version which gets converted to it's head revision

**Reasons**

Identify changes in a GIT branch properly to detected outdated workspace directoryies.

**Attachments**

tobiscr:
Agreed on following approach:

<img width="854" alt="Screen Shot 2021-11-24 at 12 34 17" src="https://user-images.githubusercontent.com/73824711/143406970-4cfec774-710b-42d8-b7d1-a56fc2f6d75b.png">

PR #457, opened 2021-11-23
Use revisionID instead of version for git based external components

moelsayed:
#443


pkg/reconciler/git/clone.go
```diff
@@ -156,3 +159,29 @@ func mapSecretKey(URL string) (string, error) {
 
 	return output, nil
 }
+
+func (r *Cloner) ResolveRevision(rev string) (*gitp.Hash, error) {
+	auth, err := r.buildAuth()
```
tobiscr, line 164: Is there any option to get a test-case for it?

pkg/reconciler/git/clone.go
```diff
@@ -156,3 +159,29 @@ func mapSecretKey(URL string) (string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	// in case this is called before the repo is cloned, we must have a clone for the resolver to work.
+	repo := r.repoClient.Repo()
+	if repo == nil {
+		repo, err = git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
```
tobiscr, line 171: Means, in the current implementation we are cloning always twice, because we are first retreiving the revision and afterwards triggering the real clone, right?

tobiscr: Let's have tomorrow a quick sync, as we agreed today to make the version optional. I think we can easily introduce this new behaviour into your PR which will also fix the side-effect we currently have with resolving first the revision and cloning the sources later again.

tobiscr:
Please see also https://github.com/kyma-incubator/reconciler/issues/462 - maybe worth to handle both in one change as it's quite close related to each other

PR #471, opened 2021-11-24
Set HSTS Header for  SendHttpError

jeremyharisch:


jeremyharisch:
HSTS-Header is not needed since our services are communicating over `http`. Furthermore, the `HSTS` Header is just used by web-browser, and thus not needed anyways.

PR #477, opened 2021-11-25
Revert "Update istio to 1.12.0"

dariusztutaj:
Reverts kyma-incubator/reconciler#461


strekm (approved):
/meow

strekm (commented):
we decided to revert it as we are having some troubles with bumping istio-configuration chart. i want to avoid situation that this change is blocking upgrade old leads to any issues with istio.

suleymanakbas91 (approved):

Issue #462, opened 2021-11-24
Workspace factory: make version optional for GIT hosted external resources


tobiscr:
**Description**

KEB and reconciler team agreed to make the version for GIT hosted external components option.

AC:

The behaviour of the workspace factory has to be aligned to handle GIT hosted external components in following way:

* If the version for a GIT hosted component is provided, use this version
* If the version is missing, use the default-branch of the GIT repository
* The workspace-directory has still to include the resolved revision-ID (not the plain version-name!) in the folder name to ensure that we are re-cloning the workspace if new commits were added to the GIT repository.
* An integration-test is implemented which is cloning a GIT hosted source
     * if no version is provided, it is using the default-branch of the repository
     * in case a version is provided, it using the provided version
     * in all cases => the workspace folder includes the revision-ID and not just the plain version

**Reasons**

Align behaviour between KEB and reconciler when referring to GIT hosted external components.

**Attachments**

tobiscr:
@moelsayed : JFY - after version is optional, please drop the hack to set the version to "main" in case of GIT urls (see https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/invoker/invoker.go#L44-L46)

Issue #432, opened 2021-11-17
Git-client should support git-V2 protocol and mulit-ack to be compatible with Azure DevOps


tobiscr:
**Description*** 

Goal is to simlify our internal GIT library in the reconciler. Currently, the interface is not well design and not offering a meaningful simplification of the underneath used go-git library.

Refactoring should:

* Verify what the required GIT use-cases are the reconciler has (probably support of converting branches to revision, clone and fetch GIT repos)
* Reduce the API to support these type of GIT actions
* Remove the dependnecy to K8s-API as it's coupling the GIT library to K8s runtimes which makes technially no sense (instead doing credential-lookups in GIT by reading K8s-secrets, pass the credentials to the GIT API cal)

Last nice-to-have feature would be to support GIT v2 protocol:

The used `go-git` library [isn't support git v2 protocol and also not handling multi-ack](https://github.com/go-git/go-git/issues/64) which makes it incompatible with common external products like Azure DevOps. This limitation can become a drawback for open source customers and we should replace the used go-GIT library with a newer implementation (see linked ticket for proposed alternatives).

Verify the effort of changing to a different go-GIT implementation which doens't have such limitations.

AC:

* Simplify GIT API to the use cases the reconciler requires (clone, fetch, convert automatically branch to revs etc.)
* Remove dependency to K8s library

Nice to have:
* Replace used git-library with an alternative GIT implementation which offers the same features as currently required in the reconciler (git clone, checkout particular branch etc.)
* Ensure implementation works with Github + Azure DevOps

**Reasons**

Simplify GIT library and provide an interface which is focusing on the required use-cases. Improve maintainability of the internal GIT lib.

**Attachments**

janmedrek:
Functionalities of the Reconciler were migrated to the new architecture, this issue will not be worked on.

PR #481, opened 2021-11-25
Refactor external git components

moelsayed:
#443
https://github.com/kyma-incubator/reconciler/issues/462


pkg/reconciler/git/client.go
```diff
@@ -26,6 +26,81 @@ func (d *Client) Worktree() (*git.Worktree, error) {
...
+	}
+	if branchRev != nil { // this is a branch
+		return d.repo.ResolveRevision(gitp.Revision(branchRev.String()))
+	}
 	return d.repo.ResolveRevision(rev)
 }
+
+func (d *Client) Fetch(path string, o *git.FetchOptions) error {
+	var err error
+	if d.repo == nil {
```
m00g3n, line 57: A couple of questions here:
- Why fetch method has a `path` argument?
- Why this methods opens the repository from the `path`?
- What the user of this method should expect if he will provide both: `a repository` and `a path`?

pkg/reconciler/git/client.go
```diff
@@ -26,6 +26,81 @@ func (d *Client) Worktree() (*git.Worktree, error) {
...
+	err = d.repo.Fetch(o)
+	if err == git.NoErrAlreadyUpToDate {
+		return nil
+	}
+	return err
+}
+
+func (d *Client) PlainCheckout(path string, o *git.CheckoutOptions) error {
+	var err error
+	if d.repo == nil {
```
m00g3n, line 72: What the user of this method should expect if he will have both: `a repository` and a `a path`?

moelsayed: Repository will take precedence. Since the the repository is not passed, it's only initialized if this client instance is used to clone the repo from upstream. In which case we don't need to "open" it again.

pkg/reconciler/git/client.go
moelsayed: 1- The path is used in case the client object is not initialized with a repo already.
2- Because `fetch` requires a cloned local copy of the repo. We need to "open" the local repo cache. 
3- Repository will take precedence. Since the the repository is not passed, it's only initialized if this client instance is used to clone the repo from upstream. In which case we don't need to "open" it again.

pkg/reconciler/chart/factory.go
```diff
@@ -326,3 +360,106 @@ func (f *DefaultFactory) Delete(version string) error {
...
+		return err
+	}
+	cloner, _ := git.NewCloner(&git.Client{}, repo, true, clientSet, f.logger)
+
+	return cloner.FetchAndCheckout(dstPath, component.version)
+}
+
+// getLatestRevOfVersion works on local cache. If "version" is a branch, it will retrun the revisionID of it's HEAD
+// if "version" is a tag, it will return it's revisionID.
+func (f *DefaultFactory) getLatestRevOfVersion(version, path string) (string, error) {
```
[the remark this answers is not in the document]
moelsayed: Unfortunately no. This function uses a "local" git client since it only needs to open the local cache. Both `clone` and `fetch` need to have access to the upstream with a remote client and auth. I believe it would be better to keep the logic here separated.

pkg/reconciler/chart/factory.go
```diff
@@ -326,3 +360,106 @@ func (f *DefaultFactory) Delete(version string) error {
...
+		f.logger.Warnf("Deleting workspace '%s' because previous download/clone does not contain all the required files", baseDir)
+		if err := os.RemoveAll(baseDir); err != nil {
+			return false, err
+		}
+
+	}
+	return false, nil
+}
+
+func (f *DefaultFactory) fetchComponent(component *Component, dstDir string) error {
```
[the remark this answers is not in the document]
moelsayed: check below

pkg/reconciler/chart/factory.go
```diff
@@ -326,3 +360,106 @@ func (f *DefaultFactory) Delete(version string) error {
 	}
 	return err
 }
+
+func (f *DefaultFactory) componentBaseDir(c *Component) string {
+	return filepath.Join(f.storageDir, gitComponentsBaseDir,
+		fmt.Sprintf("%x-%s", sha1.Sum([]byte(c.url)), c.name)) //nolint
+}
+
+func (f *DefaultFactory) indicatorExistsOrClean(baseDir string) (bool, error) {
```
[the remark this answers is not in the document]
moelsayed: done.

pkg/reconciler/chart/factory.go
```diff
@@ -1,19 +1,23 @@
...
+	"strings"
 
 	"os"
 
+	gogit "github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/kyma-incubator/reconciler/pkg/reconciler"
 	reconcilerK8s "github.com/kyma-incubator/reconciler/pkg/reconciler/kubernetes"
 	"github.com/mholt/archiver/v3"
+	"github.com/otiai10/copy"
```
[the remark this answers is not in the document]
moelsayed: I didn't want to add a moby dependency due to security issues. This is a smaller code base with a specific scope.

pkg/reconciler/git/client.go
tobiscr: The git library need in general a "refurbishment" as it's already bad designed. The `Client` struct is in general unnecessary I think.

Maybe better to fail always when `d.repo == nil` with an unexpected state error.

And initialising the repo has to happen upfront by e.g. calling `NewRepositoryFromPath(path string) (*git.Repository, error)` and return the *git.Repository as result.

pkg/reconciler/git/client.go
```diff
@@ -26,6 +26,81 @@ func (d *Client) Worktree() (*git.Worktree, error) {
...
+	}
+	w, err := d.repo.Worktree()
+	if err != nil {
+		return err
+	}
+	return w.Checkout(o)
+}
+
+func NewClientWithPath(path string) (*Client, error) {
+	var err error
```
tobiscr, line 86: Shouldn't it return `ClientRepo` ?

pkg/reconciler/git/client.go
```diff
@@ -26,6 +26,81 @@ func (d *Client) Worktree() (*git.Worktree, error) {
...
+	d := &Client{}
+	d.repo, err = git.PlainOpen(path)
+	return d, err
```
tobiscr, line 89: Would do following:

     repo, err := git.PlainOpen(path)
     if err != nil {
        return nil, err
     }
     return &Client{repo:repo}

tobiscr (changes requested):

pkg/reconciler/git/client.go
tobiscr: => follow up action to clean up the GIT library in general is https://github.com/kyma-incubator/reconciler/issues/432

pkg/reconciler/git/client.go
tobiscr: would recommend to use the Go best-practise by using the `New...` fucntion which allows to specify the repo-path just once. This will simplify the usage of the API as it's clear that the path can be defined only once and the repo will be coupled to this path.

m00g3n (approved):

tobiscr (approved):

Issue #465, opened 2021-11-24
Support PATCH strategy for Kubernetes resources which cannot be updated via REPLACE


tobiscr:
**Description**

Replacing resources in Kubernetes does not work for few resource types (e.g. Jobs, Pod, PVCs and Statefulsets which include PVCs, [Service-Accounts](https://github.com/kyma-incubator/reconciler/issues/467)). These resources have to be updated using Patching.

AC:

* The reconciler has to verify whether the current resource is of type PVC or Statefulset and includes a PVC. In this case, the update of this resource has to happen by  the Kubernetes client via PATCH instead REPLACE.
* The Interceptors which are excluding these resources from reconciliation have to be dropped after we support patching of such resources.

**Reasons**

Support PATCHING  for resources which cannot be REPLACE in Kubernetes to support a reconciliation of these resources.

**Attachments**

tobiscr:
@pbochynski : QQ - should we exclude resource we intentionally don'T want to reconcile from the scope of this ticket (e.g. Pods, Jobs)?

pbochynski:
I would exclude pods and jobs.

veichtj:
Unfortunately retest of #459 failed, after changing the postgres image within ory there is no update happening after reconciling ory

pbochynski:
@cnvergence  Can you contact @tobiscr and reproduce together the issue?

varbanv:
Still troubleshooting.

tobiscr:
Fix provided in #535 (switching o HELM kube-client)

raypinto:
Retested #459 - after changing the postgres image within ory, the `ory-postgresql-0` pod reconciles successfully!

raypinto:
Retested #459 with the latest reconciler - and the change in postgres image within ory does not reconcile `ory-postgresql-0` pod

tobiscr:
I'm try install ORY on my local K3d and it's not possible for me. It stucks starting `ory-hydra-...` pods

     kyma-system    ory-hydra-maester-5667c9dc6-2ghrh                       0/3     PodInitializing   0          2m2s
     kyma-system    ory-hydra-765b558996-88h68                              0/2     PodInitializing   0          2m2s

tobiscr:
Issue with ORY should be fixed with #597

PR #484, opened 2021-11-26
Patch resources that cannot be replaced

kwiatekus:
**Related issues**
https://github.com/kyma-incubator/reconciler/issues/465


pkg/reconciler/kubernetes/internal/kubeclient.go
```diff
@@ -132,24 +137,32 @@ func (kube *KubeClient) ApplyWithNamespaceOverride(u *unstructured.Unstructured,
...
 		return metadata, nil
 	}
 
-	replace := newReplace(helper)
-	replacedObject, err := replace(u, u.GetNamespace(), u.GetName())
-	if err != nil {
+	updateStrategyResolver := newDefaultUpdateStrategyResolver(helper)
+
+	strategy, err := updateStrategyResolver.Resolve(u)
+	if err != nil || strategy == SkipUpdateStrategy {
```
tobiscr, line 146: Why do we return an error when a SkipStragey was replied?

kwiatekus: I can separate those

kwiatekus: done

tobiscr (approved):

pkg/reconciler/kubernetes/internal/patcher.go
```diff
@@ -0,0 +1,256 @@
...
+		lookupPatchMeta strategicpatch.LookupPatchMeta
+		schema          oapi.Schema
+	)
+	// createPatchErrFormat := "creating patch with:\noriginal:\n%s\nmodified:\n%s\ncurrent:\n%s\nfor:"
+
+	// Create the versioned types from the type defined in the restmapping
+	// (which is the API version we'll be submitting the patch to)
+	versionedObject, err := scheme.Scheme.New(p.Mapping.GroupVersionKind)
+
+	switch {
```
m00g3n, line 132: I know this part was reverted, however we could consider a refactor of the entire switch to be easier to read and maintain (we could have a separate story for this).

m00g3n (approved):
Got just 2 comments regarding the `patcher` type but we can address them in a follow-up task.

Issue #407, opened 2021-11-14
Add retention mechanism for reconciliation data


tobiscr:
**Description**

Currently the tracked reconciliation data (reconciliation entities + operation entities) are stored endless. These data are only relevant for a particular time and should be removed after X days.

A retention mechanism has to be added which regularly runs and deletes these data completely from the database. The deletion has to include the reconciliation entities and operation entities. The retention time has to be configurable.

AC:
* Reconciliation mechanism verifies regularly the stored reconciliation and operation data
* Reconciliation entities (and their related operation entities) which are exceeding the retention time have to be deleted from the database (real deletion, no flagging).

**Reasons**

Avoid unnecessary storage consumption of reconciliation data / database house keeping.

**Attachments**

PR #487, opened 2021-11-26
Add retention mechanism

pPrecel:
Ticker: https://github.com/kyma-incubator/reconciler/issues/407

because of the need to remove old, not used reconciliation and operation entities I've added a new component run in the dedicated go-routine called `cleaner`.

### DIFF
--- a/cmd/mothership/mothership/start/scheduler.go
+++ b/cmd/mothership/mothership/start/scheduler.go
@@ -2,12 +2,13 @@ package cmd
 
 import (
 	"context"
+	"time"
+
 	"github.com/kyma-incubator/reconciler/pkg/logger"
 	"github.com/kyma-incubator/reconciler/pkg/scheduler/config"
 	"github.com/kyma-incubator/reconciler/pkg/scheduler/service"
 	"github.com/kyma-incubator/reconciler/pkg/scheduler/worker"
 	"github.com/spf13/viper"
-	"time"
 )
 
 func startScheduler(ctx context.Context, o *Options, configFile string) error {
@@ -41,6 +42,10 @@ func startScheduler(ctx context.Context, o *Options, configFile string) error {
 		WithBookkeeperConfig(&service.BookkeeperConfig{
 			OperationsWatchInterval: 30 * time.Second,
 			OrphanOperationTimeout:  o.OrphanOperationTimeout,
+		}).
+		WithCleanerConfig(&service.CleanerConfig{
+			PurgeEntitiesOlderThan: 14 * 24 * time.Hour, // 14 days
+			CleanerInterval:        12 * time.Hour,
 		}).
 		Run(ctx)
 }
--- a/cmd/mothership/mothership/start/scheduler.go
+++ b/cmd/mothership/mothership/start/scheduler.go
@@ -44,8 +44,8 @@ func startScheduler(ctx context.Context, o *Options, configFile string) error {
 			OrphanOperationTimeout:  o.OrphanOperationTimeout,
 		}).
 		WithCleanerConfig(&service.CleanerConfig{
-			PurgeEntitiesOlderThan: 14 * 24 * time.Hour, // 14 days
-			CleanerInterval:        12 * time.Hour,
+			PurgeEntitiesOlderThan: o.PurgeEntitiesOlderThan,
+			CleanerInterval:        o.CleanerInterval,
 		}).
 		Run(ctx)
 }
--- a/pkg/db/query.go
+++ b/pkg/db/query.go
@@ -131,7 +131,7 @@ type Select struct {
 
 func (s *Select) WhereRaw(stmt string, args ...interface{}) *Select {
 	s.addWhere()
-	s.buffer.WriteString(fmt.Sprintf(" (%s)", stmt))
+	s.buffer.WriteString(fmt.Sprintf(" %s", stmt))
 	s.args = append(s.args, args...)
 	return s
 }
--- a/pkg/db/query.go
+++ b/pkg/db/query.go
@@ -131,7 +131,7 @@ type Select struct {
 
 func (s *Select) WhereRaw(stmt string, args ...interface{}) *Select {
 	s.addWhere()
-	s.buffer.WriteString(fmt.Sprintf(" %s", stmt))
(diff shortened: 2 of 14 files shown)


pkg/scheduler/service/cleaner.go
```diff
@@ -0,0 +1,60 @@
...
+	reconciliations, err := transition.ReconciliationRepository().GetReconciliations(&reconciliation.WithCreationDateBefore{
+		Time: cretedBefore,
+	})
+	if err != nil {
+		c.logger.Error("Cleaner failed to get reconciliations older than: %s", cretedBefore)
+	}
+
+	for i := range reconciliations {
+		id := reconciliations[i].SchedulingID
+		err := transition.ReconciliationRepository().RemoveReconciliation(id)
```
pPrecel, line 55: Is this enough to remove the reconciliation entity and all connected operations? Or should we extend the `ReconciliationRepository` to add a method that would allow removing expected operations based on given filter (older than expected date)

pkg/scheduler/service/run.go
```diff
@@ -196,5 +206,13 @@ func (r *RunRemote) Run(ctx context.Context) error {
 		}
 	}()
 
+	//start cleaner
```
pPrecel, line 209: I'm not sure that the name cleaner is good enough for a component like this :D Any suggestion?

pkg/scheduler/service/cleaner.go
tobiscr: Let's start with this approach  - looks good 👍

pkg/scheduler/service/run.go
tobiscr: I like it! :)

pkg/scheduler/reconciliation/filter.go
```diff
@@ -2,11 +2,28 @@ package reconciliation
...
 	"github.com/kyma-incubator/reconciler/pkg/model"
 )
 
+type WithCreationDateBefore struct {
+	Time time.Time
+}
+
+func (wd *WithCreationDateBefore) FilterByQuery(q *db.Select) error {
+	q.WhereRaw("created<$1", wd.Time.Format("2021-11-26 00:00:00.000"))
+	return nil
```
tobiscr, line 17: Can we double check that it works with SQLite and Postgres? Not sure that both behave equal when processing timestamps.

pPrecel: Thank you for this question. I've tested this on the `SQLitez and the `PostgreSQL` on my local machine and the spell is working for both DBs.

example spell:
```
SELECT * FROM scheduler_reconciliations WHERE created<'2021-11-03 23:23:01';
```

cmd/mothership/mothership/start/scheduler.go
```diff
@@ -42,6 +43,10 @@ func startScheduler(ctx context.Context, o *Options, configFile string) error {
 			OperationsWatchInterval: 30 * time.Second,
 			OrphanOperationTimeout:  o.OrphanOperationTimeout,
 		}).
+		WithCleanerConfig(&service.CleanerConfig{
+			PurgeEntitiesOlderThan: 14 * 24 * time.Hour, // 14 days
```
tobiscr, line 47: Would be good to make the purge-interval configurable os CLI flag. Using 14days as default is good.

tobiscr (changes requested):

pkg/db/query.go
```diff
@@ -131,7 +131,7 @@ type Select struct {
 
 func (s *Select) WhereRaw(stmt string, args ...interface{}) *Select {
 	s.addWhere()
-	s.buffer.WriteString(fmt.Sprintf(" (%s)", stmt))
```
pPrecel, line 134: Are we ok with this change? I've changed it because it caused failing [this](https://github.com/kyma-incubator/reconciler/pull/487/files#diff-36c829107c81e04951311e0d023fe69e017531e811067f4c18cbf3e4cc21e8a9R22) validation

pkg/db/query.go
```diff
@@ -131,7 +131,7 @@ type Select struct {
...
+	s.buffer.WriteString(fmt.Sprintf(" %s", stmt))
```
pPrecel, line 134: Are we ok with this change? I've changed it because it caused failing [this](https://github.com/kyma-incubator/reconciler/pull/487/files#diff-36c829107c81e04951311e0d023fe69e017531e811067f4c18cbf3e4cc21e8a9R22) validation

pkg/db/validator.go
```diff
@@ -19,19 +19,19 @@ func NewValidator(blockQueries bool, logger *zap.SugaredLogger) *Validator {
 }
 
 func (v *Validator) Validate(query string) error {
-	matchSelect, err := regexp.MatchString("SELECT.*(FROM\\s*\\w+\\s*)+(WHERE (\\w*\\s*=\\s*\\$\\d+(\\s*,\\s*)?(\\s+AND\\s+)?(\\s+OR\\s+)?)*)?(\\w*\\s+IN\\s+[^;]+)?(\\w*\\s+ORDER BY\\s+[^;]+)?(\\w*\\s+GROUP BY\\s+[^;]+)?$", query)
+	matchSelect, err := regexp.MatchString("SELECT.*(FROM\\s*\\w+\\s*)+(WHERE (\\(?\\w*\\s*[=<>]\\s*\\$\\d+\\)?(\\s*,\\s*)?(\\s+AND\\s+)?(\\s+OR\\s+)?)*)?(\\w*\\s+IN\\s+[^;]+)?(\\w*\\s+ORDER BY\\s+[^;]+)?(\\w*\\s+GROUP BY\\s+[^;]+)?$", query)
```
pPrecel, line 22: Without this change, a few requests would not work. For example, queries with `<` or `>` instead of `=`. Additionally, I found a bug that causes that built queries like [here](https://github.com/kyma-incubator/reconciler/pull/487/files#diff-b70a62f2599a781583b295541d8528116e619114a361834082e2651a18234581R16) would not pass validation because of brackets

tobiscr (dismissed):

cmd/mothership/mothership/start/scheduler.go
pPrecel: Done!

tobiscr (approved):

Issue #464, opened 2021-11-24
Reconcile also clusters in error state


tobiscr:
**Description**

_WHY:_

To ensure that transient errors won't lead to non-reconciled clusters, it was agreed to not stop the reconciliation if a cluster is switching into an error state (e.g. `error`).

The reconciliation will continue and pickup the cluster regularly as part of the reoccurring reconciliation cycles. The only state in which a cluster will not be reconciled is:

* either a temporary state (e.g. `reconciling`, `deleting`)
* the cluster is disabled from the reconciliation (state `reconcile_disabled`) 
* the last reconciliation append before the reconciliation-interval was reached again 
* the cluster is in a final state (e.g. `deleted`, `delete_error`)

_AC:_

* Cluster which are in state `error` are still reconciled after they reached the configured reconcile-interval
* Number of re-tries should be counted
* Alerting for an cluster in error state should happen after the X retries happened (X should be configurable)
* ~~Retires should happen with increasing delay, e.g. 2, 4, 8, 16min.. etc. (the delay should be configurable)~~ <- descoped!
* Alerting has to stop after first successful reconciliation was possible

_Implementation hint:_

* Update the filter-logic in `inventory.GetClustersToReconcile()` function

**Reasons**

Address transient / temporary issues which can cause a cluster to switch into the error state by reconciling it.

**Attachments**

vvxxvvxx:
Hi Tobias, just to understand the current logic..
Now if the operation status is not `OperationStateDone` or `OperationStateError`, for every kyma component to be reconciled, they will be reconciled until timeout(default 10min) or MaxRetries (default 5), right? 
Status for all operations will be tracked and collected, then summarized to the ClusterStatus (9 statuses), right?


so from the above `AC`, only this one hasn't been implemented?  
- Retires should happen with increasing delay, e.g. 2, 4, 8, 16min.. etc. (the delay should be configurable)

tobiscr:
> Hi Tobias, just to understand the current logic.. Now if the operation status is not `OperationStateDone` or `OperationStateError`, for every kyma component to be reconciled, they will be reconciled until timeout(default 10min) or MaxRetries (default 5), right? Status for all operations will be tracked and collected, then summarized to the ClusterStatus (9 statuses), right?
> 
> so from the above `AC`, only this one hasn't been implemented?
> 
> * Retires should happen with increasing delay, e.g. 2, 4, 8, 16min.. etc. (the delay should be configurable)

[...]

PR #496, opened 2021-11-29
Reconcile Cluster which are in error State

jeremyharisch:
**Description**

- Introduction of new State `ClusterStatusReconcileErrorRetryable `
- `reconcileIntervalFilter ` now filters for `ClusterStatusReady ` and `ClusterStatusReconcileErrorRetryable` for reconciliation
- `ClusterStatusReconcileErrorRetryable` is a final State
- Default maximum retries for Cluster reconciliation is 150
- If retry count >= maximum retries, cluster gets Status `ClusterStatusReconcileError` and will not be retried

**Related Issue**
#464

### DIFF
--- a/pkg/scheduler/service/bookkeeper.go
+++ b/pkg/scheduler/service/bookkeeper.go
@@ -16,11 +16,13 @@ import (
 const (
 	defaultOperationsWatchInterval = 30 * time.Second
 	defaultOrphanOperationTimeout  = 10 * time.Minute
+	defaultMaxRetries = 150
 )
 
 type BookkeeperConfig struct {
 	OperationsWatchInterval time.Duration
 	OrphanOperationTimeout  time.Duration
+	MaxRetries int
 }
 
 func (wc *BookkeeperConfig) validate() error {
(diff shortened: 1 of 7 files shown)


pkg/scheduler/service/bookkeeper.go
```diff
@@ -131,6 +139,16 @@ func (bk *bookkeeper) finishReconciliation(reconResult *ReconciliationResult) bo
...
 	newClusterStatus := reconResult.GetResult()
 
+	if newClusterStatus == model.ClusterStatusReconcileError {
+		errCnt, err := bk.transition.inventory.CountRetries(reconResult.reconEntity.RuntimeID, reconResult.reconEntity.ClusterConfig)
+		if err != nil {
+			bk.logger.Errorf("failed to count error for runtime %s with error: %s", reconResult.reconEntity.RuntimeID, err)
+		}
+		if errCnt < bk.config.MaxRetries {
+			newClusterStatus = model.ClusterStatusReconcileErrorRetryable
+		}
```
tobiscr, line 149: Please add also an INFO log message that a cluster with runtimeID 'X' and clusterConfig 'Y' had a failing reconcilign but was retried because it had just 'Z' retries.

tobiscr (changes requested):

pkg/scheduler/service/bookkeeper.go
jeremyharisch: done

tobiscr (dismissed):

tobiscr (approved):

PR #509, opened 2021-11-30
Version Upgrade Check

ralikio:
Conditional update of reconciled component only if version in provided chart has changed.


franpog859 (commented):
It's ok for me `LGTM`. I tested it manually and the reconciler no longer reinstalls the Connectivity Proxy every time it's triggered, but it recognizes the same versions ✅ 

I'd love to see some other developer also approving this PR as I contributed to it and no one checked my code yet 😄

koala7659 (dismissed):
LGTM

pkg/reconciler/service/install.go
```diff
@@ -25,6 +26,36 @@ type Operation interface {
 	Invoke(ctx context.Context, chartProvider chart.Provider, model *reconciler.Task, kubeClient kubernetes.Client) error
 }
 
+//go:generate mockery --name=ManifestLookup --output=mocks --outpkg=mocks --case=underscore
+type ManifestLookup interface {
```
clebs, line 30: Do this interface and the `Lookup()` function need to be part of the install type?

This looks like specific logic for your component. The reconciler service should stay as simple as possible and only contain logic that is used by the base reconciler or all reconcilers equally.

Also if you have all this in the `connectivityproxy` package, you will only need review from your team.

clebs (changes requested):

pkg/reconciler/service/install.go
ralikio: I believe that it might be treated as a simple by universal way for searching for a specific component from in rendered manifest. Because it does not contain anything related to connectivity-proxy I would leave it as is. I would change invalid log message.

clebs: Yes this could be a general utility to search in the manifest, although I am concerned about the performance of it.

The function does a double conversion `unstructured` -> `manifest` -> `unstructured` which is very expensive (gets done later again to apply the resources). So `renderManifest` gets called twice.

Maybe there would be a better way by having `Lookup()` receive the finished manifest already instead of the unstructured map.

Then this could be called between `renderManifest` and `Apply`.

I think the idea is good but needs some more thinking on how to fit it into the architecture.

tobiscr (changes requested):

pkg/reconciler/service/install.go
franpog859: Hey! Let me jump in. What do you think about the solution I just pushed? There are no double conversions, no huge new functionalities in the `install.go`, just a bonus optional condition. I didn't want to move it straight to the `kubernetes` package as we do not operate on kubernetes resources here but only on manifests. **Didn't check it manually yet**, but I'm finishing my work for today and I wanted to know if this is an ok way to go. What do you guys think, @clebs @tobiscr ?

pkg/reconciler/instances/connectivityproxy/command.go
```diff
@@ -30,10 +36,42 @@ type CommandActions struct {
...
+					return "", errors.Errorf("Component does not have any release labels")
+				} else if unstruct.GetLabels()[ReleaseLabelKey] != appRelease {
+					return manifest, nil
+				}
+			}
+		}
+		return "", nil
+	})
+
+	return a.installOnCondition(context, chartProviderWithFilter)
```
clebs, line 68: You do not need `chartWithProviderFilter`, when calling `WithFilter` the receiver is a pointer, you are adding the filter directly to `context.ChartProvider`.
Therefore passing `context`  is enough.

The reason to return the provider in the "With" functions is the ability to do fluent programming (e.g. `x.WithA().WithB().WithC()`)

pkg/reconciler/chart/provider_test.go
```diff
@@ -53,6 +53,27 @@ func TestProvider(t *testing.T) {
...
 
+	t.Run("Render filtered manifest", func(t *testing.T) {
+		// kyma components
+		ws, err := wsFactory.Get(kymaVersion)
+		require.NoError(t, err)
+
+		provider, err := NewDefaultProvider(wsFactory, log)
+		require.NoError(t, err)
+
+		providerWithFilter := provider.WithFilter(func(manifest string) (string, error) { return "", nil })
```
clebs, line 64: You do not need `providerWithFilter`, when calling `WithFilter` the receiver is a pointer, you are adding the filter directly to `provider `.

The reason to return the provider in the "With" functions is the ability to do fluent programming (e.g. `x.WithA().WithB().WithC()`)

clebs (approved):

tobiscr (approved):

koala7659 (approved):
LGTM

PR #511, opened 2021-12-01
Enhance Unit Tests

jeremyharisch:
**Description**

This PR enhances multiple unit-test due to race conditions. Furthermore, it fixes a bug for the inMemory and persistent database implementation.

**Changes**

New Test-Cases:
- reconciliation repository:
    -  Create multiple instances of single reconciliations in parallel; Only one should be successful
    - Update single reconciliation in multiple parallel threads; Only one should be successful
    - Mark single reconciliation as finished in multiple parallel threads; Only one should be successful
- bookkeeper:
    - Mark two operations as orphan in multiple parallel threads; Each operation should only be updated once as an orphan
    - Finish two operations as orphan in multiple parallel threads; Each operation should only be updated once as an done
- Workerpool:
    - Multiple WorkerPools watching same reconciliation repository; Each operation should only be assigned to one worker / should only be invoked once
- Scheduler:
    - Multiple scheduler watching same inventory; For each cluster reconciliations should be only created once
 
Refactor:
- Introduction of `NewCluster ` function and `TestCluster` type, to reduce code duplication in unit tests; adapt testcase due to this
- Wrap errors in `factory.go`, `sqlite.go` for better error log output
- Introduction of `IsAlreadyInStateError`
- Refactor of bookkeeper: Introduction of `BookkeepingTask` to establish `Chain of Responsibility` pattern to have a better testable code (including written unit tests)

Fix:
- Increase test timeout to 20min, since Pipelines take longer than 15min, local execution takes around 3,5min
- Change host in unittest config to `127.0.0.1`, due to IPv6 issues
- In update call (persistent and inmemory): Check if new wanted state is same as current state; If yes update should fail
- `Invoke` of remoteInvoker called `i.updateOperationState(params, model.OperationStateClientError, errorReason)` twice; but it should only be called once
- Adapt Remote Invoker test due to DB  and remote invoker fix
- Fix for `runRemote` tests in inventory pkg
- Smaller fixes (Removed unneeded error wrapping)
- Switch DB isolation level to `repeadableRead`

**Related issue**
https://github.com/kyma-incubator/reconciler/issues/374


pkg/keb/cluster.go
```diff
@@ -0,0 +1,31 @@
+package keb
```
ruanxin, line 1: At now, we can put this file under some test folder like `keb/test` so it can be in `test` to indicate it's only for testing purposes. 

We should not mix code for test with production together. 

And based on our AD meeting, later we will have different test layout handle this kind of code only for testing, it can be put under dedicated test directory.

https://wiki.wdf.sap.corp/wiki/display/kyma/Project+layout+for+test+code

pkg/keb/clusterType.go
```diff
@@ -1,4 +1,9 @@
-{
+package keb
+
+type TestCluster []byte
+
+var (
+	Production TestCluster = []byte(`{
```
ruanxin, line 6: Same here, this one is also only for testing, and we can write a function to read the content directly from json file. but now leave it like this is fine, let's refactor later.

pkg/cluster/inventory_test.go
```diff
@@ -341,30 +335,6 @@ func newInventory(t *testing.T) Inventory {
 	return inventory
 }
 
-func newCluster(t *testing.T, runtimeID, clusterVersion int64, newConfigVersion bool) *keb.Cluster {
```
ruanxin, line 344: It's a good approach to extract this dummy data generator to a dedicated file👍

pkg/db/factory.go
```diff
@@ -32,7 +33,7 @@ func NewConnectionFactory(configFile string, migrate bool, debug bool) (Connecti
 	case "sqlite":
 		connFact, err := createSqliteConnectionFactory(encKey, debug, blockQueries, logQueries)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrap(err, "error creating sqliteConnectionFactory")
```
ruanxin, line 36: If this is the only place need to use `errors` package, it's not necessary to introduce it, and be careful when introducing external dependencies, the more introduced, the more compatibility issue has to concern later:)

If you check the disclaimer of this package, after Go2 is released, it's into maintenance mode

https://github.com/pkg/errors
With the upcoming Go2 error proposals this package is moving into maintenance mode. The roadmap for a 1.0 release is as follows:

    0.9. Remove pre Go 1.9 and Go 1.10 support, address outstanding pull requests (if possible)
    1.0. Final release.

ruanxin: use  `fmt/errors` should be enough in this case.

pkg/db/sqlite.go
```diff
@@ -117,18 +117,18 @@ func (scf *sqliteConnectionFactory) Init(_ bool) error {
 		//read DDL (test-table structure)
 		ddl, err := ioutil.ReadFile(scf.schemaFile)
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "error reading file DDL schema file '%s'", scf.schemaFile)
```
ruanxin, line 120: Please check my comments regarding this `errors` external package.

pkg/scheduler/reconciliation/inmemory.go
```diff
@@ -213,6 +213,10 @@ func (r *InMemoryReconciliationRepository) UpdateOperationState(schedulingID, co
 		return &repository.EntityNotFoundError{}
 	}
 
+	if err := operationAlreadyInState(op, state); err != nil && !allowInState {
```
ruanxin, line 216: I still trying to understand why we need this allowInState, but for code readability improvement. I assume the following way is a bit easy to understand, what do you think? If not allowed, then you don't have to check the op.state and state, right?
and also better use directly equal check in condition, use `!allowInState` I would suggest to rename it to `notAllowInState`
```
		if allowInState {
			err := operationAlreadyInState(op, state)
			if err != nil {
				return err
			}
		}
```

pkg/scheduler/service/bookkeepingtask.go
```diff
@@ -0,0 +1,64 @@
...
+				"(triggered by reconciliation with schedulingID '%s')",
+				recon.RuntimeID, newClusterStatus, recon.SchedulingID)
+			return nil
+		}
+		errMsg = fmt.Sprintf("finishOperation failed to update cluster '%s' to status '%s' "+
+			"(triggered by reconciliation with schedulingID '%s'): %s",
+			recon.RuntimeID, newClusterStatus, recon.SchedulingID, err)
+		fo.logger.Errorf(errMsg)
+	}
+	return []error{errors.New(errMsg)}
```
ruanxin, line 63: Let's move the `errMsg` declaration directly above this line, declaration direct before use is a fundamental rule, but can avoid many surprises.

pkg/db/sqlite.go
```diff
@@ -2,11 +2,11 @@ package db
 
 import (
 	"database/sql"
+	log "github.com/kyma-incubator/reconciler/pkg/logger"
+	"github.com/pkg/errors"
```
ruanxin, line 6: another `pkg/errors` introduce

pkg/db/factory.go
jeremyharisch: I used it, since we use it everywhere in Kyma/reconciler project. Thus, i am not introducing a new dependency to our project here.
I would for now stick with the our default way of doing, and then change it when we change it everywhere

pkg/scheduler/reconciliation/inmemory.go
jeremyharisch: Already addressed in personal meeting

pkg/scheduler/service/bookkeepingtask.go
jeremyharisch: If we would more it, then we would need to adapt my logic in this function. For now, I wasnt able to come up with a better solution, do you have a proposal?

jeremyharisch: done

pkg/scheduler/service/bookkeepingtask.go
```diff
@@ -0,0 +1,78 @@
...
+				recon.RuntimeID, newClusterStatus, recon.SchedulingID)
+			return nil
+		}
+		errMsg = fmt.Sprintf("finishOperation failed to update cluster '%s' to status '%s' "+
+			"(triggered by reconciliation with schedulingID '%s'): %s",
+			recon.RuntimeID, newClusterStatus, recon.SchedulingID, err)
+	} else {
+		errMsg = fmt.Sprintf("finishOperation failed to update cluster '%s' to status '%s' "+
+			"(triggered by reconciliation with schedulingID '%s'): CLuster is already in final state",
+			recon.RuntimeID, newClusterStatus, recon.SchedulingID)
```
ruanxin, line 75: How can we make this assumption the existing cluster is in the final state based on your `newClusterStatus` is not final?

pkg/scheduler/service/bookkeepingtask.go
```diff
@@ -0,0 +1,78 @@
...
+type BookkeepingTask interface {
+	Apply(reconResult *ReconciliationResult, maxRetries int) []error
+}
+
+type markOrphanOperation struct {
+	transition *ClusterStatusTransition
+	logger     *zap.SugaredLogger
+}
+
+func (oo markOrphanOperation) Apply(reconResult *ReconciliationResult, maxRetries int) []error {
```
ruanxin, line 20: If you don't need maxRetries here, then we should not design this `Apply` as a shared interface method.

pkg/scheduler/service/bookkeeper_test.go
```diff
@@ -67,8 +52,11 @@ func TestBookkeeper(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second) //stop bookkeeper after 5 sec
 	defer cancel()
 
+	transition := newClusterStatusTransition(dbConn, inventory, reconRepo, logger.NewLogger(true))
 	start := time.Now()
-	require.NoError(t, bk.Run(ctx))
+	require.NoError(t, bk.Run(ctx,
+		markOrphanOperation{transition: transition, logger: transition.logger},
```
ruanxin, line 58: I still not know the purpose, but why currently need to do two operation tasks in one bookkeeper run? Let's discuss in in our meeting.

pkg/scheduler/reconciliation/reconciliation.go
```diff
@@ -146,3 +146,29 @@ func concatStateReasons(state model.OperationState, reasons []string) (string, e
...
+	}
+	return nil
+}
+
+type alreadyInStateError struct {
+	op *model.OperationEntity
+}
+
+func (err *alreadyInStateError) Error() string {
+	return fmt.Sprintf("cannot update state of operation '%s' because it is already in state %s", err.op.Component, err.op.State)
```
ruanxin, line 162: As we addressed this during call, we are not clear why it's an error, I leave this comment here as a bookmark, to address in our meeting.

pkg/db/sqlite.go
ruanxin: I reopened this comment as bookmark, to address some potential tech dept, and how we deal with it. Let's discuss it also in our meeting.

pkg/scheduler/reconciliation/reconciliation.go
jeremyharisch: As already explained, typical Go-Style programming.

Here is a reference for you: https://gabrieltanner.org/blog/golang-error-handling-definitive-guide

pkg/scheduler/service/bookkeepingtask.go
jeremyharisch: Will rewrite log message

pkg/scheduler/service/bookkeepingtask.go
jeremyharisch: Decided to pass bk config to bk tasks

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -2,6 +2,10 @@ package worker
 
 import (
 	"context"
+	test2 "github.com/kyma-incubator/reconciler/pkg/keb/test"
```
clebs, line 5: I find this name pretty confusing, as if there was a `test` package also imported. I would either call it something descriptive like `kebTest`.

tobiscr (changes requested):

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -383,9 +387,9 @@ func (r *PersistentReconciliationRepository) UpdateOperationState(schedulingID,
...
 		}
-		cnt, err := q.Update().
-			Where(whereCond).
-			ExecCount()
+
+		update := q.Update().Where(whereCond)
+		cnt, err := update.ExecCount()
 
 		if cnt == 0 {
 			return fmt.Errorf("update of operation '%s' to state '%s' failed: no row was updated "+
```
tobiscr, line 395: oh, we should probably check `err != nil` before to avoid that the 0-count isn't caused by a different issue

pkg/scheduler/worker/workerpool_test.go
jeremyharisch: That is true! Sorry, I wasnt looking at the auto-generated code in Goland. I will take more care about such stuff in future!

pkg/db/query.go
```diff
@@ -103,6 +102,74 @@ func (q *Query) addWhereCondition(whereCond map[string]interface{}, plcHdrOffset
 	return args, nil
 }
 
+func (u *Update) addWhereCondition(whereCond map[string]interface{}) ([]interface{}, error) {
```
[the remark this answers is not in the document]
jeremyharisch: Done

tobiscr (approved):

Issue #516, opened 2021-12-02
Unit test TestCountRetries/"Calculate retry count for a  retryable cluster" will always fail when expectedErrRetryable > 25


vvxxvvxx:
**Description**

The unit test `TestCountRetries/"Calculate retry count for a  retryable cluster"` will always fail if `expectedErrRetryable` is set to more than 25 and the actual count is always 25.

when:
```
	t.Run("Calculate retry count for a  retryable cluster", func(t *testing.T) {
		expectedErrRetryable := 26
```
the test result:
```
...
{"level":"DEBUG","time":"2021-12-02T14:14:51.941+0800","message":"Sqlite3 QueryRow(): INSERT INTO inventory_cluster_config_statuses (runtime_id, cluster_version, config_version, status, deleted) VALUES ($1, $2, $3, $4, $5) RETURNING id, runtime_id, cluster_version, config_version, status, deleted, created | [runtime1 1 1 reconciling false]"}
{"level":"DEBUG","time":"2021-12-02T14:14:51.974+0800","message":"Sqlite3 Query(): SELECT id, runtime_id, cluster_version, config_version, status, deleted, created FROM inventory_cluster_config_statuses WHERE config_version=$1 AND runtime_id=$2 ORDER BY id desc LIMIT 50 | [1 runtime1]"}
    inventory_test.go:319:
                Error Trace:    inventory_test.go:319
                Error:          Not equal:
                                expected: 26
                                actual  : 25
                Test:           TestCountRetries/Calculate_retry_count_for_a__retryable_cluster
--- FAIL: TestCountRetries (1.76s)
    --- FAIL: TestCountRetries/Calculate_retry_count_for_a__retryable_cluster (1.51s)
FAIL
exit status 1
FAIL    github.com/kyma-incubator/reconciler/pkg/cluster        1.973s
```

when:
```
	t.Run("Calculate retry count for a  retryable cluster", func(t *testing.T) {
		expectedErrRetryable := 151
```

the test result:
```
{"level":"DEBUG","time":"2021-12-02T14:19:48.807+0800","message":"Sqlite3 Query(): SELECT id, runtime_id, cluster_version, config_version, status, deleted, created FROM inventory_cluster_config_statuses WHERE config_version=$1 AND runtime_id=$2 ORDER BY id desc LIMIT 50 | [1 runtime1]"}
    inventory_test.go:319:
                Error Trace:    inventory_test.go:319
                Error:          Not equal:
                                expected: 151
[...]

PR #517, opened 2021-12-02
Increase Limit for CountRetries

jeremyharisch:
Issue: #516 

Fix:
Increase limit for CountRetires SQL query to a multiple of `maxRetries`

### DIFF
--- a/pkg/cluster/inventory.go
+++ b/pkg/cluster/inventory.go
@@ -23,7 +23,7 @@ type Inventory interface {
 	StatusChanges(runtimeID string, offset time.Duration) ([]*StatusChange, error)
 	ClustersToReconcile(reconcileInterval time.Duration) ([]*State, error)
 	ClustersNotReady() ([]*State, error)
-	CountRetries(runtimeID string, configVersion int64) (int, error)
+	CountRetries(runtimeID string, configVersion int64, maxRetries int) (int, error)
 }
 
 type DefaultInventory struct {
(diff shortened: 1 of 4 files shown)


pkg/cluster/inventory_test.go
```diff
@@ -314,7 +317,7 @@ func TestCountRetries(t *testing.T) {
 			require.NoError(t, err)
 		}
 		//count how often retry happened
-		cnt, err := inventory.CountRetries(clusterState.Configuration.RuntimeID, clusterState.Configuration.Version)
+		cnt, err := inventory.CountRetries(clusterState.Configuration.RuntimeID, clusterState.Configuration.Version, 250)
```
ruanxin, line 320: If you put `250`, the limit number is `250 * 5`, is it the actual expectation?

pkg/cluster/inventory.go
```diff
@@ -48,12 +48,12 @@ func NewInventory(conn db.Connection, debug bool, collector metricsCollector) (I
...
 }
 
-func (i *DefaultInventory) CountRetries(runtimeID string, configVersion int64) (int, error) {
+func (i *DefaultInventory) CountRetries(runtimeID string, configVersion int64, maxRetries int) (int, error) {
 	q, err := db.NewQuery(i.Conn, &model.ClusterStatusEntity{}, i.Logger)
 	if err != nil {
 		return 0, err
 	}
-	clusterStatuses, err := q.Select().Where(map[string]interface{}{"RuntimeID": runtimeID, "ConfigVersion": configVersion}).OrderBy(map[string]string{"ID": "desc"}).Limit(50).GetMany()
+	clusterStatuses, err := q.Select().Where(map[string]interface{}{"RuntimeID": runtimeID, "ConfigVersion": configVersion}).OrderBy(map[string]string{"ID": "desc"}).Limit(maxRetries * 5).GetMany()
```
ruanxin, line 56: To be honest, I still not understand why `*5`, is it just to increase the safeness to query enough records? 
And I have some opinion to set the Limit, I think the purpose here is performance concern, but there is not so many records in our case, and use `limit` is not a magic way, it needs to combine with proper index and ordering, otherwise the performance will even be decreased, like mentioned here:
https://stackoverflow.com/questions/6037843/extremely-slow-postgresql-query-with-order-and-limit-clauses

Regarding performance improvement, a general approach is to prevent doing overengineering, we anyway can resolve it when it actually hit performance alert to us.

I suggest deleting this `Limit` clause.

pkg/cluster/inventory_test.go
jeremyharisch: Yes, since we want maximum of 250 retires in this test case ( we can also lower it, doesnt matter tbh), then we need something around 5 times the rows of the db table

jeremyharisch: But indeed I will put it to a lower number, makes more sense

pkg/cluster/inventory.go
clebs: This is a very good point! I also think that just arbitrarily multiplying by a magic number is no solution. Removing the limit isn't a solution either, we need to build a system that is scalable and we could potentially be managing 1000s of clusters when this goes live.

Instead we should see when we really need a limit, to which extent and why. My proposal would be to add some intelligence to the query builder to automatically set this depending on the query (has where? has order? is selecting by PK? Joins?)

tobiscr (dismissed):

tobiscr (approved):

PR #519, opened 2021-12-02
Return reconciliation runs from precise timebox

pPrecel:
Allow user to ask for reconciliations after/before given timestamp or/and extract only a few last results (tile filtered by creation time)

https://github.com/kyma-project/control-plane/issues/1082


cmd/mothership/mothership/start/http.go
```diff
@@ -353,6 +357,48 @@ func getReconciliations(o *Options, w http.ResponseWriter, r *http.Request) {
 		})
 	}
 
+	//filter by time
+	if after := r.URL.Query().Get(paramAfter); after != "" {
```
tobiscr, line 361: Would not do it in code, let the DB do that for you by setting the needed WHERE condition. You can set

cmd/mothership/mothership/start/helpers.go
```diff
@@ -18,3 +21,36 @@ func components(cfg model.ClusterConfigurationEntity) []keb.Component {
 	}
 	return components
 }
+
+func filterReconciliationsAfter(time time.Time, reconciliations []keb.Reconciliation) []keb.Reconciliation {
```
tobiscr, line 25: Would move this code into a filter (see https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/reconciliation/filter.go)

The `reconciliation.Repositroy` should filter that either by defining a SQL query using a WHERE condition (in `Filter.FilterByQuery` or by checking the creation-time in `Filter.FilterByInstance` << this can use the code you have implemented in your helper.go)

pkg/scheduler/reconciliation/filter.go
```diff
@@ -8,12 +8,106 @@ import (
...
+	var whereRaw string
+	for i := range ws.Statuses {
+		if i > 0 {
+			whereRaw = fmt.Sprintf("%s%s", whereRaw, " OR ")
+		}
+
+		whereRaw = fmt.Sprintf("%sstatus='%s'", whereRaw, ws.Statuses[i])
+	}
+
+	q.WhereRaw((whereRaw))
```
tobiscr, line 71: two brackets

pkg/scheduler/reconciliation/filter.go
```diff
@@ -8,12 +8,106 @@ import (
...
+		return nil
+	}
+
+	var whereRaw string
+	for i := range ws.Statuses {
+		if i > 0 {
+			whereRaw = fmt.Sprintf("%s%s", whereRaw, " OR ")
+		}
+
+		whereRaw = fmt.Sprintf("%sstatus='%s'", whereRaw, ws.Statuses[i])
```
tobiscr, line 68: Please use db.ColumnHandler for retrieving column-names, otherwise we wont notice DB-schema discrepancies properly:

		statusColHandler, err := db.NewColumnHandler(&model.ReconciliationEntity{}, q.Conn, q.Logger)
		if err != nil {
			return err
		}
		statusColName, err := statusColHandler.ColumnName("Status")

JFYI: the `conn` field in the Query object can be made public, currently it is private.

pkg/scheduler/reconciliation/filter.go
```diff
@@ -8,12 +8,106 @@ import (
...
+	}
+	return nil
+}
+
+type WithCreationDateAfter struct {
+	Time time.Time
+}
+
+func (wd *WithCreationDateAfter) FilterByQuery(q *db.Select) error {
+	q.WhereRaw(fmt.Sprintf("created>'%s'", wd.Time.Format("2006-01-02 15:04:05.000")))
```
tobiscr, line 94: (as above) Please use db.ColumnHandler to retrieve column-names. The column handler ensures, that the column-name is always in-sync with the DB schema

pkg/scheduler/reconciliation/filter.go
```diff
@@ -8,12 +8,106 @@ import (
...
+	return nil
+}
+
 type WithCreationDateBefore struct {
 	Time time.Time
 }
 
 func (wd *WithCreationDateBefore) FilterByQuery(q *db.Select) error {
-	q.WhereRaw("created<$1", wd.Time.Format("2006-01-02 15:04:05.000"))
+	q.WhereRaw(fmt.Sprintf("created<'%s'", wd.Time.Format("2006-01-02 15:04:05.000")))
```
tobiscr, line 110: use db.ColumnHandler please to get the `status` column-name

pkg/scheduler/reconciliation/filter.go
```diff
@@ -8,12 +8,106 @@ import (
 	"github.com/kyma-incubator/reconciler/pkg/model"
 )
 
+type FilterMixer struct {
```
tobiscr, line 11: Would be good to have a unit-test for it.

You can get a generated DB string from the query object (`q.String()`)

tobiscr (changes requested):

pkg/scheduler/reconciliation/filter.go
pPrecel: Thanks for this hint!

pkg/scheduler/reconciliation/filter.go
pPrecel: Good idea

tobiscr (approved):

Issue #333, opened 2021-11-02
Deletion reconciler: enable clean deletion of a Kyma installation


clebs:
**Description**

The deletion of Kyma has to happen in multiple steps of following sequence:

1. [delete-cr] Call deletion reconciler: remove custom resources
2. Call all component reconciler and 
     1. let them delete their manifests/execute custom deletion logic
     2. delete CRDs (done by base-reconciler)
3. [delete-ns] Call again deletion reconciler again and remove namespaces: if they don't get deleted after a timeout, remove finalisers and enforce deletion of namespace

AC:
* Deletion of Kyma removes all Kyma managed resources (incl. CRDs, CRs and any deployed resource)
* After deleting all Kyma resources, a new installation can Kyma is possible without any additonal steps

**Reasons**

Enable proper uninstallation of Kyma on a cluster.

tobiscr:
Hint from PB:

* `undeploy --force --timeout 0`
   - delete all resources from all apigroups with kyma (plus other well known). Actively remove finalizers

tobiscr:
Caused by current time pressure, it would be good to have following functionality in the first PR:

1. Deletion reconciler is called at the end of the regular deletion
2. Deletion reconciler is executing the existing deletion logic which the Kyma CLI is using

In next iteration, it would be great to get the flow as described in the ticket i place.

tobiscr:
Current status:
  - @Tomasz-Smelcerz-SAP  : Move CLI delete leftovers logic to reconciler:
        - no longer need to delete the `istio-system` namespace
        - No longer need to delete CRDs manually in custom logic
        - Add `kyma-system` hack
      - @clebs : Open CLI PR (needs reconciler bump only) and should work once the reconciler is ready
      - @tobiscr :
          - Merge Tomasz’s PR
          - Bump reconciler in CLI PR
          - Merge CLI PR

PR #521, opened 2021-12-02
Add "cleanup" reconciler

Tomasz-Smelcerz-SAP:
**Description**
Add a reconciler for a fake "cleanup" component. This should be invoked as the last operation during the "delete" operation.

**Changes proposed in this pull request:**
- Add new reconciler that only handles delete operations
- Copy existing CLI logic from kyma "undeploy" command
- Adjust with changes from https://github.com/kyma-project/cli/pull/1109

**See also**: https://github.com/kyma-incubator/reconciler/issues/333


pkg/reconciler/service/install.go
```diff
@@ -38,9 +38,6 @@ func (r *Install) Invoke(ctx context.Context, chartProvider chart.Provider, task
 	}
 
 	if task.Type == model.OperationTypeDelete {
-		if task.Component == model.CRDComponent {
```
tobiscr, line 41: To save unnecessary call to the deletion-reconciler when we are reconciling, it would be good to return when the Operation-Type is "Reconile", E.g.

     if task.Type == model.OperationTypeReconcile && task.Component == model. CleanupComponent {
       return nil
    }

tobiscr (changes requested):

pkg/model/clusterconfigentity.go
```diff
@@ -114,8 +121,10 @@ func newReconciliationSequence(preComponents [][]string) *ReconciliationSequence
 		preComponents: preComponents,
 	}
 	reconSeq.Queue = append(reconSeq.Queue, []*keb.Component{ //CRDs are always processed at the very beginning
-		crdComponent,
+		//TODO: These are executed in parallel, take care of the proper order in the final implementation.
```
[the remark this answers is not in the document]
clebs: This is also a good option, the cleaner could be the highest prio pre component, making it deterministic to know when it runs.

pkg/reconciler/service/install.go
clebs: In that case we need to turn around the whole if-else block, but definitely good to return early.

pkg/model/clusterconfigentity.go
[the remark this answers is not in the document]
clebs: As discussed, we do not need to add the `cleaner` component to the prerequisites, just in a separate priority, making it run sequentially with the CRD component.

pkg/reconciler/service/install.go
```diff
@@ -30,17 +30,14 @@ func (r *Install) Invoke(ctx context.Context, chartProvider chart.Provider, task
 	var manifest string
 	if task.Component == model.CRDComponent {
 		manifest, err = r.renderCRDs(chartProvider, task)
-	} else {
+	} else if task.Component != model.CleanupComponent { // TODO add better support for components that do not have manifests
```
tobiscr, line 33: @clebs : please directly open an issue for it so that we won't forget it

tobiscr (approved):

clebs (approved):

PR #542, opened 2021-12-08
Disabled clusters can't change their status

### LINKED ISSUE #491: Clusters which are disabled should not change status to `pending_reconciliation`
**Description**

When a cluster is disabled, an update from KEB should not enable the cluster again. The new created cluster should also be disabled until it gets enabled again by SRE.

**Expected result**

When a cluster is disabled but a new cluster version gets created caused by a cluster update from KEB, the new cluter version should also be in state `disabled`.

**Actual result**

KEB updates are creating new cluster versions which are initailly in state `pending`.

**Steps to reproduce**

**Troubleshooting**

clebs:
Fixes #491

### DIFF
--- a/pkg/cluster/inventory.go
+++ b/pkg/cluster/inventory.go
@@ -209,6 +209,9 @@ func (i *DefaultInventory) createStatus(configEntity *model.ClusterConfiguration
 			i.Logger.Debugf("No differences found for status of cluster '%s': not creating new database entity", configEntity.RuntimeID)
 			return oldStatusEntity, nil
 		}
+		if oldStatusEntity.Status.IsDisabled() {
+			newStatusEntity.Status = model.ClusterStatusReconcileDisabled
+		}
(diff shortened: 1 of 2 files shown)


pkg/cluster/inventory.go
```diff
@@ -209,6 +209,9 @@ func (i *DefaultInventory) createStatus(configEntity *model.ClusterConfiguration
 			i.Logger.Debugf("No differences found for status of cluster '%s': not creating new database entity", configEntity.RuntimeID)
 			return oldStatusEntity, nil
 		}
+		if oldStatusEntity.Status.IsDisabled() {
+			newStatusEntity.Status = model.ClusterStatusReconcileDisabled
+		}
 	} else if !repository.IsNotFoundError(err) {
```
tobiscr, line 215: awesome. that it already.. much shorter as initially expected :D

tobiscr (approved):

Issue #524, opened 2021-12-03
Reconciler terminates scaled pods


veichtj:
**Description**

Kyma 2.0 is installed successfully via reconciler.
ORY oathkeeper pods are deployed within production profile with a min replica of 3, also the hpa sepc says min 3, max 10.
When k8s hpa scales up the oathkeeper pods to more than 3, because of more load on the cluster, during a reconciler run all pods that were scaled up are going into termination. After reconciler run is finished, the hpa again scales up. This potentially leads to interruption and also additional load with pod starting and terminating.

**Expected result**

Reconciler does not terminate additional scaled up pods

**Actual result**
Reconciler applies the default set for scaling.

**Steps to reproduce**

- Kyma 2.0 installed with production profile
- check ory-oathkeeper pods (should be 3 per default)
- create load on ory-oathkeeper for e.g. calling vs
- check hpa should scale up additional pods for ory-oathkeeper
- run reconciler for ory component
- check ory-oathkeeper pod status, you see Terminating for additionally scaled ones
- after reconciler is finished, again additional Pods are initialized

**Troubleshooting**

cnvergence:
https://github.com/kubernetes/website/pull/29739

cnvergence:
Helm seems to have templated this setting: https://github.com/helm/helm/pull/7381
Will prepare PR on our side @tobiscr

cnvergence:
When investigating this issue with the adjusted Helm chart I have still encountered this problem.
This is happening to any HPA configuration with minimum Replicas setting larger than one.
The same issue apply also to application-connector - central-application-gateway deployment.

tobiscr:
We agreed with @cnvergence  to come up with a kind of interceptor which is able to iterate through the manifest of a component with following functionality:

1. Search for HPA
     1. YES: continue at step 2
     2. NO: done
2. Check if referenced resource in HPA is already deployed in cluster
     1. NO: check if referenced resource (deployment etc.) includes already a `replica` field
          1. YES: done
          2. NO: set field `replica` using the `minValue` from HPA resource
     2.  YES: retrieve value of `replica` field from the deployed resource and set this value in the manifest of the referenced resource

tobiscr:
Implementation hint:

To avoid different types of interceptor interfaces, we could enrich the existing interceptor interface like

     type ResourceInterceptor interface {
          Intercept(resource []*unstructured.Unstructured, namespace string) (InterceptionResult, error)
          SupportedKinds() []string
     }

[...]

PR #546, opened 2021-12-08
Modify ResourceInterceptor interface

suleymanakbas91:
See https://github.com/kyma-incubator/reconciler/issues/524 (more specifically https://github.com/kyma-incubator/reconciler/issues/524#issuecomment-988083858)

### DIFF
--- a/pkg/reconciler/kubernetes/adapter_test.go
+++ b/pkg/reconciler/kubernetes/adapter_test.go
@@ -62,7 +62,7 @@ type testInterceptor struct {
 func (i *testInterceptor) Intercept(resources Resources, _ string) error {
 	interceptorFunc := func(kind string, u *unstructured.Unstructured) error {
 		u.SetLabels(expectedLabels)
-		return nil
+		return i.err
 	}
 
 	return resources.Visit(interceptorFunc)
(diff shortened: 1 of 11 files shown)


pkg/reconciler/kubernetes/adapter_test.go
```diff
@@ -55,13 +56,16 @@ var expectedResourcesWithNs = []*Resource{
...
 type testInterceptor struct {
-	result InterceptionResult
-	err    error
+	err error
 }
 
-func (i *testInterceptor) Intercept(resource *unstructured.Unstructured, _ string) (InterceptionResult, error) {
-	resource.SetLabels(expectedLabels)
-	return i.result, i.err
+func (i *testInterceptor) Intercept(resources map[string][]*unstructured.Unstructured, _ string) error {
```
tobiscr, line 62: Maybe for the later PR: could we make the map to an object which offers some convenient looping and getter function.

E.g.

     type Resources map[string][]*unstructured.Unstructured

     func (r resources) Visit(callback func(kind string, u *unstructured.Unstructured) error) error {
     	for kind := range r {
     		for _, resource := range r[kind] {
     			if err := callback(kind, resource); err != nil {
     				return err
     			}
     	     	}
     	}
     	return nil
     }

    func (r resources) Get(kind string) []*unstructured.Unstructured {
      	return r[kind]
     }

tobiscr (dismissed):

tobiscr (approved):

pkg/reconciler/kubernetes/adapter_test.go
suleymanakbas91: Done

tobiscr (approved):

Issue #527, opened 2021-12-03
Updating cluster config uses wrong component version


wozniakjan:
**Description**

When updating cluster configuration through reconciler API the reconciler reports error.

```
{"level":"DEBUG","time":"2021-12-03T09:46:02.128Z","message":"Bookkeeper evaluated reconciliation (schedulingID:fb7c515e-c0dd-4808-b94a-757eed445615--3a4cb5bf-0003-493e-99d4-d8565b4dc1fd) for cluster 'fb7c515e-c0dd-4808-b94a-757eed445615' to cluster status 'reconciling': Done=CRDs,cluster-essentials,istio-configuration,certificates,cluster-users,rma,busola-migrator,service-catalog-addons,serverless,eventing,api-gateway,helm-broker,monitoring,ory,rafter,connectivity-proxy,application-connector,service-manager-proxy,compass-runtime-agent,service-catalog,logging / Error= / Other=btp-operator[error: Failed to get manifest for component 'btp-operator' in Kyma version '2.0.0-rc7' using repository 'https://github.com/kyma-incubator/sap-btp-service-operator/releases/download/v0.1.18-custom/sap-btp-operator-0.1.18.tar.gz' : stat /tmp/reconciler/2.0.0-rc7-btp-operator/btp-operator/Chart.yaml: no such file or directory],sc-migration[error: Failed to get manifest for component 'sc-migration' in Kyma version '2.0.0-rc7' using repository 'https://github.com/kyma-incubator/sc-removal/releases/download/0.5.0/sap-btp-operator-migration-v0.5.0.tgz' : stat /tmp/reconciler/2.0.0-rc7-sc-migration/sc-migration/Chart.yaml: no such file or directory]"}
```
[reconciler_payload.json](https://github.com/kyma-incubator/reconciler/files/7650339/reconciler_payload.json.txt)

**Expected result**

Reconciler to install BTP Operator

**Actual result**

Reconciler failed to install BTP Operator

**Steps to reproduce**

https://github.com/kyma-project/control-plane/blob/2b282dc12b723ca238ec75c308a563caecd98123/components/kyma-environment-broker/internal/process/update/apply_reconciler_configuration.go#L41-L53

**Troubleshooting**

wozniakjan:
fwiw, same bug with rc8
```
[...]

Issue #463, opened 2021-11-24
Workspace factory: download of external components has to consider URL in workspace folder name


tobiscr:
**Description**

When downloading an external component (form a cloud storage etc, not a GIT hosted component), the generated workspace-dir does not distinguish between different URLs used for the same component in the workspace-directory. It's using the version as a separate, but the version is no longer mandatory and can be undefined.

Implies: after downloading the component which does not set a version, the component wonÄt be downloaded again even when the URL was changed.

To solve this issue, the URL should be considered in the workspace directory name (e.g. as hash etc.)

**Expected result**

* Download of an external (non GIT hosted) component should not use the version in the workspace folder name (version is not mandatory), instead it should use the URL (e.g. as a hash).

**Actual result**

* Version is used in the workspace folder name

**Steps to reproduce**

See workspace factory unit test.

**Troubleshooting**

tobiscr:
Maybe worth to also address #450 as part of this change as it's quite close related to each other.

tobiscr:
requires #462

Issue #450, opened 2021-11-22
Workspace factory: deletion of external component not working


tobiscr:
**Description**

The deletion of an external component is not removing the sources on disk (currently is is using the `Delete` command for Kyma which isn't working for external components as the name of workspace-folder is differently created for a Kyma workspace).

A dedicated delete function for external components is required which is removing the folder-name used for external components.

**Expected result**

Deleting an external component has to remove the workspace-folder used for the external component.

**Actual result**

Removing an external component tries to removes the workspace-folder which would be used for Kyma download.

**Steps to reproduce**

**Troubleshooting**

tobiscr:
requires #462 + #463

ralikio:
Related to #469

tobiscr:
@ralikio: how is this issue related to the #469- this bug was created because we could not delete workspaces in our unit-tests when caling the `Delete` function of the workspace factory. This `Delete`-fct us currently not used somewhere else afaik.

ralikio:
@tobiscr Issue removed.

PR #580, opened 2021-12-15
Change workspace structure for downloaded components (#463)

m00g3n:
- #463
- #450

See also: 
- https://github.com/kyma-incubator/reconciler/issues/527


pkg/reconciler/chart/component.go
```diff
@@ -15,6 +15,10 @@ type Component struct {
 	configuration map[string]interface{}
 }
 
+func (c *Component) isRepository() bool {
```
tobiscr, line 18: Naming can be a bit confusing.. `component.isRepository()`  - maybe we could point out whether its an external component coming form a GIT repo, like `isExternalGitComponent` or something.

tobiscr (commented):
Looks in general very good! Just one short comment

tobiscr (changes requested):
Maybe we can use a bit more expressive name for `isRepository`.

suleymanakbas91 (approved):

PR #585, opened 2021-12-15
Strip namespace from templates of selected non namespaced kinds

kwiatekus:
Remove `namespace` field  from templates for some selected cluster-wide resources where we realise namespace caused problems in rendering of kyma resources at the reconciler


m00g3n (approved):
LGTM

strekm (approved):
/meow

franpog859 (approved):

Issue #245, opened 2021-10-09
Protect endless operation rescheduling: Mothership reconciler has to track operation retries and pass the max-retries to component reconcilers


tobiscr:
**Description**

To avoid endless reconciled operations (which can happen in case of a technical bug, failing error-reports by component reconcilers etc.), the mothership reconciler has count the retries for an operation.

The mothership reconciler will also tell the component reconcilers how often they should maximal retry an operation.

Proposal for the technical implementation:

1. Retries should be configured in the mothership reconciler and no longer on the component reconciler side.

   * A new configuration option has to be added to the mothership reconciler

        * define a new field in the [worker pool config struct](https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/worker/config.go) (e.g. `maxOperationRetries`)
        * add a [default value](https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/worker/config.go#L9-L13) for it (recommended value is `5`)
        * if no value was configured, set the default for the field in the [`validate` function](https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/worker/config.go#L24) 

   * remove the `maxRetries` configuration option [from the component-reconciler](https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/service/reconciler.go#L41)
   * rename the [`WithRetry`](https://github.com/kyma-incubator/reconciler/blob/main/pkg/reconciler/service/reconciler.go#L181) to `WithRetryDelay` and offer only the option to configure the retry delay

2. On mothership reconciler side is we have per operation to record the number of retries and the latest retryID.  The field `retryID` tracks the `retryID` which the heartbeat-mechanism of the component reconciler is continuously sending (why this is done is explained in point 6).

    * Add the fields `retries` (int64) and `retryID` (string) to the [`model.OperationEntity`](https://github.com/kyma-incubator/reconciler/blob/main/pkg/model/operationentity.go#L13)
    * Provide DB migration scripts to `config/db/postgres` directory  for adding new columns to the DB schema (see [`00004_add_cluster_status_delete_column.*.sql`](https://github.com/kyma-incubator/reconciler/blob/main/configs/db/postgres/000004_add_cluster_status_delete_column.up.sql) as example for DB migration scripts).

3. The worker pool of the mothership reconciler has to consider only operations which haven't yet reached the `maxOperationRetries` limit

[...]

PR #586, opened 2021-12-15
Protect endless reconciliation of operations -  245

khlifi411:
#245


configs/db/postgres/000009_update_operation_retries.up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE scheduler_operations
```
tobiscr, line 1: you can actually do it in step 000008, can't you?

pkg/scheduler/reconciliation/inmemory.go
```diff
@@ -240,3 +242,31 @@ func (r *InMemoryReconciliationRepository) UpdateOperationState(schedulingID, co
...
+	if !ok {
+		return &repository.EntityNotFoundError{}
+	}
+	op, ok := r.operations[schedulingID][correlationID]
+	if !ok {
+		return &repository.EntityNotFoundError{}
+	}
+
+	// copy the operation to avoid having data races while writing
+	opCopy := *op
```
tobiscr, line 260: Not needed.. there cannot be data-races as we use a mutex above

configs/db/postgres/000009_update_operation_retries.up.sql
khlifi411: Yup. I will do the ddl changes in one step.

pkg/scheduler/reconciliation/inmemory.go
khlifi411: I copied the code from the `UpdateOperationState` method implementation, which is also using a lock. should we skip the copy step in the `UpdateOperationState` method too?

tobiscr: ok, that would be good. Please drop this part form the other fct as well. I don't think that a copy will make anything better. A copy won't avoid race-conditions when two routines are changing the same operations... finally, the last one will overwrite with his copy the change of the other one

tobiscr (dismissed):

configs/db/postgres/000008_add_operation_retry_id_and_retries.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE scheduler_operations DROP COLUMN "retry_id" VARCHAR(255);
+ALTER TABLE scheduler_operations DROP COLUMN "retries" int;
```
ruanxin, line 2:
```suggestion
ALTER TABLE scheduler_operations DROP COLUMN "retries" int;

```

configs/db/postgres/000008_add_operation_retry_id_and_retries.up.sql
```diff
@@ -0,0 +1,24 @@
...
+    ALTER COLUMN "retry_id" SET NOT NULL;
+
+ALTER TABLE scheduler_operations
+    ADD COLUMN "retries" int;
+
+UPDATE scheduler_operations
+SET retries = 0;
+
+ALTER TABLE scheduler_operations
+    ALTER COLUMN "retries" SET NOT NULL;
```
ruanxin, line 24:
```suggestion
    ALTER COLUMN "retries" SET NOT NULL;
    
```

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -87,6 +88,63 @@ func TestWorkerPool(t *testing.T) {
...
 
+func TestWorkerPoolMaxOpRetriesReached(t *testing.T) {
+	test.IntegrationTest(t) //required because a valid Kubeconfig is required to create test cluster entry
+
+	//create cluster inventory
+	inventory, err := cluster.NewInventory(db.NewTestConnection(t), true, &cluster.MetricsCollectorMock{})
+	require.NoError(t, err)
+
+	//add cluster to inventory
+	clusterState, err := inventory.CreateOrUpdate(1, kebTest.NewCluster(t, "1", 1, false, kebTest.OneComponentDummy))
```
ruanxin, line 99: let's choose a new `runtimeID`, not use `"1"` which is used in another test in case some data is not cleaned up.

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -87,6 +88,63 @@ func TestWorkerPool(t *testing.T) {
...
+	}()
+
+	//create test invoker to be able to verify invoker calls
+	testInvoker := &testInvoker{}
+
+	//create reconciliation for cluster
+	testInvoker.reconRepo = reconciliation.NewInMemoryReconciliationRepository()
+	_, err = testInvoker.reconRepo.CreateReconciliation(clusterState, nil)
+	require.NoError(t, err)
+	opsProcessable, err := testInvoker.reconRepo.GetProcessableOperations(0)
```
ruanxin, line 114: Can we test with `defaultMaxParallelOperations`? which is 25, not 0

pkg/scheduler/worker/workerpool_test.go
khlifi411: That's a good idea! It's done!

pkg/scheduler/worker/workerpool_test.go
khlifi411: Yup! Have updated the code 👍

ruanxin (approved):

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -87,6 +88,63 @@ func TestWorkerPool(t *testing.T) {
...
+	require.Len(t, opsProcessable, 1)
+	require.NoError(t, err)
+
+	//create worker pool config
+	testConfig := &Config{MaxOperationRetries: 1}
+
+	//set operation Retries as to be equal or greater than MaxOpRetries
+	op := opsProcessable[0]
+	op.Retries = 1
```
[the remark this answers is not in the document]
ruanxin: done

pkg/scheduler/worker/workerpool_test.go
```diff
@@ -87,6 +88,63 @@ func TestWorkerPool(t *testing.T) {
...
+	//cleanup created cluster
+	defer func() {
+		require.NoError(t, inventory.Delete(clusterState.Cluster.RuntimeID))
+	}()
+
+	//create test invoker to be able to verify invoker calls
+	testInvoker := &testInvoker{}
+
+	//create reconciliation for cluster
+	testInvoker.reconRepo = reconciliation.NewInMemoryReconciliationRepository()
```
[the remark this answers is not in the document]
ruanxin: done

pkg/scheduler/worker/workerpool_test.go
ruanxin: done

pkg/scheduler/worker/workerpool_test.go
ruanxin: done

PR #596, opened 2021-12-17
Wait for all components

tobiscr:
Edge cases we have to consider after switching to cluster error-retry:

Only set clusters to error if there is no other task in progress.. otherwise it can happen that the cluster gets reconciled even if other tasks are still in progress.


pkg/scheduler/service/result.go
```diff
@@ -70,18 +70,23 @@ func (rs *ReconciliationResult) GetResult() model.Status {
...
 	}
-	if len(rs.error) > 0 {
-		if isDelete {
-			return model.ClusterStatusDeleteError
-		}
-		return model.ClusterStatusReconcileError
-	}
+	//this if-clause has always to be evaluated first:
+	//if one operation is not in a final state, the cluster is still in reconciling-state
 	if len(rs.other) > 0 {
```
tobiscr, line 75: Check first for errors, and if an error exists check if there is no other task in progress 

If YES: RECONCILING
If NO: check that all remaining tasks are in state "NEW
   If YES: ERROR
   If NO: RECONCILING

clebs (approved):

Issue #609, opened 2021-12-21
Reconciler uses per default evaluation profile


veichtj:
**Description**

When I install Kyma 2.0 with reconciler without passing a profile, reconciler uses per default the evaluation profile.
The Kyma cli has a different behaviour, it just uses the default values out of the values.yaml files.

**Expected result**

Reconciler and Kyma cli have the same behaviour if no profile is passed.

**Actual result**

If no profile is passed then evaluation profile is used.

**Steps to reproduce**

-cli: `kyma install --source=1-24.8`
-reconciler: `./bin/mothership-darwin local --value=global.domainName=some.domain.com`

**Troubleshooting**

N/A

Tomasz-Smelcerz-SAP:
By "When I install Kyma 2.0 with reconciler" I assume we're reffering to the mentioned `mothership-darwin local` command.  This command is fixed with #611 and it no longer uses the evaluation profile.
As for pipelines we are using evaluation profile explicitly for Kyma1 installation, and the same profile is used for reconciling Kyma 2.0 with reconciler - it's embedded in the mothership configuration files.
Please also take a look at: https://github.com/kyma-incubator/reconciler/issues/620

PR #611, opened 2021-12-21
#609 - default value in case no profile is passed

veichtj:
Let profile empty in case of none is passed
#609


tobiscr (approved):

tobiscr:
@ruanxin  will support @Tomasz-Smelcerz-SAP to ensure the "evaluation" profile will be used in reconciler pipelines everywhere.

Issue #619, opened 2021-12-22
Ory reconciler does not wait for hydra-maester deployment to start


veichtj:
**Description**

The ORY component reconciler triggers a hydra-maester deployment rollout in case of the need for synchronization of OAuth2clients is needed (only in evaluation profile).
The rollout himself is triggered but the reconciler does not wait for this rollout to be finished, and returns a success.
If one directly after that runs fast-integration tests, like it is done in this pipeline, the tests can fail as the OAuth2Clients might not be synced: 
https://storage.googleapis.com/kyma-prow-logs/logs/periodic-main-kyma-incubator-reconciler-kyma1-kyma2-upgrade/1473609312612913152/build-log.txt

**Expected result**

ORY component reconciler waits for the hydra-maester deployment to be running, after it triggered a rollout 

**Actual result**

ORY component reconciler returns success after triggering rollout.

**Steps to reproduce**

- install kyma 1.24.8 with kyma cli
- run fast-integration-test upgrade-test-prep + test
- use reconciler to upgrade the cluster to kyma 2.0
- run fast-integration-test upgrade-test-tests again 
they will fail with the error that a token can not be fetched for (not yet) synced OAuth2Client

**Troubleshooting**

veichtj:
PR : https://github.com/kyma-incubator/reconciler/pull/621

PR #621, opened 2021-12-22
#619 - wait for hydra-maester deployment to be running

veichtj:
In case hydra-maester deployment is rolled out to trigger OAuth2Client synchronization, reconciler needs to wait till deployment is up. Otherwise it might happen that following tests not running successfully as the clients are not yet available


pkg/reconciler/instances/ory/hydra/hydra.go
```diff
@@ -2,12 +2,12 @@ package hydra
```
raypinto, line 2: How about renaming this file and folder to syncer or hydraSyncer?

pkg/reconciler/instances/ory/k8s/rollout.go
```diff
@@ -0,0 +1,66 @@
...
+//go:generate mockery --name=RolloutHandler --outpkg=mock --case=underscore
+//RolloutHandler exposes functionality to rollout k8s objects
+type RolloutHandler interface {
+	//Rollout a given deployment and wait till it successfully up
+	RolloutAndWaitForDeployment(ctx context.Context, deployment, namespace string, client internalKubernetes.Client, logger *zap.SugaredLogger) error
+}
+
+type DefaultRolloutHandler struct{}
+
+func NewDefaultRolloutHandler() *DefaultRolloutHandler {
```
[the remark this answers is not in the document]
veichtj: Well sure a comment could be added, but the `NewDefaultRolloutHandler ` is just a constructor without any additional functionality, so the comment would basically repeat that what the name says

pkg/reconciler/instances/ory/hydra/hydra.go
veichtj: Yes, good proposal, I've renamed it

raypinto (approved):

strekm (approved):
/meow

suleymanakbas91 (approved):

Issue #268, opened 2021-10-14
Kitbag: simple simulation of failing reconciliations / operations


tobiscr:
**Description**

In cases of failing reconciliation or failing operations, our operational tooling (KCP CLI + Reconciler CLI) should support the analysis as much as possible.

**=> Use Kitbag for applying it**

Feature for the Reconciler CLI:

* The Reconciler CLI has support the execution of a reconciliation run by the cluster configuration as parameter (e.g. by enriching the `./bin/mothership mothership local` command to accept provided `cluster.State` definitions), like `./bin/mothership mothership local --cluster '{json-definition-of-the-cluster-state}' --kubeconfig <stringToKubeconfig> ...` (cluster can also be passed as file if's simpler to handle)

Feature in Kitbag:

* retrieve cluster entity as JSON from KCP CLI and store it as file in temporary folder
* if user wants to simulate just a single operation, drop all other components from the cluster JSON
* show the user the JSON and ask whether it contains any sensitive configuration data: if NO, proceed otherwise open VIM for editing it
* call `mothership local` command and pass cluster entity as JSON to it

AC:
* Kitbag provides a convenient way to simulate locally a failed reconciliation by providing just the scheduling-ID
* Kitbag provides a convenient way to simulate locally a failed operation by providing just the correlation-ID
* User can verify the retrieve cluster-JSON for sensitive data before it's triggering the local reconciliation run

**Reasons**

Simplify debugging and reproduce failing reconciliations/operations.

**Attachments**

kwiatekus:
@tobiscr Any ideas fro cluster specific config?

I.e domainName configuration will not be apliable on SRE troubleshhoting runtime

kwiatekus:
As agreed last friday we will try to include [kitbag](https://github.tools.sap/I539990/kitbag/tree/master) in the tooling for operators. 
@tobiscr 
I wonder if we should wait with this one until kcp-cli is already kitbag-enabled module

jeremyharisch:
All PRs are merged, can this be clsoed?

PR #632, opened 2021-12-29
Mothership local with clusterstatus

pPrecel:
Description

Changes proposed in this pull request:

- Allow users to use a custom `Cluster State` as a template to reconcile cluster

**Related issue(s)**

- https://github.com/kyma-incubator/reconciler/issues/268

### DIFF
--- a/cmd/mothership/local/options.go
+++ b/cmd/mothership/local/options.go
@@ -188,7 +188,6 @@ func (o *Options) Validate() error {
 		}
 
 		o.clusterState = string(b)
-		fmt.Printf("PIPE DETECTED!!!!!! - %s \n", o.clusterState)
 	}
 
 	if o.kubeconfigFile == "" {
--- a/cmd/mothership/local/options.go
+++ b/cmd/mothership/local/options.go
@@ -182,13 +182,13 @@ func (o *Options) Validate() error {
 	}
 
 	if isInputFromPipe() {
-		fmt.Println("PIPE DETECTED!!!!!!")
 		b, err := ioutil.ReadAll(os.Stdin)
 		if err != nil {
 			return err
 		}
 
 		o.clusterState = string(b)
+		fmt.Printf("PIPE DETECTED!!!!!! - %s \n", o.clusterState)
 	}
 
 	if o.kubeconfigFile == "" {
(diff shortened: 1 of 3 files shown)


cmd/mothership/local/options.go
```diff
@@ -112,33 +113,66 @@ func componentsFromStrings(list []string, values []string) ([]*keb.Component, er
 	return comps, nil
 }
 
+func componentsFromClusterState(cluster cluster.State, values []string) ([]*keb.Component, error) {
```
jeremyharisch, line 116: It would be nice to have a unit test for this function

jeremyharisch (commented):
It would be great to cover the options.go with unit tests, other than that it looks good

cmd/mothership/local/options.go
```diff
@@ -112,33 +113,66 @@ func componentsFromStrings(list []string, values []string) ([]*keb.Component, er
...
+		}
+	}
+
+	components := cluster.Configuration.Components
+	for i := range components {
+		name := components[i].Component
+		if vals[name] != nil {
+			mapValue, ok := vals[name].(map[string]interface{})
+			if ok {
+				for key, value := range mapValue {
```
mjakobczyk, line 131: Second inner loop with double `if` statement may be hard to read - what about decomposing it into smaller functions?

cmd/mothership/local/options.go
mjakobczyk, line 116: What does this function do? It creates components from cluster state, removes them, sorts them? Some verb is missing here

pPrecel: The name of this function is consistent with other functions with similar responsibility (in the same go file) like `componentsFromFile` and `componentsFromStrings`

pPrecel: good idea! thanks

cmd/mothership/local/options.go
pPrecel: This part of the code is almost the same as the [origin](https://github.com/kyma-incubator/reconciler/pull/632/files#diff-7aa83bd90c022a0af21cae41a9a134bff0edb05889da94b4d544b2e558c0cd54L93-L102) on which I based my code. I think that consistency is more important in this context. Later we can think about improving both parts

jeremyharisch (dismissed):

jeremyharisch (approved):

Issue #567, opened 2021-12-13
Refactoring: Rename preComponent to priorityList and improve priority handling in mothership reconciler


tobiscr:
**Description**

Any kind of component can have a higher priority than others without being a pre-component. The configuration of the mothership-reconciler should provide a more expressive and flexible component priority handling.

The configurable priority-list in the mothership-reconciler should support a defined order but also a 'wildcard' priority group:

         priorityList:
          - [cleanup]
          - [CRDs]
          - [cluster-essentials, istio-configuration, certificates]
          - [*]
          - [deleteCRs]

Components which are not explicitly listed in one of the priority-groups have to be added to the wildcard group. If no wildcard group is defined, the wildcard group will implicitly be added at the end of the priority list.

Example:

         priorityList:
          - [a]
          - [b]
          - [c, d, e]
          
If no wilcard group was defined in the priorityList, it will be implicitly added to the end:
          
         priorityList:
          - [a]
          - [b]
          - [c, d, e]
          - [*]

**Reasons**

Make pre-components more self-explaining and provide a higher flexibility on defining the reconciliation order of components.

**Attachments**

janmedrek:
Functionalities of the Reconciler were migrated to the new architecture, this issue will not be worked on.

PR #635, opened 2021-12-29
Order Correction

ralikio:
Correcting order of service-manager-proxy so that it is installed after service catalog component.

### DIFF
--- a/pkg/scheduler/reconciliation/persistent.go
+++ b/pkg/scheduler/reconciliation/persistent.go
@@ -95,8 +95,14 @@ func (r *PersistentReconciliationRepository) CreateReconciliation(state *cluster
 		for idx, components := range reconSeq.Queue {
 			priority := idx + 1
 			for _, component := range components {
+
+				componentPriority := priority
+				if component.Component == "service-manager-proxy" {
+					componentPriority++
+				}
+
 				createOpQ, err := db.NewQuery(tx, &model.OperationEntity{
-					Priority:      int64(priority),
+					Priority:      int64(componentPriority),
 					SchedulingID:  reconEntity.SchedulingID,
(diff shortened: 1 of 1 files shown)


clebs (changes requested):
Hi @ralikio,
just had a look at this and I think this approach is probably not the best.
IMHO hardcoding some sort of dependency between components in the reconciler is not the way to go. We have the requirement to not have any dependency between components, except for a few select ones in the prerequisites.
Additionally, this is a sort of "magic" buried deep in the reconciler code which will confuse users and make maintenance and future development much harder.

I would suggest that we discuss this issue together to find a better solution.

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -95,8 +95,14 @@ func (r *PersistentReconciliationRepository) CreateReconciliation(state *cluster
 		for idx, components := range reconSeq.Queue {
 			priority := idx + 1
 			for _, component := range components {
+
+				componentPriority := priority
+				if component.Component == "service-manager-proxy" {
```
tobiscr, line 100: Sorry, hard-coded component-names is a no-go. 

Components should not have dependencies and installation-order is only allowed for a very small set of components (like istio, cluster-essentials).

Why has the service-manager-proxy to be installed at the end of the installation?

=> Please see https://github.com/kyma-incubator/reconciler/issues/567 - this is about dealing with enhanced ordering.

But in general: a Kyma component (except if if's a very special mandatory component) should not care when it get's installed. We are retrying installations and a component should rely on the retry.

tobiscr (changes requested):

ralikio:
Closing in agreement to @tobiscr comments.

PR #636, opened 2022-01-03
Wait for the Hydra rollout after an update

cnvergence:


pkg/reconciler/instances/ory/action.go
```diff
@@ -47,13 +49,22 @@ type postDeleteAction struct {
 var (
 	jwksNamespacedName = types.NamespacedName{Name: "ory-oathkeeper-jwks-secret", Namespace: oryNamespace}
 	dbNamespacedName   = types.NamespacedName{Name: "ory-hydra-credentials", Namespace: oryNamespace}
+	rolloutHydra       = false
```
mjakobczyk, line 52: How about getting rid of this var and read data from cluster dynamically? Such statefulness could harm us.

pkg/reconciler/instances/ory/action.go
```diff
@@ -21,10 +21,11 @@ import (
 )
 
 const (
-	oryChart     = "ory"
-	oryNamespace = "kyma-system"
-	jwksAlg      = "RSA256"
-	jwksBits     = 2048
+	oryChart        = "ory"
+	oryNamespace    = "kyma-system"
```
mjakobczyk, line 25: Do we really need this const? Ory reconciliation should take place in namespace in which it is said to. From `Context` API perspective it can be retrieved in the following way: `context.Task.Namespace`

pkg/reconciler/instances/ory/action.go
```diff
@@ -113,19 +127,16 @@ func (a *preReconcileAction) Run(context *service.ActionContext) error {
 			return errors.Wrap(err, "failed to update db credentials data for Ory Hydra")
 		}
 
-		if len(newSecretData) == 0 {
+		if !isUpdate(newSecretData) {
 			logger.Info("Ory DB secret is the same as values, no need to update")
 		} else {
 			logger.Info("Ory DB secret is different than values, updating it")
 			dbSecretObject.StringData = newSecretData
+			rolloutHydra = true
```
mjakobczyk, line 135: If we want to deal with it this way we should also do opposite and set it to `false` when it is not an update.

pkg/reconciler/instances/ory/action.go
```diff
@@ -194,15 +201,12 @@ func (a *preReconcileAction) updateSecret(ctx context.Context, client kubernetes
...
-	_, err := client.AppsV1().Deployments("kyma-system").Patch(ctx, "ory-hydra", types.StrategicMergePatchType, []byte(data), metav1.PatchOptions{})
-
+func (a *postReconcileAction) rolloutHydraDeployment(ctx context.Context, client internalKubernetes.Client, deployment string, logger *zap.SugaredLogger) error {
+	err := a.rolloutHandler.RolloutAndWaitForDeployment(ctx, deployment, oryNamespace, client, logger)
 	if err != nil {
-		return errors.Wrap(err, "Failed to rollout ory hydra")
+		return errors.Wrapf(err, "Failed to rollout %s deployment", deployment)
 	}
-	logger.Info("ory-hydra restarted")
+	logger.Infof("%s deployment restarted", deployment)
```
mjakobczyk, line 209: We used the phrase `Failed to rollout` in a negative scenario. How about sticking to it? In this case you could call it: `(...) deployment rolled out` or to be fully compliant: `Performed rollout restart of deployment (...)`.

pkg/reconciler/instances/ory/action_test.go
```diff
@@ -517,7 +491,87 @@ func TestOryDbSecret(t *testing.T) {
 		})
 	}
 }
+func TestOryDbSecret_IsUpdate(t *testing.T) {
+	t.Run("should return false when applying the same secret", func(t *testing.T) {
+		// given
+		name := types.NamespacedName{Name: "test-db-secret", Namespace: "test"}
+		ctx := context.Background()
+		k8sClient := fake.NewSimpleClientset()
+		logger := zaptest.NewLogger(t).Sugar()
```
mjakobczyk, line 500: You can freely move this logger initialisation above all the test cases and reuse it.

pkg/reconciler/instances/ory/action_test.go
```diff
@@ -517,7 +491,87 @@ func TestOryDbSecret(t *testing.T) {
 		})
 	}
 }
+func TestOryDbSecret_IsUpdate(t *testing.T) {
+	t.Run("should return false when applying the same secret", func(t *testing.T) {
+		// given
+		name := types.NamespacedName{Name: "test-db-secret", Namespace: "test"}
+		ctx := context.Background()
```
mjakobczyk, line 498: You can freely move this context initialisation above all the test cases and reuse it.

pkg/reconciler/instances/ory/action_test.go
```diff
@@ -517,7 +491,87 @@ func TestOryDbSecret(t *testing.T) {
 		})
 	}
 }
+func TestOryDbSecret_IsUpdate(t *testing.T) {
+	t.Run("should return false when applying the same secret", func(t *testing.T) {
+		// given
+		name := types.NamespacedName{Name: "test-db-secret", Namespace: "test"}
```
mjakobczyk, line 497: You can freely move this namespaces name initialisation above all the test cases and reuse it.

pkg/reconciler/instances/ory/action_test.go
```diff
@@ -517,7 +491,87 @@ func TestOryDbSecret(t *testing.T) {
...
+		ctx := context.Background()
+		k8sClient := fake.NewSimpleClientset()
+		logger := zaptest.NewLogger(t).Sugar()
+		component := chart.NewComponentBuilder("main", componentName).
```
mjakobczyk, line 501: You can freely move this component initialisation above all the test cases and reuse it.

pkg/reconciler/instances/ory/action_test.go
```diff
@@ -517,7 +491,87 @@ func TestOryDbSecret(t *testing.T) {
 		})
 	}
 }
+func TestOryDbSecret_IsUpdate(t *testing.T) {
```
mjakobczyk, line 494: I have a problem with this testing function. Is it what you really want to test? The implementation of the function is as simple as this:
```go
func isUpdate(diff map[string]string) bool {
	return len(diff) != 0
}
```
I would imagine test scenarios where the tested function retrieves some crafted input of type `map[string]string` and nothing else. Whereas here you do many things, including retrieval, creation and update of a secret. If it was supposed to be the integration test for the action (which one in fact?) then you should reconsider the test name and not imply that what is really tested is `isUpdate()` function because now it is not. If you unit test it in some other place you could in fact use it as an assertion and expect that it says that the updated is needed or not. For now I feel a mixed responsibility of the testing function. Could you somehow adjust it?

mjakobczyk (changes requested):

pkg/reconciler/instances/ory/action.go
cnvergence: This would require more refactoring, as it is used in most of those actions

pkg/reconciler/instances/ory/action.go
cnvergence: Currently, we are limited by the db.Update method, we should open the new issue and work on the new Update logic.

pkg/reconciler/instances/ory/action_test.go
cnvergence: I think I wanted to test the update logic, because there were no such tests in place before. I will rename this function to not being considered as a test for the isUpdate method.

mjakobczyk (approved):
LGTM

PR #641, opened 2022-01-03
Adding compass-runtime-agent and application-connector as pre components

koala7659:
In order to be successfully installed `compass-runtime-agent` requires `application-connector` component to be installed first.


jakobmoellersap:
@koala7659 still valid?

pbochynski:
It is not going to be merged.

PR #646, opened 2022-01-04
Fix helm chart pathing bug

wozniakjan:
Installing component via regular helm chart created by `helm pack` fails
```
Running=btp-operator[4][error: Failed to get manifest for component 'btp-operator' in Kyma version '2.0.0' using repository 'https://github.com/kyma-incubator/sap-btp-service-operator/releases/download/v0.1.18-custom/sap-btp-operator-0.1.18.tar.gz' : stat /tmp/reconciler/dcdb6d7dd915d574d31e7ffd286d16921cee1fe0-btp-operator/btp-operator/Chart.yaml: no such file or directory
```

However, running the chart by `helm` works well
```
$ helm template https://github.com/kyma-incubator/sap-btp-service-operator/releases/download/v0.1.18-custom/sap-btp-operator-0.1.18.tar.gz
```

This is an effort to enhance support for any valid helm chart by recursively looking for `Chart.yaml` instead of having the path hardcoded.

closes: https://github.com/kyma-incubator/reconciler/issues/527

### DIFF
--- a/pkg/reconciler/chart/component.go
+++ b/pkg/reconciler/chart/component.go
@@ -15,6 +15,10 @@ type Component struct {
 	configuration map[string]interface{}
 }
 
+func (c *Component) isExternalComponent() bool {
+	return c.url != ""
+}
+
 func (c *Component) isExternalGitComponent() bool {
 	return strings.HasSuffix(c.url, ".git")
 }
--- a/pkg/reconciler/chart/helm.go
+++ b/pkg/reconciler/chart/helm.go
@@ -2,6 +2,7 @@ package chart
 
 import (
 	"fmt"
+	"io/fs"
 	"path/filepath"
 	"strings"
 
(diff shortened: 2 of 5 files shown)


pkg/reconciler/chart/helm.go
```diff
@@ -31,20 +32,44 @@ func NewHelmClient(chartDir string, logger *zap.SugaredLogger) (*HelmClient, err
...
+	if component.isExternalComponent() {
+		path := ""
+		err := filepath.WalkDir(c.chartDir, func(p string, d fs.DirEntry, err error) error {
+			if filepath.Base(p) == "Chart.yaml" && path == "" {
+				path = p
+			}
+			return nil
+		})
+		if err != nil {
+			return "", err
```
wozniakjan, line 45: I don't think this can happen because the `WalkDirFunc` doesn't return an error ever, but linter was complaining...

pkg/reconciler/chart/component.go
```diff
@@ -15,6 +15,10 @@ type Component struct {
 	configuration map[string]interface{}
 }
 
+func (c *Component) isExternalComponent() bool {
```
wozniakjan, line 18: this is used to figure out path of the `Chart.yaml` for external components. The reason is that native kyma components have different pathing structure and for those, the hardcoded path is fine
```
/tmp/reconciler/2.0.0/[component_name]/Chart.yaml
```
for external components, we don't always have full control over how it's packed, the `Chart.yaml` might be at root level or nested few folders deep, so the hardcoded path may not suffice
```
/tmp/reconciler/[hash]-[component_name]/[component_name]/Chart.yaml
```
Instead for those I would like to propose recursive search for `Chart.yaml` starting at
```
/tmp/reconciler/[hash]-[component_name]/
```

jeremyharisch (approved):

Issue #637, opened 2022-01-03
POC integration test for ORY


strekm:
**Description**

Create simple integration test listing ORY pods. Test should assume that cluster is provisioned and ORY stack installed. Use kube client to list pods. 

**Attachments**

https://github.com/kyma-project/kyma/issues/12968

PR #648, opened 2022-01-05
Initial Integration test for Ory reconciler

cnvergence:
### Description
- introduce exemplary integration test for Ory reconciler

### Links
See: #637


Makefile
```diff
@@ -67,6 +67,11 @@ test:
 test-all: export RECONCILER_INTEGRATION_TESTS = 1
 test-all: test
 
+.PHONY: test-ory
+test-ory: export ORY_RECONCILER_INTEGRATION_TESTS=1
+test-ory:
```
mjakobczyk, line 72: Is this target duplication required? Can't we have a list of commands to be executed?

Makefile
```diff
@@ -67,6 +67,11 @@ test:
...
+		go test -v -timeout 5m ./pkg/reconciler/instances/ory/test
```
mjakobczyk, line 73: This formatting is strange - looks different when comparing to the rest of the file.

pkg/reconciler/instances/ory/test/ory_integration_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/kyma-incubator/reconciler/pkg/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubectl/pkg/util/podutils"
+)
+
+const (
+	envOryIntegrationTests = "ORY_RECONCILER_INTEGRATION_TESTS"
```
mjakobczyk, line 21: This constant only refers to the key of the env, not the value, so maybe you could include `key` keyword here?

pkg/reconciler/instances/ory/test/ory_integration_test.go
```diff
@@ -0,0 +1,63 @@
...
+	namespace              = "kyma-system"
+)
+
+func OryIntegrationTest(t *testing.T) {
```
mjakobczyk, line 25: The name of the function does not reflect what it does. In fact it does not perform any test but checks for env to optionally skip the suite. What about changing the name to `skipTestIfDisabled()`?

pkg/reconciler/instances/ory/test/ory_integration_test.go
```diff
@@ -0,0 +1,63 @@
...
+	if !isIntegrationTestEnabled() {
+		t.Skipf("Integration tests disabled: skipping parts of test case '%s'", t.Name())
+	}
+}
+
+func isIntegrationTestEnabled() bool {
+	expensiveTests, ok := os.LookupEnv(envOryIntegrationTests)
```
mjakobczyk, line 32: What does `expensiveTests` mean in this context? Did you mean something like `testEnabled`?

pkg/reconciler/instances/ory/test/ory_integration_test.go
```diff
@@ -0,0 +1,63 @@
...
+	return ok && (expensiveTests == "1" || strings.ToLower(expensiveTests) == "true")
```
mjakobczyk, line 33: Why do you mix conditions? You verbosely set the env value to `1` so it is what should be expected. Envs are strings so they will not be converted into boolean out-of-the-box.

pkg/reconciler/instances/ory/test/ory_integration_test.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func TestOryIntegration(t *testing.T) {
+	OryIntegrationTest(t)
+
+	log := logger.NewLogger(false)
```
mjakobczyk, line 39: I can see a pattern here. Each test suite would contain the same set of properties such as: logger, kubeClient, context and metadata. What about preparing this in a separate function? For now the initialisation part is longer than a test logic itself so we could move this boilerplate elsewhere and refuse later. Maybe an introduction of a helper struct could become handy here so whenever we create the struct it will be created from scratch to not be reused between test cases?

pkg/reconciler/instances/ory/test/ory_integration_test.go
```diff
@@ -0,0 +1,63 @@
...
+	options := metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/instance=ory",
+	}
+
+	podsList, err := client.CoreV1().Pods(namespace).List(ctx, options)
+	require.NoError(t, err)
+
+	for i, pod := range podsList.Items {
+		log.Infof("Pod %v is deployed", pod.Name)
+		assert.Equal(t, v1.PodPhase("Running"), pod.Status.Phase)
```
mjakobczyk, line 59: I do not like mixing hard `require` with soft `assert`. What about choosing only one?

mjakobczyk (changes requested):

Makefile
cnvergence: it's target-specific variable, for the makefile this is acceptable.

pkg/reconciler/instances/ory/test/ory_integration_test.go
cnvergence: I would leave it as env, for the environment variable, so the name ORY_RECONCILER_INTEGRATION_TESTS fits here.

pkg/reconciler/instances/ory/test/ory_integration_test.go
cnvergence: Makes sense, we should refactor this code from reconciler to our use case.

pkg/reconciler/instances/ory/test/ory_integration_test.go
cnvergence: Let's start with this structure, indeed.

mjakobczyk (approved):
After changes LGTM 💯

Makefile
```diff
@@ -67,6 +67,11 @@ test:
 test-all: export RECONCILER_INTEGRATION_TESTS = 1
 test-all: test
 
+.PHONY: test-ory
+test-ory: export ORY_RECONCILER_INTEGRATION_TESTS = 1
```
tobiscr, line 71: Is there a need to run the ORY reconciler not together with all other integration tests?

cnvergence: Currently we want to run those tests only when Ory component changes, we will soon include more test cases.
We need it as Ory keeps on having new issues.

tobiscr: alright

tobiscr (approved):

Makefile
tobiscr: But please open an issue to drop this part after we migrated to Envoy, would that be possible?

Issue #647, opened 2022-01-05
Delete: Finalisers still get stuck and gardener can't deprovision


clebs:
**Description**

After adding the temporary cleanup after deletion it is still not working as expected. Resources need to be removed so that finalisers do not get stuck.

**Expected result**
After delete reconciliation gardener can deprovision the cluster without issues.

**Actual result**
Gardener can not deprovison the cluster because of stuck finalisers/namespaces.

**Steps to reproduce**

1. Provision a gardener cluster
2. Install Kyma via reconciler (e.g. `kyma deploy`)
3. Delete Kyma via reconciler (e.g. `kyma undeploy`)
4. Delete the cluster on gardener and check if it gets stuck.

**Troubleshooting**
Suggested way to find out the resources that block the deprovisioning:
- Do the steps to reproduce
- When deprovisioning gets stuck, check on gardener UI which will tell what resurce is blocking the deprovisioning.

Tomasz-Smelcerz-SAP:
I have found the following error in Gardener using the procedure:
```
Flow "Shoot cluster deletion" encountered task errors: [task "Cleaning Kubernetes resources" failed: 1 error occurred:
	* retry failed with context deadline exceeded, last error: remaining objects are still present: [*v1.PersistentVolumeClaim kyma-system/data-helm-broker-etcd-stateful-0 *v1.PersistentVolumeClaim kyma-system/data-ory-postgresql-0 *v1.PersistentVolumeClaim kyma-system/prometheus-monitoring-prometheus-db-prometheus-monitoring-prometheus-0 *v1.PersistentVolumeClaim kyma-system/storage-logging-loki-0]
```

tobiscr:
Conclusion: detect orphan resources and remove them

1. query for all resources which are terminating for a particular time
2. delete these resources by dropping their finalizers

Tomasz-Smelcerz-SAP:
Another error to consider, this one may be transient:
```
Action required
 Cleaning up the cluster failed as some resource are stuck in deletion. Please remove these resources properly or a forceful deletion will happen if this error persists. 

task "Cleaning extended API groups" failed: 1 error occurred:
[...]

PR #669, opened 2022-01-13
improve cleanup process

Tomasz-Smelcerz-SAP:
**Description**
Added a step of explicit removal  of PersistentVolumeClaims remaining in the 'kyma-system' namespace.
Such PVCs prevent the namespace from termination and make the cluster deprovisioning blocked in Gardener.
The new step happens before anything else and is "blocking" (i.e. no concurrent), but usually it takes no more than 3 to 5 seconds.

In addition:
- eliminate the code that tries to remove the 'kyma-system` namespace finalizer. It seems unnecessary and getting rid of it actually reduces undeploy time from 9 to 3.5 minutes.
- small enhancement to Istio reconciler error log message

See: https://github.com/kyma-incubator/reconciler/issues/647


pkg/reconciler/instances/istio/bootstrap.go
```diff
@@ -80,7 +80,7 @@ func newDefaultCommanderResolver(paths []string, log *zap.SugaredLogger) (action
 func parsePaths(input string, isValid func(string) error) ([]string, error) {
 	trimmed := strings.TrimSpace(input)
 	if trimmed == "" {
-		return nil, errors.Errorf("No paths defined")
+		return nil, errors.Errorf("%s env variable is undefined or empty", istioctlBinaryPathEnvKey)
```
clebs, line 83: Nice improvement!

pkg/reconciler/instances/cleaner/pkg/cleanup/pvc.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	var attempts uint = 3
+	delay := time.Second * 2
+	err := retry.Do(retryFunc, retry.Attempts(attempts), retry.Delay(delay))
+	if err != nil {
+		return err
+	}
+
+	return nil
```
clebs, line 39:
```suggestion
	return retry.Do(retryFunc, retry.Attempts(3), retry.Delay(time.Second * 2))
```

tobiscr (approved):

veichtj (approved):

PR #691, opened 2022-01-18
Refactor helmclient deploy delete logic

ruanxin:


tobiscr (approved):

tobiscr (dismissed):

pkg/reconciler/kubernetes/adapter.go
```diff
@@ -299,11 +301,14 @@ func getRestConfig(kubeconfig string) (*rest.Config, error) {
 	})
 }
 
-func (g *kubeClientAdapter) convertToInfoList(unstructs []*unstructured.Unstructured, namespaceOverride string) ([]*resource.Info, error) {
+func (g *kubeClientAdapter) filterAndConvertToInfoList(unstructs []*unstructured.Unstructured, namespaceOverride string, ignoreNotMatchError bool) ([]*resource.Info, error) {
```
tobiscr, line 304: We should maybe add a comment what the "filter" means... based on the code, it's not obvious that the filter related to the `ignoreNoMatchError` flag

tobiscr (approved):

pkg/reconciler/kubernetes/adapter.go
ruanxin:
```suggestion
// During the conversion from unstructured.Unstructured to resource.Info, some resource.Info might not be 
// constructed due to the dependent resources are not been deployed, For example, CR related CRD is missing, 
// and NoResourceMatchError or NoKindMatchError will be thrown, use this function to determine if we can ignore
// those errors and filtering resources in the result list during conversion by ignoreNotMatchError param.
func (g *kubeClientAdapter) filterAndConvertToInfoList(unstructs []*unstructured.Unstructured, namespaceOverride string, ignoreNotMatchError bool) ([]*resource.Info, error) {
```

PR #695, opened 2022-01-18
Do not remove the eventing-nats service together with old nats-operator resources

VladislavPaskar:
**Changes proposed in this pull request**

We are deleting the old nats-operator resources upon each reconciliation run. The deletion step defines the resource by name, kind and namespace. The problem is, that the new nats helm charts have the same service as the old one, which should be deleted after 2.x. This causes the deletion of the actual service which is not needed and causes downtime of the nats backend. 
This PR checks if the cluster is already in the 2.x state by checking whether the stateful set is deployed (the old nats-operator was deployed using CR, but not statefulset). If the statefulset is deployed, do not remove the service, as it is the actual one.

The perfect approach would be to check the reconciled and the cluster versions, but as it's currently not provided by the reconciler's team, so the approach from this PR should work for now

**Related issues**
https://github.com/kyma-project/kyma/issues/13042


pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -90,10 +93,47 @@ func (r *removeNatsOperatorStep) removeNatsOperatorResources(context *service.Ac
...
 	manifest.Manifest = strings.ReplaceAll(manifest.Manifest, natsSubChartPath, eventingNats)
 
+	clientSet, err := kubeClient.Clientset()
+	tracker, err := progress.NewProgressTracker(clientSet, logger, progress.Config{Interval: progressTrackerInterval, Timeout: progressTrackerTimeout})
+	if err != nil {
+		return err
+	}
+
 	logger.Info("Removing nats-operator chart resources")
 	// remove all the existing nats-operator resources, installed via charts
```
mfaizanse, line 103: This comment is not relevant to the next line.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -90,10 +93,47 @@ func (r *removeNatsOperatorStep) removeNatsOperatorResources(context *service.Ac
...
+		// do not delete the nats service if there is a statefulSet deployed
+		if statefulSet != nil && u.GetName() == eventingNats && strings.EqualFold(u.GetKind(), serviceKind) {
+			continue
+		}
+
+		//if resource is watchable, add it to progress tracker
+		watchable, err := progress.NewWatchableResource(u.GetKind())
+		if err == nil { //add only watchable resources to progress tracker
+			tracker.AddResource(watchable, namespace, u.GetName())
+		}
```
mfaizanse, line 122: Print the error if `err != nil` so that later we can know which resources were not watched.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -32,9 +36,19 @@ func TestDeletingNatsOperatorResources(t *testing.T) {
 
 	// ensure the right calls were invoked
 	mockProvider.AssertCalled(t, "RenderManifest", mockedComponentBuilder)
-	k8sClient.AssertCalled(t, "Delete", actionContext.Context, manifestString, namespace)
+	m := []byte(manifestString)
+	unstructs, err := kubernetes.ToUnstructured(m, true)
+	require.NoError(t, err)
+	for _, u := range unstructs {
+		if u.GetName() == eventingNats && strings.EqualFold(u.GetKind(), serviceKind) {
+			k8sClient.AssertNotCalled(t, "DeleteResource", actionContext.Context, u.GetKind(), u.GetName(), namespace)
```
mfaizanse, line 44: (nit) The `TestDeletingNatsOperatorResources` only covers the scenario where the statefulset exists and the service should not be deleted.
Maybe add another test which covers the other scenario where the statefulset does not exists and the service should BE deleted.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -90,10 +93,47 @@ func (r *removeNatsOperatorStep) removeNatsOperatorResources(context *service.Ac
...
 	// remove all the existing nats-operator resources, installed via charts
-	_, err = kubeClient.Delete(context.Context, manifest.Manifest, namespace)
-	return err
+	statefulSet, err := getStatefulSet(context, kubeClient, eventingNats)
+	if err != nil {
+		return err
+	}
+
+	m := []byte(manifest.Manifest)
+	unstructs, err := kubernetes.ToUnstructured(m, true)
```
mfaizanse, line 110: (nit) Rename the `unstructs` to something more meaningful like `natsK8sResources` or `natsOperatorResources` etc.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -90,10 +93,47 @@ func (r *removeNatsOperatorStep) removeNatsOperatorResources(context *service.Ac
...
+		logger.Infof("Deleting: kind: %s, name: %s, namespace: %s", u.GetKind(), u.GetName(), namespace)
+		_, err = kubeClient.DeleteResource(context.Context, u.GetKind(), u.GetName(), namespace)
+		if err != nil {
+			return err
+		}
+	}
+
+	//wait until all resources were deleted
+	if err := tracker.Watch(context.Context, progress.TerminatedState); err != nil {
+		logger.Warnf("Watching progress of deleted resources failed: %s", err)
```
mfaizanse, line 133: shouldn't we `return err` here as well, because if the watch is failed, then the reconciler will go ahead and apply the new yamls, which could result in conflicting resources on the cluster.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -90,10 +93,47 @@ func (r *removeNatsOperatorStep) removeNatsOperatorResources(context *service.Ac
...
+		return err
+	}
+
+	m := []byte(manifest.Manifest)
+	unstructs, err := kubernetes.ToUnstructured(m, true)
+	for _, u := range unstructs {
+		// since the old nats-operator was deployed using the k8s deployment
+		// do not delete the nats service if there is a statefulSet deployed
+		if statefulSet != nil && u.GetName() == eventingNats && strings.EqualFold(u.GetKind(), serviceKind) {
+			continue
```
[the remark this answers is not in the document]
VladislavPaskar: I thought about this, but this more a defensive if, which fixes the wrong deletion, so basically adding the log record would mean: "everything went well here", so that's why I omitted it. I hope you're fine if I leave it

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
VladislavPaskar: removed

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
VladislavPaskar: yes, you are right, I will print it

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
VladislavPaskar: added a test case

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
VladislavPaskar: renamed

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -90,10 +93,47 @@ func (r *removeNatsOperatorStep) removeNatsOperatorResources(context *service.Ac
...
-	_, err = kubeClient.Delete(context.Context, manifest.Manifest, namespace)
-	return err
+	statefulSet, err := getStatefulSet(context, kubeClient, eventingNats)
+	if err != nil {
+		return err
+	}
+
+	m := []byte(manifest.Manifest)
+	unstructs, err := kubernetes.ToUnstructured(m, true)
+	for _, u := range unstructs {
```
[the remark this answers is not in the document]
VladislavPaskar: renamed

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
VladislavPaskar: yes, I added it

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -2,11 +2,13 @@ package preaction
 
 import (
 	"context"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/kubernetes/progress"
```
mfaizanse, line 5: (nit) group the similar imports together as in go standards.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -90,10 +93,54 @@ func (r *removeNatsOperatorStep) removeNatsOperatorResources(context *service.Ac
 	// set the right eventing name, which went lost after rendering
 	manifest.Manifest = strings.ReplaceAll(manifest.Manifest, natsSubChartPath, eventingNats)
 
+	clientSet, err2 := kubeClient.Clientset()
+	if err2 != nil {
```
mfaizanse, line 97: (nit) why not reuse `err` instead of `err2`?

VladislavPaskar: linter was complaining because of the ineffective usage of the `err`, which was already initialised before in the code

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
VladislavPaskar: done

mfaizanse (approved):

Issue #677, opened 2022-01-14
Run DB migration in init-container to avoid false-positive liveness-probes


tobiscr:
**Description**

The DB migration can take several minutes for a fully productive mothership reconciler as the amount of data can become quite huge.

In the past, we had the challenge that the K8s liveness-probe was triggered while a DB migration was ongoing and reported the mothership-pod to be down. This caused a restart of the pod while the DB migration was still ongoing, and caused finally a dirty/incomplete DB schema migration.

To avoid this issue, we decided to move the DB migration step into an init-container. The init-container is not considered by the liveness-probe, and we can avoid such colliding situations. We agreed on the following approach:

1. Mothership reconciler gets extended by a flag (e.g. `--stop-after-migrate` or similar) which lead the mothership-reconciler to stop after the DB migration is finished.
2. The K8s deployment of the mothership-reconciler gets extended by an init-container which runs the mothership-reconciler and sets also the `--stop-after-migrate` (or similar, see step 1). This will run the DB migration up-front and allows even to run custom logic between migration steps without being considered by the liveness-probe.
3. The GCP cloud_sql_proxy container gets extracted to a separate deployment and service, and the mothership reconciler configuration is adapted to point to the new service for the cloud_sql_proxy 

AC:
* Mothership reconciler has to be extended by a flag which lets the mothership-reconciler process stop after the DB migration is finished.
* GCP cloud_sql_proxy container is deployed separately, and the mothership deployment is configured to communicate with the DB using the cloud_sql_proxy service.
* An init container is added to the K8s deployment of the mothership-reconciler, which starts the mothership-reconciler and sets the new flag  to ensure the process stops after it has migrated the DB schema.

**Reasons**

Solve issue that mothership-reconciler gets restarted by failing liveness-probe if a long-running DB migration is happening.

**Attachments**

janmedrek:
Functionalities of the Reconciler were migrated to the new architecture, this issue will not be worked on.

PR #696, opened 2022-01-18
added new flag to stop after migration and improved migrate logs

khlifi411:
Part 1 of #677

### DIFF
--- a/cmd/mothership/mothership/start/options.go
+++ b/cmd/mothership/mothership/start/options.go
@@ -26,6 +26,7 @@ type Options struct {
 	AuditLog                 bool
 	AuditLogFile             string
 	AuditLogTenantID         string
+	StopAfterMigration       bool
 }
 
 func NewOptions(o *cli.Options) *Options {
@@ -44,6 +45,7 @@ func NewOptions(o *cli.Options) *Options {
 		false,           //AuditLog
 		"",              //AuditLogFIle
 		"",              //AuditLogTenant
+		false,           //StopAfterMigration
 	}
 }
 
--- a/cmd/mothership/mothership/start/options.go
+++ b/cmd/mothership/mothership/start/options.go
@@ -43,7 +43,7 @@ func NewOptions(o *cli.Options) *Options {
 		false,           //CreateEncyptionKey
 		0,               //MaxParallelOperations
 		false,           //AuditLog
-		"",              //AuditLogFIle
+		"",              //AuditLogFile
 		"",              //AuditLogTenant
 		false,           //StopAfterMigration
 	}
--- a/pkg/db/postgres.go
+++ b/pkg/db/postgres.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-
 	log "github.com/kyma-incubator/reconciler/pkg/logger"
 	"github.com/pkg/errors"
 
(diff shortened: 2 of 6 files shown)


pkg/db/postgres.go
```diff
@@ -19,6 +18,29 @@ import (
 	"go.uber.org/zap"
 )
 
+type migrateLogger struct {
+	logger  *zap.SugaredLogger
```
tobiscr, line 22: Why do we need the migrate logger?

Currently, the verbose option is not used in the code. Can't we use the regular logger?

khlifi411: We can use the regular logger, but we cannot get the logs of the steps that migrate executes against the db during the `migrate.Up()` call. I thought we are using the verbose option, that's why I made it related to the `debug` attribute of the connectionFactory. I did this because when troubleshooting the migration the last time, me and Xin have had a hard time knowing which step in the migration failed.  But if we don't need it, I can use the regular logger as before.

cmd/mothership/mothership/start/options.go
```diff
@@ -44,6 +45,7 @@ func NewOptions(o *cli.Options) *Options {
 		false,           //AuditLog
 		"",              //AuditLogFIle
```
ruanxin, line 46:
```suggestion
		"",              //AuditLogFile
```

pkg/db/postgres.go
```diff
@@ -19,6 +18,29 @@ import (
...
+
+func newMigrateLogger(debug bool) *migrateLogger {
+	logger := log.NewLogger(debug)
+	return &migrateLogger{
+		logger:  logger,
+		verbose: debug,
+	}
+}
+
+func (ml migrateLogger) Printf(format string, v ...interface{}) {
```
ruanxin, line 34:
```suggestion
func (ml *migrateLogger) Printf(format string, v ...interface{}) {
```

pkg/db/postgres.go
```diff
@@ -19,6 +18,29 @@ import (
...
+	if ml.verbose {
+		ml.logger.Debugf(format, v...)
+	}
+}
+
+func (ml migrateLogger) Verbose() bool {
```
ruanxin, line 40:
```suggestion
func (ml *migrateLogger) Verbose() bool {
```

pkg/db/postgres.go
ruanxin, line 34: There is no usage for this `Printf`

pkg/db/postgres.go
```diff
@@ -19,6 +18,29 @@ import (
...
+	return &migrateLogger{
+		logger:  logger,
+		verbose: debug,
+	}
+}
+
+func (ml migrateLogger) Printf(format string, v ...interface{}) {
+	if ml.verbose {
+		ml.logger.Debugf(format, v...)
+	}
```
ruanxin, line 37: As reminder: we have to also log something in non verbose mode

pkg/db/postgres.go
```diff
@@ -194,14 +216,14 @@ func (pcf *postgresConnectionFactory) checkPostgresIsolationLevel() error {
 }
 
 func (pcf *postgresConnectionFactory) migrateDatabase() error {
-	logger := log.NewLogger(pcf.debug)
+	migrateLogger := newMigrateLogger(pcf.debug)
```
ruanxin, line 219: As reminder after discuss: forgot to assign `m.Log = migrateLogger`

khlifi411: Thanks! Applied 👍

pkg/db/postgres.go
khlifi411: Thanks! Applied 👍

jeremyharisch (approved):

Issue #668, opened 2022-01-12
Improve housekeeping: store only latest X reconciliations and those which were failed


tobiscr:
**Description**

We are facing a huge amount of recorded operations in the DEV DB (even when just 8 clusters are continuously reconciled). The housekeeping of records for 14 days isn't sufficient as we are generating per cluster >80k entries in 2 weeks.

To address this issue of a non-scaling data amount, we have to adjust our housekeeping (see [link to sources](https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/service/cleaner.go)):

1. Housekeeping should run more frequently a day (e.g. each second hours)
2. The housekeeping has to ignore
    * the latest X reconciliations of a cluster, independent of their status (X has to be a configurable value)
    * any reconciliations within Y days if it was not successfully finished (e.g. error, error_retryable etc.)
3. Any other reconciliations inclusive their related operations have to be dropped

AC:
* Housekeeping runs multiple times a day
* It removes any reconciliation entity of a cluster except the last X reconciliations and those within Y days which were not successfully finished
* X and Y have to be configurable values

**Reasons**

Solve non-scaling DB storage consumption caused by huge amount of generated operation entries.

**Attachments**

PR #700, opened 2022-01-20
Improved db cleanup

Tomasz-Smelcerz-SAP:
**Description**
Introduced a new logic for the reconciliation cleanup - see #668. 
The new mechanism is **disabled** by default and the old logic stays to allow for safe transition to a new mechanism.
Once the environments are migrated (new configuration flags are used) the old logic along with command-line flags may be removed in a separate PR.

See: https://github.com/kyma-incubator/reconciler/issues/668


pkg/scheduler/service/cleaner.go
```diff
@@ -61,3 +125,99 @@ func (c *cleaner) purgeReconciliations(transition *ClusterStatusTransition, conf
 		}
 	}
 }
+
+func (c *cleaner) getLatestReconciliations(latestEntitiesCount int, transition *ClusterStatusTransition) ([]*model.ReconciliationEntity, error) {
+	return transition.ReconciliationRepository().GetReconciliations(&reconciliation.Limit{Count: latestEntitiesCount})
```
ruanxin, line 130: Should we set order by? otherwise, the returned results might not be the latest.

ruanxin: OK, resolved, the implementation of limit contains order by
```
func (l *Limit) FilterByQuery(q *db.Select) error {
	q.OrderBy(map[string]string{"Created": "DESC"}).Limit(l.Count)
	return nil
}

File: pkg/scheduler/service/cleaner.go [hunk 2]
```diff
@@ -34,14 +45,67 @@ func (c *cleaner) Run(ctx context.Context, transition *ClusterStatusTransition,
...
+	if err != nil {
+		c.logger.Errorf("Cleaner failed to get last %d reconciliations: %s", config.keepLatestEntitiesCount(), err.Error())
+	}
+
+	if len(latestReconciliations) < config.keepLatestEntitiesCount() {
+		//Nothing to clean up
+		return
+	}
+
+	oldestInRange := findOldestReconciliation(latestReconciliations)
```
ruanxin, line 82: I think the returned `latestReconciliations` array is already sorted by `{"Created": "DESC"}`, you can just use the last one as oldest

pkg/scheduler/service/cleaner.go
```diff
@@ -61,3 +125,99 @@ func (c *cleaner) purgeReconciliations(transition *ClusterStatusTransition, conf
...
+		return nil, err
+	}
+	return reconciliations, nil
+}
+
+func (c *cleaner) removeReconciliations(list []*model.ReconciliationEntity, transition *ClusterStatusTransition, filter func(*model.ReconciliationEntity) bool) {
+	cnt := 0
+	for i := range list {
+		id := list[i].SchedulingID
+		//TODO: does this mean "successful" ?
```
ruanxin, line 176: is this statement clear now? we can remove the TODO.

pkg/scheduler/service/cleaner.go
```diff
@@ -61,3 +125,99 @@ func (c *cleaner) purgeReconciliations(transition *ClusterStatusTransition, conf
...
+func (c *cleaner) dropSuccessfulRecordsOlderThan(t time.Time, transition *ClusterStatusTransition) error {
+	deadline := t.UTC()
+	list, err := c.findReconciliationsOlderThan(deadline, transition)
+	if err != nil {
+		return err
+	}
+
+	c.logger.Infof("[CLEANER] Removing successful records older than: %s. Records found: %d (only some must be removed)", deadline.String(), len(list))
+	if len(list) > 0 {
+		c.removeReconciliations(list, transition, func(m *model.ReconciliationEntity) bool { return m.Status.IsFinalStable() })
```
ruanxin, line 156: Can we give this anonymous function a proper name?

pkg/scheduler/service/cleaner.go
```diff
@@ -61,3 +125,99 @@ func (c *cleaner) purgeReconciliations(transition *ClusterStatusTransition, conf
 		}
 	}
 }
+
+func (c *cleaner) getLatestReconciliations(latestEntitiesCount int, transition *ClusterStatusTransition) ([]*model.ReconciliationEntity, error) {
+	return transition.ReconciliationRepository().GetReconciliations(&reconciliation.Limit{Count: latestEntitiesCount})
+}
+
+func (c *cleaner) dropRecordsOlderThan(t time.Time, transition *ClusterStatusTransition) error {
```
ruanxin, line 133: I think we can extract the shared logic of `dropRecordsOlderThan` and `dropSuccessfulRecordsOlderThan` to something like `dropRecordsOlderThanByFilter(t time.Time, transition *ClusterStatusTransition,  filter func(*model.ReconciliationEntity) bool)` to get rid of duplicate code

pkg/scheduler/service/cleaner.go
```diff
@@ -34,14 +45,67 @@ func (c *cleaner) Run(ctx context.Context, transition *ClusterStatusTransition,
...
+	}
+
+	//if we're here, there may exist unsuccessful entities older than 'oldestInRange', but within 'KeepUnsuccessfulEntitiesDays' time range.
+	//We have to preserve these (if exist) and remove everything else.
+	deadline := beginningOfTheDay(now.UTC()).AddDate(0, 0, -1*config.keepUnsuccessfulEntitiesDays())
+	err = c.dropRecordsOlderThan(deadline, transition)
+	if err != nil {
+		c.logger.Errorf("Cleaner failed to remove reconciliations older than %s: %s", deadline.String(), err.Error())
+	}
+	err = c.dropSuccessfulRecordsOlderThan(oldestInRange.Created, transition)
```
ruanxin, line 102: Should we move this `dropSuccessfulRecordsOlderThan` before `if oldestInRangeAgeDays > config.keepUnsuccessfulEntitiesDays() `? I mean this removal doesn't need to depend on that condition, right? We can run this on each cleaning step.

pkg/scheduler/service/cleaner.go
Tomasz-Smelcerz-SAP: I am aware of that, this is just to be sure in case someone in the future breaks the contract and the returned list is not sorted as we expected. I think spending few dozens microseconds on iteration is worth it. After all the oldest record found is then taken as the reference for removal of really important entities (auditing, on-call debugging etc). In cases like this I would say a "double-check" is justified.

pkg/scheduler/service/cleaner.go
Tomasz-Smelcerz-SAP: Nope, it's not :( I don't know, have to ask Tobias.

Tomasz-Smelcerz-SAP: OK, I have the confirmation the condition is correct.

pkg/scheduler/service/cleaner.go
Tomasz-Smelcerz-SAP: Done.

pkg/scheduler/service/cleaner.go
Tomasz-Smelcerz-SAP: TLDR: No, we should not. We could do that, but that would be a waste of time, because we anyway have to execute more broad deletion here: "Delete all records older than [this one]"

Explanation:
One of requirements for this algorithm is: `Keep at most "X" last records, regardless of their state`, where "X" is configurable positive integer.
The other requirement is 'Keep "unsuccessfull" for at least "Y" days back (from now)`, where "Y" is a configurable positive integer. You can write the condition in a more formal way as: `record.status != success && record.age <= Y`

Now consider a situation, where you have the last "X" records, and the oldest of them is from Z days before, where Z > Y.
It means that, with such a set, you satisfy BOTH conditions already: You have last "X" records, and the oldest of them is from Z days before. If there are any "unsuccessful" ones in the set, they all satisfy condition record.age <= Y (because Z > Y)
In that scenario, all you have to do is to drop all remaining records older than the "oldest" in the set, and you're done.
Exactly this scenario is detected in the code by the condition: `if oldestInRangeAgeDays > config.keepUnsuccessfulEntitiesDays()`
It's explained in the commentary.

The other line, `err = c.dropSuccessfulRecordsOlderThan` covers a different scenario.
Consider previous scenario, but now assume Y > Z.
We have three sets now:
1: The set of "X" records we have to keep. The oldest of these is from Z days before. 
2: The set of unsuccessful records where record.age <=Y (Y > Z). This set may partially overlap set number 1, or be distinct, or be empty
3: The set of successful records older than the oldest in set 1.

We have to keep the union of sets 1 and 2: 1 + 2, but we have to remove set 3. This is exactly what `err = c.dropSuccessfulRecordsOlderThan(oldestInRange.Created, transition)` does.

pkg/scheduler/service/cleaner.go
Tomasz-Smelcerz-SAP: Removed

pkg/scheduler/service/cleaner.go
Tomasz-Smelcerz-SAP: Done

tobiscr (approved):

PR #703, opened 2022-01-21
No blind mapping

jeremyharisch:
We have to verify the DB errors instead of blindly mapping them to be a `repository.NotFoundError` -  To avoid that the system report `Operation {...} belongs to a no longer existing cluster...` which is obviously not true in many cases. 

This has been adjusted by using the already existing function `MapError` instead of directly mapping to `NotFoundError`.

Since this error log can only be accessed from the outside using RBAC, we can return the sql error.

### DIFF
--- a/pkg/repository/errors.go
+++ b/pkg/repository/errors.go
@@ -20,10 +20,6 @@ func (r *Repository) NewNotFoundError(err error, entity db.DatabaseEntity,
 func (r *Repository) MapError(err error, entity db.DatabaseEntity, identifier map[string]interface{}) error {
 	if err != sql.ErrNoRows {
 		return err
-	} else if err != sql.ErrTxDone {
-		return err
-	} else if err != sql.ErrConnDone {
-		return err
 	}
 	return &EntityNotFoundError{
 		entity:     entity,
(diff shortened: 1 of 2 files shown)


pkg/repository/errors.go
```diff
@@ -20,6 +20,10 @@ func (r *Repository) NewNotFoundError(err error, entity db.DatabaseEntity,
 func (r *Repository) MapError(err error, entity db.DatabaseEntity, identifier map[string]interface{}) error {
 	if err != sql.ErrNoRows {
 		return err
+	} else if err != sql.ErrTxDone {
```
ruanxin, line 23: Can we rewrite to following?
```
	if err == sql.ErrNoRows || err == sql.ErrTxDone || err == sql.ErrConnDone {
		return &EntityNotFoundError{
			entity:     entity,
			identifier: identifier,
			err:        err,
		}
	}
	return err

jeremyharisch: Resolved in call

ruanxin (approved):

tobiscr:
Like it +1

PR #711, opened 2022-01-25
Remove orphaned NATS pods when upgrading Kyma from 1.X to 2.X

marcobebway:
**Description**

Sometimes nats-operator fails to remove NATS pods during Kyma upgrade from version 1.X to 2.X, which fails the upgrade process.

This pull request introduces a fix in the Eventing component reconciler (as a pre-action step) by removing orphaned NATS pods when upgrading Kyma from version 1.X to 2.X.

> Note: Image bump is done as part of https://github.com/kyma-project/control-plane/pull/1300.

### DIFF
--- a/pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
+++ b/pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"strings"
 
+	v1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
 
 	"go.uber.org/zap"
-	v1 "k8s.io/api/apps/v1"
 
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/chart"
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/instances/eventing/log"
@@ -25,6 +27,7 @@ const (
 	newConfigValue             = "eu.gcr.io/kyma-project"
 	crdPlural                  = "customresourcedefinitions"
 	serviceKind                = "service"
+	podKind                    = "pod"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
```diff
@@ -63,25 +66,36 @@ func (r *removeNatsOperatorStep) Execute(context *service.ActionContext, logger
 	//	return nil
```
pxsalehi, line 66: I'd say not required for your PR, but if you could remove all these commented out code in these two files, it would be nice! :-)

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -2,64 +2,80 @@ package preaction
...
...0-destination-rule.yaml\napiVersion: networking.istio.io/v1alpha3\nkind: DestinationRule\nmetadata:\n  name: eventing-nats\nspec:\n  host: eventing-nats.kyma-system.svc.cluster.local\n  trafficPolicy:\n    tls:\n      mode: DISABLE\n---\n# Source: nats/templates/40-cr.yaml\napiVersion: nats.io/v1alpha2\nkind: NatsCluster\nmetadata:\n  name: eventing-nats\nspec:\n  size: 1\n  version: \"2.1.8\"\n  serverImage: \"eu.gcr.io/kyma-project/external/nats\"\n  pod:\n    annotations:\n      sidecar.istio.io/inject: \"false\"\n    labels:\n      helm.sh/chart: nats-1.0.0\n      app: nats\n      app.kubernetes.io/managed-by: Helm\n      kyma-project.io/dashboard: eventing\n    resources:\n      limits:\n        cpu: 20m\n        memory: 64Mi\n      requests:\n        cpu: 5m\n        memory: 16Mi\n  natsConfig:\n    debug: true\n    trace: true\n  template:\n    spec:\n      affinity:\n        podAntiAffinity:\n          preferredDuringSchedulingIgnoredDuringExecution:\n            - podAffinityTerm:\n                labelSelector:\n                  matchLabels:\n                    nats_cluster: eventing-nats\n                topologyKey: kubernetes.io/hostname\n              weight: 100\n"
```
pxsalehi, line 27: Multiline strings would probably make this way more understandable/editable.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -2,64 +2,80 @@ package preaction
...
-		givenStatefulSet: true,
+		givenStatefulSet:    true,
+		givenNatsPodsLength: 0,
 	},
 	{
-		givenStatefulSet: false,
+		givenStatefulSet:    false,
+		givenNatsPodsLength: 3,
 	},
 }
```
pxsalehi, line 46: Why is `givenNatsPodsLength` > 0 and `givenStatefulSet: true` not a valid test case?

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
```diff
@@ -2,64 +2,80 @@ package preaction
...
 	//kyma1xVersion  = "1.24.8"
 	//kyma2xVersion  = "2.0"
+
+	natsPodNameFormat = "nats-%d"
 )
 
 var testCases = []struct {
-	givenStatefulSet bool
+	givenStatefulSet    bool
```
pxsalehi, line 35: A small comment that `givenStatefulSet` refers to NATS installation via Helm and the NATS pod length is about the cluster size of the NATS operator, would be useful, IMO. (If I have understood this correctly).

pxsalehi (changes requested):

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
marcobebway: These two are not dependant on each other. Setting `givenStatefulSet` to `true` or `false` does not affect how the `givenNatsPodsLength` is used in the test.

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep.go
marcobebway: Done

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
marcobebway: Done

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
marcobebway: Done

pkg/reconciler/instances/eventing/preaction/removenatsoperatorstep_test.go
pxsalehi: Yes, I understand. My question is whether we need a test case that represents a cluster where cleanup of NATS pods (created by the operator) were not successful/finished, and there also exists new NATS pods from the statefulset. Can the cluster be in such a state? Or we are sure that until the cleanup is not successful the statefulset is not create?

marcobebway: This state can exist on already broken clusters, and our pre-action step should be able to fix it. But as I mentioned before, the way the test is written does not make `givenStatefulSet` and `givenNatsPodsLength` depend on each other.

Do you think I should add the test anyway?

pxsalehi: If it is valid corner case, why not test it? If I remember correctly, one of the main motivations for this change to the reconciler was to cover these broken cluster cases.

marcobebway: Done

pxsalehi (approved):
Thanks!

PR #712, opened 2022-01-25
Drop CR finalizers

Tomasz-Smelcerz-SAP:
**Description**

This PR extends logic for dropping Custom Resource finalizers. Such finalizers may, in certain conditions, block cluster deprovisioning.

There are two modes of operation (already added to kyma-cli)
- "system": Drop only finalizers from Custom Resource instances of types (CustomResourceDefinitions) introduced with Kyma 
- "all": Drop finalizers from all Custom Resource instances found in the cluster.

Both modes operate on all namespaces.
For "system" mode, the logic collects CRDs from Kyma sources (i.e. from Kyma charts), then adds CRDs from the cluster labelled with `reconciler.kyma-project.io/managed-by=reconciler`, then adds Istio CRDs. This set of CRDs is then used to find all existing CR instances.

### DIFF
--- a/pkg/reconciler/instances/cleaner/action.go
+++ b/pkg/reconciler/instances/cleaner/action.go
@@ -1,9 +1,20 @@
 package cleaner
 
 import (
+	"fmt"
+
 	"github.com/kyma-incubator/reconciler/pkg/model"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/chart"
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/instances/cleaner/pkg/cleanup"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/kubernetes"
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/service"
+	"go.uber.org/zap"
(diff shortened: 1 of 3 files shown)


pkg/reconciler/instances/cleaner/action.go
```diff
@@ -16,17 +27,117 @@ func (a *CleanupAction) Run(context *service.ActionContext) error {
...
+
+	var kymaCRDsFinder cleanup.KymaCRDsFinder = func() ([]schema.GroupVersionResource, error) {
+		crdManifests, err := context.ChartProvider.RenderCRD(context.Task.Version)
+		if err != nil {
+			return nil, err
+		}
+		return findKymaCRDs(crdManifests, context.Logger)
+	}
+
+	dropFinalizersOnlyForKymaCRs := readDeleteStrategy(context.Task.Configuration) != "all"
```
tobiscr, line 46: Can we maybe move the KEY + VALUE as public constant to it and use in the CLI to set this config-parameter in the task?

@clebs : JFYI

tobiscr (approved):

PR #732, opened 2022-01-30
Treat orphans as new ops

tobiscr:


jeremyharisch:
Tested it locally with local postgres DB setup. I was changing the status of multiple operations to `orphan` several times. The operations have been picked up again and progressed as if they where new. 

LGTM

jeremyharisch (approved):

Issue #638, opened 2022-01-03
[ORY] implement test for ORY evaluation installation


strekm:
**Description**

Implement integration test testing ORY installation using evaluation profile on k3d cluster. Test should assume that connection to a cluster is provided and ORY stack is installed along with kyma pre-reqs. Test should check proper configuration for evaluation profile: if all necessary pods are running, HPA configuration. 

Expected result:

- ory pods should be running ( ory-hydra, oathkeeper, hydra-maester)
- check that there is not hpa in place (only one pod per mentioned one)

**Attachments**

depends on: https://github.com/kyma-incubator/reconciler/issues/637, part of: https://github.com/kyma-project/kyma/issues/12968

dariusztutaj:
Waiting for #750

PR #742, opened 2022-02-01
Test for ORY evaluation installation

dariusztutaj:
### Description
 - implement test for ORY evaluation installation

### Links:
Resolves: #638
Based on: #637

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ test-ory:
 .PHONY: test-ory-eval
 test-ory-eval: export ORY_RECONCILER_INTEGRATION_TESTS_EVALUATION = 1
 test-ory-eval:
-	go test -v -timeout 5m ./pkg/reconciler/instances/ory/test/ory_integration_evaluation_test.go
+	go test -v -timeout 5m ./pkg/reconciler/instances/ory/test/evaluation/ory_integration_evaluation_test.go
 
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ test-ory:
 	go test -v -timeout 5m ./pkg/reconciler/instances/ory/test/ory_integration_test.go
 
 .PHONY: test-ory-eval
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -70,7 +70,12 @@ test-all: test
 .PHONY: test-ory
 test-ory: export ORY_RECONCILER_INTEGRATION_TESTS = 1
 test-ory:
-	go test -v -timeout 5m ./pkg/reconciler/instances/ory/test
+	go test -v -timeout 5m ./pkg/reconciler/instances/ory/test/ory_integration_test.go
+
+.PHONY: test-ory-eval
+test-ory-eval: export ORY_RECONCILER_INTEGRATION_TESTS_EVALUATION = 1
```
tobiscr, line 76: Do we really one more env-var to introduce? I'm not even happy that we have already two of them.

dariusztutaj: You're right. Removed

Makefile
```diff
@@ -70,7 +70,12 @@ test-all: test
 .PHONY: test-ory
 test-ory: export ORY_RECONCILER_INTEGRATION_TESTS = 1
 test-ory:
-	go test -v -timeout 5m ./pkg/reconciler/instances/ory/test
+	go test -v -timeout 5m ./pkg/reconciler/instances/ory/test/ory_integration_test.go
+
+.PHONY: test-ory-eval
+test-ory-eval: export ORY_RECONCILER_INTEGRATION_TESTS = 1
+test-ory-eval:
+	go test -v -timeout 5m ./pkg/reconciler/instances/ory/test/evaluation/ory_integration_evaluation_test.go
```
cnvergence, line 78: We can also make use of EXECUTION_PROFILE env variable in the prow jobs, so we won't need to create separate tests. I will be happy to discuss it tomorrow.

dariusztutaj:
Closing. It will be implemented in different way.

PR #747, opened 2022-02-01
periodic-reconciler-upgrade-pipeline-fix

khlifi411:
Modified occupancy update logic during callback


khlifi411:
Doesn't solve the upgrade pipeline issue

Issue #754, opened 2022-02-01
Grouping of components is not considering pre-components


tobiscr:
**Description**

The log output of the job also indicates that the array of pre-components was not considered when defining the execution sequence:

      {“CRDs@default[1],cleaner@default[2],istio-resources@istio-system[3],tracing@kyma-system[3],monitoring@kyma-system[3],eventing@kyma-system[3],ory@kyma-system[3],service-catalog-addons@kyma-system[3],helm-broker@kyma-system[3],istio-configuration@istio-system[3],application-connector@kyma-integration[3],serverless@kyma-system[3],api-gateway@kyma-system[3],cluster-essentials@kyma-system[3],kiali@kyma-system[3],rafter@kyma-system[3],cluster-users@kyma-system[3],certificates@istio-system[3],service-catalog@kyma-system[3],logging@kyma-system[3]"}

This shows that the preComponents are not properly grouped. Local tests confirmed that the grouping of components is not correct anymore and ignores any pre-components.

<img width="1160" alt="Screen Shot 2022-02-01 at 21 42 56" src="https://user-images.githubusercontent.com/73824711/152048139-4caf515d-a95c-40b4-891e-431c51762ec7.png">

**Expected result**

Pre-components have to be considered when defining the execution sequence.

**Actual result**

Ordering or components it no correct and pre-components seem not to be considered anymore.

**Steps to reproduce**

**Troubleshooting**

tobiscr:
After a sync with @veichtj we also figured out, that the REST call send to the mothership reconciler includes the `istio-resources` component which is part of KYma 2.1 but should not be used in Kyma 2.0.4.

This has also to be adjusted.

tobiscr:
Fix coming in https://github.com/kyma-incubator/reconciler/pull/757

PR #757, opened 2022-02-01
Fix non-considered pre-components and retry failing DB transactions

tobiscr:
#754


ruanxin (approved):

pkg/db/transaction.go
```diff
@@ -5,30 +5,65 @@ import (
...
-		return nil, err
-	}
 
-	result, err := dbOps(txConnection)
-	if err != nil {
-		log("Rollback transactional DB context because an error occurred: %s", err)
-		if rollbackErr := txConnection.Rollback(); rollbackErr != nil {
-			err = errors.Wrap(err, fmt.Sprintf("Rollback of db operations failed: %s", rollbackErr))
+	for i := 0; i < txMaxRetries; i++ {
+		if i > 0 {
```
ruanxin, line 29: Can we add some commit to explain this behavior? is the purpose here to give some delay to prevent concurrent transaction calls (like the previous one is during rollback and prevent new one begin)?

tobiscr: Sure, will add a comment.

What is does is: if the commit failes caused by collsiions, it will rollback the transaction and run it from scratch again until `txMaxRetries` is reached

ruanxin (approved):

pkg/db/transaction.go
```diff
@@ -5,30 +5,66 @@ import (
...
-	if err != nil {
-		log("Rollback transactional DB context because an error occurred: %s", err)
-		if rollbackErr := txConnection.Rollback(); rollbackErr != nil {
-			err = errors.Wrap(err, fmt.Sprintf("Rollback of db operations failed: %s", rollbackErr))
+	//retry the transaction until txMaxRetries is reached. Each retry has a few msec delay for better load balancing
+	for i := 0; i < txMaxRetries; i++ {
+		if i > 0 {
+			rand.Seed(time.Now().UnixNano())
+			//nolint:gosec //no security relevance, linter complains can be ignored
+			jitter := time.Duration(rand.Int63n(txMaxJitter))
```
Tomasz-Smelcerz-SAP, line 33: Isn't that Duration in range [0..350) nanoseconds?
I vote for extracting that to a separate function with an integer argument that returns a Duration.
Also add a simple unit test that verifies if the returned random Duration is within bounds.
The `time.Sleep(jitter)` should stay here and we shouldn't unit-test it's invocation.

pkg/db/transaction.go
```diff
@@ -5,30 +5,66 @@ import (
...
+		commitErr := txConnection.commit()
+		if commitErr == nil {
+			return result, nil
 		}
-		return result, err
+
+		txErr = errors.Wrap(txErr, commitErr.Error())
+		log("Rollback transactional DB context because commit failed: %s", commitErr.Error())
+		if txRollbackErr := txConnection.Rollback(); txRollbackErr != nil {
+			txErr = errors.Wrap(txErr, txRollbackErr.Error())
```
Tomasz-Smelcerz-SAP, line 62: We could log that here, otherwise only the last one will be returned at the end of the function

PR #759, opened 2022-02-02
Fix DB transaction-retry

tobiscr:


Tomasz-Smelcerz-SAP (approved):

pkg/db/transaction.go
```diff
@@ -46,27 +40,44 @@ func TransactionResult(conn Connection, dbOps func(tx *TxConnection) (interface{
...
-			txErr = errors.Wrap(txErr, txRollbackErr.Error())
+			log("Rollback failed: %s", txRollbackErr)
 		}
 
 	}
 
 	return result, txErr
 }
 
+func randomJitter() time.Duration {
```
ruanxin, line 71: is it make sense change to this? the jitter can be just a pure int number, right? If `* time.Millisecond` at the end.
```
	jitter := rand.Intn(txMaxJitter)
	if jitter < txMinJitter {
		jitter = jitter + txMinJitter
	}
	return time.Duration(jitter) * time.Millisecond
```

tobiscr: The jitter should be dynamic to achieve some load balancing on DB side (right now, we can assume that the problem happens because too many transactions are trying to update their data at the same time).

Using a int-number which is not dynamic would lead to a next round of collisison when it retries all transactions at the same time later.

ruanxin: Ok, I got the seed purpose, but then we just need to bring the seed in the beginning to offer a wide range of int numbers, but regarding the `jitter`, it's the same logic, just the time when it converts to `time.Duration`

Issue #755, opened 2022-02-01
Implement retry mechanism for failing DB transactions


tobiscr:
**Description**

Caused by the increased isolation level in PG, it can happen that particualr queries are failing caused by colliding DB transactions. 

Postgres expectes that the client is retrying such queries by its own when using an isolatio level >= repeatable read ([see PG docs](https://www.postgresql.org/docs/9.1/transaction-iso.html)):

_Applications using this level must be prepared to retry transactions due to serialisation failures._

The DB layer of the mothership reconciler has to consider such error and retry such issues automatically.

**Expected result**

DB error indicating a colliding DB transaction (e.g. `pq: could not serialize access due to read/write dependencies among transactions`) have to be catched and a retry has to happen.

**Actual result**

DB retry errors are ignored and lead to errors in the log.

**Steps to reproduce**

**Troubleshooting**

tobiscr:
Fix is part of https://github.com/kyma-incubator/reconciler/pull/757

Issue #753, opened 2022-02-01
Unit test suite fails when running on Postgres


tobiscr:
**Description**

Unittest have to be implemented in a way that they also run on Postgres. Currently are multiple tests failing when using PG as database for the test execution:

- [x] 1. Test-Inventory

          --- FAIL: TestInventory (6.45s)
              --- FAIL: TestInventory/Get_clusters_with_particular_status (1.29s)
                 inventory_test.go:138:
            	     Error Trace:	inventory_test.go:138

         --- FAIL: TestInventory/Get_clusters_to_reconcile (3.74s)
             inventory_test.go:240:
                 	Error Trace:	inventory_test.go:240

2. TestCountRetries

            --- FAIL: TestCountRetries (5.13s)
                testfactory.go:49:
                    	Error Trace:	testfactory.go:49
              	                  				inventory_test.go:339
              	      Error:      	Received unexpected error:
              	                  	pq: cannot truncate a table referenced in a foreign key constraint
                    	Test:       	TestCountRetries

- [x] 3. TestBookkeepingTask

            --- FAIL: TestBookkeepingtask (0.40s)
                --- FAIL: TestBookkeepingtask/Mark_operations_as_orphan (0.20s)
                    bookkeepingtask_test.go:56:
                        	Error Trace:	bookkeepingtask_test.go:56
                        	Error:      	Received unexpected error:
                        	            	cluster 'runtime1' is already considered for reconciliation (schedulingID:runtime1--1c3a42a7-7ff6-40c2-95e1-b7cc1e6a9270)
                        	Test:       	TestBookkeepingtask/Mark_operations_as_orphan
                --- FAIL: TestBookkeepingtask/Finish_operations (0.20s)
                    bookkeepingtask_test.go:56:
                              	Error Trace:	bookkeepingtask_test.go:56
                              	Error:      	Received unexpected error:
                  	            	cluster 'runtime1' is already considered for reconciliation (schedulingID:runtime1--1c3a42a7-7ff6-40c2-95e1-b7cc1e6a9270)
            	Test:       	TestBookkeepingtask/Finish_operations

4. TestBookkeepingTaskParallel

            --- FAIL: TestBookkeepingtaskParallel (3.18s)
                      --- FAIL: TestBookkeepingtaskParallel/Mark_two_operations_as_orphan_in_multiple_parallel_threads (2.67s)
                    bookkeepingtask_test.go:209:
                        	Error Trace:	bookkeepingtask_test.go:209
[...]

PR #760, opened 2022-02-02
Fix DB transaction retry issue

tobiscr:
See: https://github.com/kyma-incubator/reconciler/issues/755 and https://github.com/kyma-incubator/reconciler/issues/753


pkg/cluster/inventory.go
```diff
@@ -20,6 +20,7 @@ type Inventory interface {
 	Delete(runtimeID string) error
 	Get(runtimeID string, configVersion int64) (*State, error)
 	GetLatest(runtimeID string) (*State, error)
+	GetAll() ([]*State, error)
```
tobiscr, line 23: Required to stabilise inventory unit-tests when using PG

pkg/db/interface.go
```diff
@@ -44,3 +48,15 @@ type DataRows interface {
 	Scan(dest ...interface{}) error
 	Next() bool
 }
+
+func newID() string {
```
jeremyharisch, line 52: Why not using uuid package instead? We are already using it in other parts of the project

pkg/db/query.go
```diff
@@ -283,6 +285,24 @@ func (s *Select) GetMany() ([]DatabaseEntity, error) {
 	return result, nil
 }
 
+func (s *Select) sql() string {
```
jeremyharisch, line 288: It would be nice to have a unit test for this function 👍🏻

tobiscr: done

pkg/db/interface.go
tobiscr: Changed to UUID

jeremyharisch (approved):

clebs (approved):

Issue #744, opened 2022-02-01
Istio images version mismatch after names change


mjakobczyk:
**Description**

Istio reconciliation fails after images' offered by SoloIO team changed their names.

**Expected result**

No version mismatch, as this one is expected. Implemented changes need to support all versions of managed kyma (e.g. 2, main)

**Actual result**

```bash
2022-01-31T22:50:01.838Z	DEBUG	service/runner.go:132	Runner: Reconcile action of 'istio-configuration' with version 'local' failed: Istio components version mismatch detected: pilot version: 1.12.2, data plane version: 1.12.2-solo-fips
```

**Steps to reproduce**

Changes introduced by [this PR](https://github.com/kyma-project/kyma/pull/13229/files) popped this up.

**Troubleshooting**

Have a look at the following Reconciler code:

`pkg/reconciler/instances/istio/action.go:369`
```go
func isMismatchPresent(istioStatus actions.IstioStatus) bool {
	pilot := newHelperVersionFrom(istioStatus.PilotVersion)
	dataPlane := newHelperVersionFrom(istioStatus.DataPlaneVersion)
	return pilot.compare(&dataPlane) != 0
}
```

Then this mismatch is being used here:

`pkg/reconciler/instances/istio/action.go:105`
```go
	if isMismatchPresent(istioStatus) {
		errorMessage := fmt.Sprintf("Istio components version mismatch detected: pilot version: %s, data plane version: %s", istioStatus.PilotVersion, istioStatus.DataPlaneVersion)
		return errors.New(errorMessage)
	}
```

Consider logging a warning it instead of throwing an error.

**Links**
Found during: https://github.com/kyma-project/kyma/pull/13229

strekm:
Bug fix needs mandatory tests agains kyma 2.0 and main

mjakobczyk:
Tested the solution again against both Kyma `main` and `release-2.0` branches. I used istioctl in version 1.11.4 for testing purposes.

Results for branch `main`:
- command:
```bash
./bin/mothership-darwin local --value global.ingress.domainName=example.com,global.domainName=example.com --components istio-configuration --version main
```
- output:
```log
{"level":"INFO","time":"2022-02-17T12:41:28.573+0100","message":"Component 'istio-configuration' has status 'running'"}
{"level":"INFO","time":"2022-02-17T12:41:29.335+0100","message":"Resolved target Istio version: 1.11.4-solo-fips-distroless from values"}
```

[...]

PR #764, opened 2022-02-02
Load Istio target version from values.yaml instead of Chart.yaml

mjakobczyk:
### Description

Suggested changes:
- remove loading Istio target version from Chart.yaml
- load Istio target version from values.yaml
- fix tests

### Links


cnvergence:
Install to 1.11.2 and upgrade to 1.11.4 works properly.
LGTM after lint fixes

cnvergence (approved):

PR #771, opened 2022-02-03
Resolve issue when using Postgres for db unit test

ruanxin:
See: https://github.com/kyma-incubator/reconciler/issues/753

### DIFF
--- a/cmd/mothership/mothership/start/integration_test.go
+++ b/cmd/mothership/mothership/start/integration_test.go
@@ -5,6 +5,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/kyma-incubator/reconciler/pkg/cluster"
+	"github.com/kyma-incubator/reconciler/pkg/db"
+	"github.com/kyma-incubator/reconciler/pkg/scheduler/reconciliation"
 	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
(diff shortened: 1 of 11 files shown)


cmd/mothership/mothership/start/integration_test.go
```diff
@@ -470,16 +470,6 @@ func TestMothership(t *testing.T) {
 			responseModel:    &keb.HTTPClusterResponse{},
 			verifier:         requireClusterStateChangeFct("ready"),
 		},
-		{
```
ruanxin, line 473: To Reviewer:

The reason why I delete this test case is because this delete request only mark cluster to be deleted, which adds a record in `inventory_cluster_config_statuses` to indicate the runtimeid of this cluster be `delete_pending`, but the record in `inventory_clusters` will not change, and this test we don't have component reconciler to do actual deleting, so eventually the data will not reflect this cluster is deleted.

That's the reason why I add actual cleanup code in the above defer function.

pkg/keb/test/cluster.go
```diff
@@ -10,23 +10,39 @@ import (
 )
 
 func NewCluster(t *testing.T, runtimeID string, clusterVersion uint64, newConfigVersion bool, clusterType Cluster) *keb.Cluster {
+	runtimeID += uuid.NewString()
```
jeremyharisch, line 13: I would ust use the uuid and remove the runtimeId as input

ruanxin: That will involve many code refactor, and to allow manually adding something to runtimeID can help during debug (search that prefix in runtimeID)

ruanxin: But maybe better change the `runtimeID` to `runtimeIDPrefix` so that not leave the wrong impression that this ID is same as given input. What do you think?

jeremyharisch (approved):
As mentioned in the code comment, I would remove the

Issue #329, opened 2021-11-01
Load and stress test for reconciler


tobiscr:
**Description**

The performance profile of the reconciler has to be analysed to identify performance bottlenecks and be aware about the thresholds / limits we should not exceeds.

AC:

* Implement a load test scenario for the reconciler (e.g. increase the number of parallel reconciliations until X reconciliations are running in parallel)
* measure the resource consumption of the mothership and component reconcilers and track these figures in a central dashboard
* visualise the recorded consumption data in the dashboard
* setup CI / CD pipeline for executing te load test

- [x] Setup a load test scenario
     - [x] Create a test script which creates X cluster entites the mothership
     - [x] implement a simple mock-component reconciler which is capable to retrieve requests from mothership and sends regularly heartbeat messages back
- [x] #774 
- [x] https://github.com/kyma-incubator/reconciler/issues/818

**Reasons**

Get awareness of the reconciler scalability, bottlenecks and performance thresholds / limits.

**Attachments**

dariusztutaj:
https://github.tools.sap/kyma/security-scans/issues/11301

adityabhatia:
Load and stress test focus moved to the kyma-operator at the moment

janmedrek:
Functionalities of the Reconciler were migrated to the new architecture, this issue will not be worked on.

Issue #549, opened 2021-12-09
Refactoring: remove `utils` package from `reconciler/instances` pkg


tobiscr:
**Description**

There was a `utils` package added to the `pkg/reconciler/instances` package which includes some K8s-specific helper functions.

As `utils` packages are in general a bit weird, I would challenge whether the code shouldn't be moved into an existing pkg.

Currently, this `utils` pkg causes also the side-effect, that the `reconcilerctl.sh` script misbehaves and includes this `utils` pkg into the `loader.go`which is unnecessary.

**Expected result**

* Utils pkg should not be a sub-package of the `instances` pkg as its code looks more like helper-functions which can become part of an existing pkg (e.g. `reconciler/kubernetes`) or moved to a component-reconciler specific code (if it's only used by one comp-reconciler).

**Actual result**

The `reconciler/instances/utils` pkg causes wrongly rendered `loader.go` file when using `reconcilerctl.sh` script.

**Steps to reproduce**

**Troubleshooting**

janmedrek:
Functionalities of the Reconciler were migrated to the new architecture, this issue will not be worked on.

PR #776, opened 2022-02-04
Introduce reconciler mock component

mjakobczyk:
### Description

Changes proposed:
- introduce `mock` component reconciler

### Links
See: https://github.com/kyma-incubator/reconciler/issues/329

### DIFF
--- a/pkg/reconciler/instances/loader.go
+++ b/pkg/reconciler/instances/loader.go
@@ -17,6 +17,8 @@ import (
 	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/eventing"
 	//import required to register component reconciler 'istio' in reconciler registry
 	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/istio"
+	//import required to register component reconciler 'mock' in reconciler registry
+	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/mock"
 	//import required to register component reconciler 'ory' in reconciler registry
 	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/ory"
 	//import required to register component reconciler 'rafter' in reconciler registry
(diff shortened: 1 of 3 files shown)


pkg/reconciler/instances/loader.go
```diff
@@ -17,6 +17,8 @@ import (
 	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/eventing"
 	//import required to register component reconciler 'istio' in reconciler registry
 	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/istio"
+	//import required to register component reconciler 'mock' in reconciler registry
+	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/mock"
```
tobiscr, line 21: This will also load the mock in KCP as dedicated service. If you need a mock in your unit-test or simliar, please [have a look here](https://github.com/kyma-incubator/reconciler/blob/main/cmd/reconciler/start/service/integration_test.go#L99-L105): this create a component reconciler dynamically during the test execution.

pkg/reconciler/instances/loader.go
```diff
@@ -27,4 +29,6 @@ import (
 	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/scmigration"
 	//import required to register component reconciler 'serverless' in reconciler registry
 	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/serverless"
+	//import required to register component reconciler 'utils' in reconciler registry
+	_ "github.com/kyma-incubator/reconciler/pkg/reconciler/instances/utils"
```
tobiscr, line 33: This package is wrongly located and has to be remove urgently (see #549 )

tobiscr (dismissed):

pkg/reconciler/instances/loader.go
dariusztutaj: I assume that if we moved pr to protected branch we can proceed with this approach

pkg/reconciler/instances/loader.go
dariusztutaj: it's created by mechanism to create new component reconcilers. It could be removed when mentioned pr will be merged

jeremyharisch (approved):

PR #778, opened 2022-02-04
Revert "Load Istio target version from values.yaml instead of Chart.yaml"

mjakobczyk:
### Description

Revert changes made to Istio image fetching due to inconsistency that turned out after merge.

### Links


strekm (approved):
/meow

Issue #705, opened 2022-01-21
Well defined actions for Istio reconciler


strekm:
**Description**

Istio reconciler should have well defined actions which represents flow of Istio installation / upgrade.

**AC**
- [x] sanity checks which are pre action responsible for checking version mismatch on the cluster and client compatibility
- [x] Istio reconcile command which is main action responsible for installation / upgrade of Istio
- [x] mutating webhook configuration patch command responsible for applying proper sidecar injection configuration
- [x] proxy reset command which is post action responsible for reseting sidecars

**Reasons**

Having single responsibility components and getting rid of technical debt

**Attachments**

https://github.com/kyma-incubator/reconciler/issues/543

mjakobczyk:
During the conversation with @werdes72 we did the following:
- checked the actions flow of the Reconciler runner - when any action fails then full chain is repeated from scratch (not only failed action)
- as a result we decided that sanity checks for Istio can be done as a separate action not to be repeated every time
- still Istio status should be fetched in every reconciler action (not once for the chain)
- main reconcile action should be about handling Istio itself so it should manage installation and upgrade
- there can be two post actions - the first one for webhook and second one for proxy reset
- Acs were updated appropriately

PR #789, opened 2022-02-07
Well defined actions for Istio reconciler

werdes72:
See #705


mjakobczyk (approved):

pkg/reconciler/instances/istio/action.go
```diff
@@ -21,27 +20,16 @@ const (
...
-	getIstioPerformer bootstrapIstioPerformer
-}
-
-// NewReconcileAction returns an instance of ReconcileAction
-func NewReconcileAction(getIstioPerformer bootstrapIstioPerformer) *ReconcileAction {
-	return &ReconcileAction{getIstioPerformer}
-}
-
-type UninstallAction struct {
+type PreReconcileAction struct {
```
mjakobczyk, line 23: You renamed post actions in pretty cool format, e.x. `MutatingWebhookPostAction`. What about sticking to this convention here? How about naming this action: `SanityChecksPreAction` or `IstioChecksPreAction` or somewhat similar?

pkg/reconciler/instances/istio/action.go
```diff
@@ -52,31 +40,27 @@ func (a *UninstallAction) Run(context *service.ActionContext) error {
...
 	}
+
+	if !isClientCompatibleWithTargetVersion(istioStatus) {
+		return errors.New(fmt.Sprintf("Istio could not be updated since the binary version: %s is not compatible with the target version: %s - the difference between versions exceeds one minor version", istioStatus.ClientVersion, istioStatus.TargetVersion))
+	}
+
 	return nil
 }
 
+type ReconcileAction struct {
```
mjakobczyk, line 56: As above. What about naming it: `IstioHandlingMainAction` or `IstioInstallUpgradeMainAction` or something similar?

pkg/reconciler/instances/istio/action.go
```diff
@@ -3,7 +3,6 @@ package istio
 import (
```
mjakobczyk, line 3: In general I don't like the fact that all these actions are mixed in the same file. The tests are huge then because they are all defined in the same testing file too.

mjakobczyk (dismissed):
Looks better! In the future we can think of splitting these actions into multiple files. 👍

pkg/reconciler/instances/istio/action.go
mjakobczyk: Yup, so this way we would simply omit the first word

mjakobczyk (approved):

Issue #310, opened 2021-10-25
OpsTest Finding: Add support for autoscaling to reconcilers


tobiscr:
**Description**

Depending on the saturation of the resources, another pod should be started for mothership or component reconciler.

For all reconcilers is the saturation of the worker pool relevant + CPU/RAM allocation. For the mothership has the max-scaled pods limit to be defined by considering the max. amount of open connections defined on Postgres side (to avoid more open connections than the DB is able to handle).

AC:

* Trigger for creating another replica is based on resource saturation of the pod and on the number of available workers in the worker pool (have a loon on https://keda.sh/)
* Downscaling has to happen after a pod was >5min below the defined saturation levels
* Max. scaled pods of mothership reconciler has to be defined based on the max. open connection limit of the Postgres DB. Max. number of replicas can be be calculated using the max-DB connection limit and the DB-connection pool size per pod: `max replicas = floor(max connection limit of DB / connection-pool size per pod)`

**Reasons**

Scale application depending on given load.

**Attachments**

tobiscr:
For testing this feature the implementation of #329 is required.

janmedrek:
Functionalities of the Reconciler were migrated to the new architecture, this issue will not be worked on.

PR #791, opened 2022-02-08
Workerpool metrics

khlifi411:
#310

### DIFF
--- a/cmd/mothership/main.go
+++ b/cmd/mothership/main.go
@@ -65,6 +65,7 @@ func newCmd(o *cli.Options, name, shortDesc, longDesc string) *cobra.Command {
 	cmd.PersistentFlags().BoolVarP(&o.Verbose, "verbose", "v", false, "Show detailed information about the executed command actions")
 	cmd.PersistentFlags().BoolVar(&o.NonInteractive, "non-interactive", false, "Enables the non-interactive shell mode")
 	cmd.PersistentFlags().BoolVarP(&o.InitRegistry, "init-registry", "r", false, "Auto-initialize application registry ")
+	cmd.PersistentFlags().BoolVarP(&o.OccupancyTracking, "occupancy-tracking", "ot", false, "Activate worker pool occupancy tracking")
 	cmd.PersistentFlags().BoolP("help", "h", false, "Command help")
 	return cmd
 }
--- a/cmd/mothership/main.go
+++ b/cmd/mothership/main.go
@@ -65,7 +65,7 @@ func newCmd(o *cli.Options, name, shortDesc, longDesc string) *cobra.Command {
 	cmd.PersistentFlags().BoolVarP(&o.Verbose, "verbose", "v", false, "Show detailed information about the executed command actions")
 	cmd.PersistentFlags().BoolVar(&o.NonInteractive, "non-interactive", false, "Enables the non-interactive shell mode")
 	cmd.PersistentFlags().BoolVarP(&o.InitRegistry, "init-registry", "r", false, "Auto-initialize application registry ")
-	cmd.PersistentFlags().BoolVarP(&o.OccupancyTracking, "occupancy-tracking", "ot", false, "Activate worker pool occupancy tracking")
+	cmd.PersistentFlags().BoolVarP(&o.OccupancyTracking, "occupancy-tracking", "t", false, "Activate worker pool occupancy tracking")
 	cmd.PersistentFlags().BoolP("help", "h", false, "Command help")
 	return cmd
 }
--- a/cmd/mothership/mothership/start/http.go
+++ b/cmd/mothership/mothership/start/http.go
@@ -42,6 +42,7 @@ const (
 	paramAfter      = "after"
 	paramLast       = "last"
 	paramTimeFormat = time.RFC3339
+	paramPoolID     = "poolID"
 )
 
 func startWebserver(ctx context.Context, o *Options) error {
@@ -110,8 +111,12 @@ func startWebserver(ctx context.Context, o *Options) error {
 		fmt.Sprintf("/v{%s}/clusters/{%s}/config/{%s}", paramContractVersion, paramRuntimeID, paramConfigVersion),
 		callHandler(o, getKymaConfig)).Methods(http.MethodGet)
 
+	apiRouter.HandleFunc(
+		fmt.Sprintf("/v{%s}/occupancy/{%s}", paramContractVersion, paramPoolID),
+		callHandler(o, deleteComponentWorkerPoolOccupancy)).Methods(http.MethodDelete)
+
 	//metrics endpoint
-	metrics.RegisterAll(o.Registry.Inventory(), o.Logger())
+	metrics.RegisterAll(o.Registry.Inventory(), o.Registry.OccupancyRepository(), o.ReconcilerList, o.Logger())
 	metricsRouter.Handle("", promhttp.Handler())
 
 	//liveness and readiness checks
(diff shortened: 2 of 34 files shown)


jeremyharisch (approved):

cmd/mothership/mothership/start/http.go
```diff
@@ -110,8 +113,16 @@ func startWebserver(ctx context.Context, o *Options) error {
 		fmt.Sprintf("/v{%s}/clusters/{%s}/config/{%s}", paramContractVersion, paramRuntimeID, paramConfigVersion),
 		callHandler(o, getKymaConfig)).Methods(http.MethodGet)
 
+	apiRouter.HandleFunc(
```
tobiscr, line 116: I thought we agreed to use a interval based occupancy updating on the mothership side - why do we need these endpoints?

cmd/mothership/main.go
```diff
@@ -65,6 +65,7 @@ func newCmd(o *cli.Options, name, shortDesc, longDesc string) *cobra.Command {
 	cmd.PersistentFlags().BoolVarP(&o.Verbose, "verbose", "v", false, "Show detailed information about the executed command actions")
 	cmd.PersistentFlags().BoolVar(&o.NonInteractive, "non-interactive", false, "Enables the non-interactive shell mode")
 	cmd.PersistentFlags().BoolVarP(&o.InitRegistry, "init-registry", "r", false, "Auto-initialize application registry ")
+	cmd.PersistentFlags().BoolVarP(&o.OccupancyTracking, "occupancy-tracking", "t", false, "Activate worker pool occupancy tracking")
```
tobiscr, line 68: We have to agree how we track feature flags.. I would not put them as args to the CLI otherwise we will end up over time with too many CLI arguments.

pkg/scheduler/worker/workerpool.go
```diff
@@ -97,6 +106,11 @@ func (w *Pool) invokeProcessableOpsOnce(ctx context.Context, workerPool *ants.Po
 
 func (w *Pool) startWorkerPool(ctx context.Context) (*ants.PoolWithFunc, error) {
 	w.logger.Infof("Starting worker pool with capacity of %d workers", w.config.PoolSize)
+	w.poolID = uuid.NewString()
```
tobiscr, line 109: The worker pool has to be coupled to the pod name/id. Each mothership instance has one worker pool and we have to be able to map each instance to the trakced worker pool in our DB otherwise we aren't able to detect orphan worker pool of a pod got killed and had no proper shutdown.

pkg/scheduler/worker/workerpool.go
```diff
@@ -173,6 +187,10 @@ func (w *Pool) invokeProcessableOps(workerPool *ants.PoolWithFunc) (int, error)
 		idx++
 	}
 	w.logger.Infof("Worker pool assigned %d of %d processable operations to workers", idx, opsCnt)
+	err = w.occupancyRepo.UpdateWorkerPoolOccupancy(w.poolID, workerPool.Running())
```
tobiscr, line 190: I would highly recommend to move the whole tracking of the worker pool occupany into a dedicated component and not to add it on top of the worker pools. The occupancy tracking could happen in intervals and also the synchronisation of worker pools with running mothership-pods has to happen in intervals.

cmd/mothership/mothership/start/scheduler.go
```diff
@@ -56,6 +58,15 @@ func startScheduler(ctx context.Context, o *Options, schedulerCfg *config.Config
 		Run(ctx)
 }
 
+func getReconcilers(cfg *config.Config) []string {
+	reconcilerList := make([]string, 0, len(cfg.Scheduler.Reconcilers)+1)
+	for reconciler := range cfg.Scheduler.Reconcilers {
+		formattedReconciler := strings.Replace(reconciler, "-", "_", -1)
```
tobiscr, line 64: Don't think this is the right place for such logic: if the metrics canno deal with the names as they are, the conversion of these names should happen inside of the metrics struct.

cmd/mothership/mothership/start/test/requests/new_occupancy.json
```diff
@@ -0,0 +1,5 @@
+{
+  "component": "component",
```
tobiscr, line 2: Naming of the field is a bit confusing.. its' actually the name of the component-reconciler and not of the component (e.g. the component reconciler `base` is used for multiple components).

pkg/reconciler/service/workerpool.go
```diff
@@ -16,10 +23,12 @@ type workPoolBuilder struct {
...
 type WorkerPool struct {
-	debug        bool
-	logger       *zap.SugaredLogger
-	antsPool     *ants.Pool
-	newRunnerFct func(context.Context, *reconciler.Task, callback.Handler, *zap.SugaredLogger) func() error
+	debug                bool
+	logger               *zap.SugaredLogger
+	antsPool             *ants.Pool
+	PoolID               string
+	occupancyCallbackURL string
```
tobiscr, line 30: We should think of using a push instead of a pull model: the occupancy-collector module in the mothership reconciler is requesting the occupancy per reconciler.

This scales much better instead letting each reconciler report it to the mothership.

pkg/reconciler/service/workerpool.go
```diff
@@ -53,23 +74,86 @@ func (pb *workPoolBuilder) Build(ctx context.Context) (*WorkerPool, error) {
 		return nil, err
 	}
 	pb.workerPool.antsPool = antsPool
+	pb.workerPool.PoolID = uuid.NewString()
 
 	go func(ctx context.Context, antsPool *ants.Pool) {
-		<-ctx.Done()
-		log.Info("Shutting down worker pool")
-		antsPool.Release()
+		ticker := time.NewTicker(defaultInterval)
```
tobiscr, line 80: This logic should be moved to the mothership into the occupancy collector.

tobiscr, line 80: From architectural side, we should think to move this logic into a separte module of the component-reconcilers as the measuring of the metrics (currently, we measure only the worker pool, but I assume a few more values will be added) should not be connected ot the worker-pool module itself.

pkg/metrics/metrics.go
```diff
@@ -2,12 +2,18 @@ package metrics
...
 import (
 	"github.com/kyma-incubator/reconciler/pkg/cluster"
+	"github.com/kyma-incubator/reconciler/pkg/scheduler/occupancy"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 )
 
-func RegisterAll(inventory cluster.Inventory, logger *zap.SugaredLogger) {
+func RegisterAll(inventory cluster.Inventory, occupancyRepo occupancy.Repository, reconcilerList []string, logger *zap.SugaredLogger, occupancyTracking bool) {
 	reconciliationWaitingCollector := NewReconciliationWaitingCollector(inventory, logger)
```
tobiscr, line 11: Probably better to split that into decicated methods.. one for inventory and another one for the occupancy.

cmd/mothership/mothership/start/options.go
```diff
@@ -29,27 +29,29 @@ type Options struct {
...
+		0 * time.Minute,        //CleanerInterval
+		0,                      //KeepLatestEntitiesCount
+		0,                      //KeepUnsuccessfulEntitiesDays
+		false,                  //CreateEncyptionKey
+		0,                      //MaxParallelOperations
+		false,                  //AuditLog
+		"",                     //AuditLogFile
+		"",                     //AuditLogTenant
+		false,                  //StopAfterMigration
+		[]string{"mothership"}, //ReconcilerList
```
tobiscr, line 54: Better to use an empty array to avoid confusion

cmd/mothership/mothership/start/options.go
```diff
@@ -29,27 +29,29 @@ type Options struct {
 	AuditLogFile                 string
 	AuditLogTenantID             string
 	StopAfterMigration           bool
+	ReconcilerList               []string
```
tobiscr, line 32: Maybe better to rename it to `serviceList` or something as it indicates which microservice or which deployment we want to scale

internal/cli/options.go
```diff
@@ -13,13 +13,14 @@ import (
...
-	logger         *zap.SugaredLogger
-	Registry       *persistency.Registry //will be initialized during CLI bootstrap in main.go
+	Migrate           bool
+	Verbose           bool
+	InitRegistry      bool
+	NonInteractive    bool
+	OutputFormat      string
+	logger            *zap.SugaredLogger
+	Registry          *persistency.Registry //will be initialized during CLI bootstrap in main.go
+	OccupancyTracking bool
```
tobiscr, line 23: As mentioned above, we have to align with the team how we deal in the long-term with feature-flags.

Issue #774, opened 2022-02-04
Expose processing metrics over Prometheus


tobiscr:
**Description**

To give more transparency about the mothership and component reconciler performance (required for load testing), we have to expose more details about invested processing times via Prometheus metrics.

Metrics are exposed by the mothership reconciler and have to be collected / aggregated there.

Following metrics are required:

* Processing time of a particular component on mothership side:
     * Time spend from creating until the operation was finished
     * Time spend from communicating the operation to the component reconciler until it was reported to be finished by the component reconciler
* Processing time of a particular component on component reconciler side:
     * Time spend by the component reconciler for processing the particular component (time spend for rendering + sending it to K8s until it was finished)

All metrics have to be stored in the Database of mothership reconciler per operation. These data have to be reported via the existing Prometheus endpoint.

**AC:**
Prometheus grapes are extend to report per component 
* the avg. lifetime time of a component-operation in the mothership reconciler (created and successfully finished)
* the avg. lifetime time of a component-operation in the mothership reconciler (created and non-successfully finished)
* the avg. processing time by the mothership reconciler (picked up and finished by worker in mothership-reconciler if operation was successful)
* the avg. processing time by the mothership reconciler (picked up and finished by worker in mothership-reconciler if operation was non-successful)
* the avg. processing time by the component reconciler (rendered and finished to be deployed in K8s successfully)
* the avg. processing time by the component reconciler (rendered and finished to be deployed in K8s non-successfully)
* All above mentioned metrics data are stored in DB per operation

**Reasons**

Enable proper analysis during load testing and help to improve the general performance of reconcilers and Kyma components.

**Attachments**

PR #808, opened 2022-02-11
Expose processing duration metrics over Prometheus

jeremyharisch:
This PR introduces new metrics which will be exposed to Prometheus.

The metrics are regarding the duration of the operations, more information in the linked Issue ([link](https://github.com/kyma-incubator/reconciler/issues/774))

I already tested it on a local setup and it worked as expected, for more questions please reach out to me.

Screenshot 2022-02-14 at 14 53 38


pkg/reconciler/service/reconciler.go
```diff
@@ -261,6 +261,6 @@ func (r *ComponentReconciler) newRunnerFunc(ctx context.Context, model *reconcil
 	return func() error {
 		timeoutCtx, cancel := context.WithTimeout(ctx, r.timeout)
 		defer cancel()
-		return (&runner{r, NewInstall(logger), logger}).Run(timeoutCtx, model, callback)
+		return (&runner{r, NewInstall(logger), logger}).Run(timeoutCtx, model, callback) //TODO: pass schedulingID and correlationID
```
khlifi411, line 264: I think you don't need this any more? Right?

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -266,7 +266,7 @@ func (r *PersistentReconciliationRepository) GetOperations(filter operation.Filt
 			return nil, errors.Wrap(err, "failed to apply sql filter")
 		}
 	}
-
+	fmt.Println(selectQ.String())
```
khlifi411, line 269: We do not need this any more? Right?

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -447,3 +447,135 @@ func (r *PersistentReconciliationRepository) UpdateOperationRetryID(schedulingID
 	}
 	return db.Transaction(r.Conn, dbOps, r.Logger)
 }
+
+func (r *PersistentReconciliationRepository) UpdateOperationPickedUp(schedulingID, correlationID string) error {
+	dbOps := func(tx *db.TxConnection) error {
+		op, err := r.GetOperation(schedulingID, correlationID)
```
khlifi411, line 453: Here, you should use the repository with the `tx` parameter like this:
```
rTx := r.WithTx(tx) 
rTx.GetOperation(...)
```

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -447,3 +447,135 @@ func (r *PersistentReconciliationRepository) UpdateOperationRetryID(schedulingID
...
+		if err != nil {
+			if repository.IsNotFoundError(err) {
+				r.Logger.Warnf("ReconRepo could not find operation (schedulingID:%s/correlationID:%s)", schedulingID, correlationID)
+			}
+			return err
+		}
+
+		if !op.PickedUp.IsZero() {
+			r.Logger.Warnf("Operation with schedulingID %s has already picked up at %s", schedulingID, op.PickedUp)
```
khlifi411, line 462: A clearer message would be`Operation with schedulingID %s (was or has been) already picked up at %s`
Also, shouldn't we return an error when the `pickedUp` time is different from zero, because that means it was already updated, right?

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -447,3 +447,135 @@ func (r *PersistentReconciliationRepository) UpdateOperationRetryID(schedulingID
...
+		}
+
+		//update operation-entity
+		op.PickedUp = time.Now().UTC()
+
+		//prepare update query
+		q, err := db.NewQuery(r.Conn, op, r.Logger)
```
khlifi411, line 469: Here, you should use `tx` instead of `r.Conn`

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -447,3 +447,135 @@ func (r *PersistentReconciliationRepository) UpdateOperationRetryID(schedulingID
...
+		if err != nil {
+			return err
+		}
+		if len(operations) != 1 {
+			return fmt.Errorf("error finding operation with schedulingID %s and correlationID %s, found instead: %v", schedulingID, correlationID, operations)
+		}
+		operations[0].ProcessingDuration = processingDuration
+
+		//prepare update query
+		q, err := db.NewQuery(r.Conn, operations[0], r.Logger)
```
khlifi411, line 506: Same as above, we should use `tx` instead of `r.Conn`.

pkg/scheduler/reconciliation/inmemory.go
```diff
@@ -270,3 +270,112 @@ func (r *InMemoryReconciliationRepository) UpdateOperationRetryID(schedulingID,
...
+			&operation.WithStates{
+				States: []model.OperationState{state},
+			},
+			&operation.Limit{Count: metricsQueryLimit},
+		},
+	})
+	if err != nil {
+		return 0, err
+	}
+	if len(operations) == 0 {
```
khlifi411, line 335: I think we need to lock `mu` before accessing `operations` and defer its unlock here.

pkg/scheduler/reconciliation/inmemory.go
```diff
@@ -270,3 +270,112 @@ func (r *InMemoryReconciliationRepository) UpdateOperationRetryID(schedulingID,
...
```
khlifi411, line 365: Same as above. I think we need to lock `mu` before accessing `operations` and defer its unlock here too.

khlifi411 (commented):
Great work Jeremy! I have left some tiny comments.

pkg/scheduler/reconciliation/reconciliation.go
```diff
@@ -26,6 +34,10 @@ type Repository interface {
 	UpdateOperationState(schedulingID, correlationID string, state model.OperationState, allowInState bool, reasons ...string) error
 	WithTx(tx *db.TxConnection) (Repository, error)
 	UpdateOperationRetryID(schedulingID, correlationID, retryID string) error
+	UpdateOperationPickedUp(schedulingID, correlationID string) error
+	UpdateComponentOperationProcessingDuration(schedulingID, correlationID string, processingDuration int64) error
```
khlifi411, line 38: Here, I think it would be better to have `processingDuration` of type `int` instead of `int64` as data conversion responsibility falls on the repository. Currently, it is done on the Http Handler level.

pkg/scheduler/reconciliation/persistent.go
jeremyharisch: Done! Also I need to use the `tx` instead of `r.Conn` 👍🏻

pkg/scheduler/reconciliation/persistent.go
jeremyharisch: It is not an error since this could happen, when the operation was an orphan and is picked up again. but indeed there is a small bug, since to update the timestamp should only happen when an operation gets set to 'inProgress', and if it has already been picmed up before it should not warn the user but it should not update it again. Will fix it

pkg/scheduler/reconciliation/inmemory.go
jeremyharisch: I had implemented in such way before, but this ends in a deadlock. Since, the `GetOperations` already locks, and this function uses the GetOperations function, thus it will already be locked

pkg/scheduler/reconciliation/inmemory.go
jeremyharisch: Resolved in meeting

khlifi411 (approved):

Issue #798, opened 2022-02-09
[Istio] implement test for Istio reconciler


strekm:
**Description**

Implement integration test testing Istio installation. Test should assume that connection to a cluster is provided and Istio is installed along with cluster-essentials Test should check if istio is installed and running

Expected result:

- Istio pods should be running ( istiod, ingressgateway)
- healthz returns 200

**Attachments**

https://github.com/kyma-incubator/reconciler/issues/785

strekm:
@piotrkpc i do not see a reason to test anything in `kyma-system`, please correct me if i'm wrong. if this testing is not needed please remove it to keep this test clean

piotrkpc:
> @piotrkpc i do not see a reason to test anything in `kyma-system`, please correct me if i'm wrong. if this testing is not needed please remove it to keep this test clean

Solved with https://github.com/kyma-incubator/reconciler/pull/841

PR #815, opened 2022-02-13
Istio integration tests

piotrkpc:
introducing basic integration tests for Istio reconciler

Related issue:
https://github.com/kyma-incubator/reconciler/issues/798

### DIFF
--- a/pkg/reconciler/instances/istio/tests/integration_test.go
+++ b/pkg/reconciler/instances/istio/tests/integration_test.go
@@ -1,0 +1,167 @@
+package tests
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/avast/retry-go"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	"k8s.io/kubectl/pkg/util/podutils"
+)
+
+const (
+	istioNamespace         = "istio-system"
+	pilotIngressGwSelector = "istio in (pilot, ingressgateway)"
+	healthzTestURLFormat   = "http://%s:%s/healthz/ready"
+	gatewayManifest        = `
+apiVersion: networking.istio.io/v1beta1
+kind: Gateway
+metadata:
+  name: test-gateway
+spec:
+  selector:
+    istio: ingressgateway # use Istio default gateway implementation
+  servers:
+    - port:
+        number: 80
+        name: http
+        protocol: HTTP
+      hosts:
+        - "*"
+`
+	virutalServiceManifest = `
+apiVersion: networking.istio.io/v1beta1
+kind: VirtualService
+metadata:
+  name: istio-healthz
+spec:
+  gateways:
+  - istio-system/test-gateway
+  hosts:
+  - "*"
+  http:
+  - match:
+    - uri:
+        exact: /healthz/ready
+    route:
+    - destination:
+        host: istio-ingressgateway.istio-system.svc.cluster.local
+        port:
+          number: 15021
+`
+)
+
+var (
+	gatewayGVR = schema.GroupVersionResource{Resource: "gateways", Group: "networking.istio.io", Version: "v1beta1"}
+	vsGVR      = schema.GroupVersionResource{Resource: "virtualservices", Group: "networking.istio.io", Version: "v1beta1"}
+)
+
+func TestIstioIntegration(t *testing.T) {
+	skipTestIfDisabled(t)
+
+	setup := newIstioTest(t)
+	defer setup.contextCancel()
+
+	options := metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/instance=istio",
+	}
+
+	clientset, err := setup.kubeClient.Clientset()
+	require.NoError(t, err)
+
+	podsList, err := clientset.CoreV1().Pods(namespace).List(setup.context, options)
+	require.NoError(t, err)
+
+	for i, pod := range podsList.Items {
+		setup.logger.Infof("Pod %v is deployed", pod.Name)
+		require.Equal(t, v1.PodPhase("Running"), pod.Status.Phase)
+		ready := podutils.IsPodAvailable(&podsList.Items[i], 0, metav1.Now())
+		require.Equal(t, true, ready)
+	}
+
+	t.Run("istio pods should be running", func(t *testing.T) {
+		pilotSelectorOpt := metav1.ListOptions{LabelSelector: pilotIngressGwSelector}
+		deploymentList, err := clientset.AppsV1().Deployments(istioNamespace).List(setup.context, pilotSelectorOpt)
+		require.NoError(t, err)
+		require.True(t, len(deploymentList.Items) >= 2)
+		for _, dep := range deploymentList.Items {
+			require.Equal(t, *dep.Spec.Replicas, dep.Status.ReadyReplicas)
+			require.Equal(t, *dep.Spec.Replicas, dep.Status.AvailableReplicas)
+		}
+	})
+
+	t.Run("healthz returns 200", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cnvergence:
LGTM

cnvergence (approved):

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+	defer setup.contextCancel()
+
+	options := metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/instance=istio",
+	}
+
+	clientset, err := setup.kubeClient.Clientset()
+	require.NoError(t, err)
+
+	podsList, err := clientset.CoreV1().Pods(namespace).List(setup.context, options)
```
mjakobczyk, line 79: Could it become a separate test case? I mean lines 79-87

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+	require.NoError(t, err)
+
+	for i, pod := range podsList.Items {
+		setup.logger.Infof("Pod %v is deployed", pod.Name)
```
mjakobczyk, line 83: We could instead say that pod is deployed and running, because that's we essentially check in the next assertion

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+		require.Equal(t, v1.PodPhase("Running"), pod.Status.Phase)
+		ready := podutils.IsPodAvailable(&podsList.Items[i], 0, metav1.Now())
```
mjakobczyk, line 85: If pod is running doesn't it mean it is available?

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+		require.Equal(t, true, ready)
+	}
+
+	t.Run("istio pods should be running", func(t *testing.T) {
```
mjakobczyk, line 89: Imho the test base above verifies that pods are running, specifically because of the used assertion: `require.Equal(t, v1.PodPhase("Running"), pod.Status.Phase)` - shouldn't we rephrase this title then to something else?

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+		require.NoError(t, err)
+		require.True(t, len(deploymentList.Items) >= 2)
+		for _, dep := range deploymentList.Items {
+			require.Equal(t, *dep.Spec.Replicas, dep.Status.ReadyReplicas)
+			require.Equal(t, *dep.Spec.Replicas, dep.Status.AvailableReplicas)
+		}
+	})
+
+	t.Run("healthz returns 200", func(t *testing.T) {
+		gateway := &unstructured.Unstructured{}
```
mjakobczyk, line 101: Wouldn't it be a better idea to return a newly-created unstructured object from the function instead of passing a reference to it? With a return value the contract is clear at first glance. Now it works as a side-effect when you deserialise a yaml into an object

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+		deploymentList, err := clientset.AppsV1().Deployments(istioNamespace).List(setup.context, pilotSelectorOpt)
+		require.NoError(t, err)
+		require.True(t, len(deploymentList.Items) >= 2)
+		for _, dep := range deploymentList.Items {
+			require.Equal(t, *dep.Spec.Replicas, dep.Status.ReadyReplicas)
+			require.Equal(t, *dep.Spec.Replicas, dep.Status.AvailableReplicas)
+		}
+	})
+
+	t.Run("healthz returns 200", func(t *testing.T) {
```
mjakobczyk, line 100: I lack some consistency in this text. It is quite long - some things happened around like setup / cleanup and so on. How about dividing it into GWT sections for clarity? The same would be applicable for the rest of test cases

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+		gateway := &unstructured.Unstructured{}
+		vs := &unstructured.Unstructured{}
+		setupIstioGateway(t, gateway, setup)
+		defer cleanupObject(t, setup, gatewayGVR, gateway)
+		setupVirtualService(t, vs, setup)
+		defer cleanupObject(t, setup, vsGVR, vs)
+		healthzURL := readHealthzURL(t)
+		var statusCode int
+		err := retry.Do(func() error {
+			resp, err := http.Get(healthzURL) //nolint:gosec
```
mjakobczyk, line 110: Why was this `nolint` comment needed?

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+			statusCode = resp.StatusCode
+			return nil
+		}, retry.DelayType(retry.BackOffDelay))
+		require.Equal(t, http.StatusOK, statusCode)
+		require.NoError(t, err)
+	})
+}
+
+func readHealthzURL(t *testing.T) string {
+	clusterDomain, ok := os.LookupEnv("CLUSTER_DOMAIN")
```
mjakobczyk, line 123: You can extract this env to a constant above the test definition

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+	if !ok {
+		t.Fatal("CLUSTER_DOMAIN env var is required")
+	}
+	ingressPort, ok := os.LookupEnv("INGRESS_PORT")
```
mjakobczyk, line 127: As above

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+	createObject(t, setup, vsGVR, vs)
+}
+
+func cleanupObject(t *testing.T, setup *istioTest, gvr schema.GroupVersionResource, gw *unstructured.Unstructured) {
+	err := retry.Do(func() error {
+		deletionPropagation := metav1.DeletePropagationBackground
+		err := setup.dynamicClient.Resource(gvr).Namespace(istioNamespace).
+			Delete(setup.context, gw.GetName(), metav1.DeleteOptions{PropagationPolicy: &deletionPropagation})
+		return err
+	}, retry.Attempts(5))
```
mjakobczyk, line 151: Could we extract this magic number?

pkg/reconciler/instances/istio/tests/integration_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	"k8s.io/kubectl/pkg/util/podutils"
+)
+
+const (
+	istioNamespace         = "istio-system"
+	pilotIngressGwSelector = "istio in (pilot, ingressgateway)"
+	healthzTestURLFormat   = "http://%s:%s/healthz/ready"
+	gatewayManifest        = `
```
mjakobczyk, line 23: How about extracting these resources to dedicated files under some `resources/` dir? Maybe that would allow easier maintenance or adjustments in the future

pkg/reconciler/instances/istio/tests/integration_test.go
piotrkpc: Not necessary, running != ready != available

piotrkpc: From `podutils.IsPodAvailable help;
```there are two cases when a pod can be considered available: 1. minReadySeconds == 0, or 2. LastTransitionTime (is set) + minReadySeconds < current time```

pkg/reconciler/instances/istio/tests/integration_test.go
piotrkpc: see comment above

pkg/reconciler/instances/istio/tests/integration_test.go
piotrkpc: yes, `gosec` lint is to harsh for this test case

PR #831, opened 2022-02-16
Occupancy-tracker-for-component-reconcilers

khlifi411:
created a separate occupancy tracker for component reconciler

### DIFF
--- a/pkg/reconciler/service/occupancytracker.go
+++ b/pkg/reconciler/service/occupancytracker.go
@@ -14,38 +14,51 @@ import (
 	"time"
 )
 
-type occupancyTracker struct {
-	logger               *zap.SugaredLogger
-	occupancyID          string
-	occupancyCallbackURL string
-	poolSize             int
+type OccupancyTracker struct {
+	logger      *zap.SugaredLogger
+	occupancyID string
+	callbackURL string
 }
 
-func newOccupancyTracker(debug bool, poolSize int) *occupancyTracker {
-	return &occupancyTracker{
-		logger:      logger.NewLogger(debug),
-		occupancyID: getHostname(),
-		poolSize:    poolSize,
+func newOccupancyTracker(debug bool) *OccupancyTracker {
+	return &OccupancyTracker{
+		logger: logger.NewLogger(debug),
 	}
 }
 
-func (t *occupancyTracker) Track(ctx context.Context, pool *WorkerPool, reconcilerName string) {
+func (t *OccupancyTracker) Track(ctx context.Context, pool *WorkerPool, reconcilerName string) {
+	podName, err := os.Hostname()
+	if err != nil {
+		t.logger.Errorf("occupancy tracker could not retrieve pod name: %s", err)
+		return
+	}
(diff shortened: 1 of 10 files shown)


pkg/reconciler/service/occupancytracker.go
```diff
@@ -0,0 +1,117 @@
...
+		_, err = client.Do(req)
+		if err != nil {
+			t.logger.Warn(err.Error())
+		}
+	}
+}
+
+func getHostname() string {
+	hostname, err := os.Hostname()
+	if err != nil {
```
tobiscr, line 99: We should fire an error if this happens.. otherwise we start tracking values for an empty pool name

pkg/reconciler/service/workerpool.go
```diff
@@ -23,12 +16,11 @@ type workPoolBuilder struct {
...
-	debug                bool
-	logger               *zap.SugaredLogger
-	antsPool             *ants.Pool
-	PoolID               string
-	occupancyCallbackURL string
-	newRunnerFct         func(context.Context, *reconciler.Task, callback.Handler, *zap.SugaredLogger) func() error
+	debug        bool
+	logger       *zap.SugaredLogger
+	antsPool     *ants.Pool
+	CallbackURL  string
```
tobiscr, line 22: This is actually nothing the worker pool should have to know as it's occupancy-tracker specific.

Can'T we move it as field to the occupany-tracker struct?

pkg/reconciler/service/workerpool.go
```diff
@@ -179,3 +100,7 @@ func (wa *WorkerPool) IsClosed() bool {
 	}
 	return wa.antsPool.IsClosed()
 }
+
+func (wa *WorkerPool) RunningWorkers() int {
+	return wa.antsPool.Running()
+}
```
tobiscr, line 106: The worker pool size could be retrieved by the occupancyTracker by calling another function like `func (wa *WorkerPool) Size() int {...}`

tobiscr (changes requested):

pkg/reconciler/service/occupancytracker.go
```diff
@@ -0,0 +1,125 @@
...
+		return
+	}
+	resp, err := http.Post(t.occupancyCallbackURL, "application/json", bytes.NewBuffer(jsonPayload))
+	if err != nil {
+		t.logger.Error(err.Error())
+		return
+	}
+
+	if resp.StatusCode < http.StatusOK || resp.StatusCode > 299 {
+		t.logger.Errorf("mothership failed to update occupancy for '%s' service with status code: '%d'", t.occupancyID, resp.StatusCode)
```
tobiscr, line 79: Would treat it as warning - error means normally that a serious issue happend. This situation can also happen when the mothership is temporarily down.

tobiscr (approved):

pkg/reconciler/service/occupancytracker.go
khlifi411: Thanks! Will modify it in the subsequent PR 👍

PR #833, opened 2022-02-16
Revert "Expose processing duration metrics over Prometheus"

tobiscr:
Reverts kyma-incubator/reconciler#808


tobiscr:
Fixed in https://github.com/kyma-incubator/reconciler/pull/840

PR #835, opened 2022-02-17
Load Istio target version from values.yaml instead of Chart.yaml

mjakobczyk:
### Description

Suggested changes:
- remove loading Istio target version from Chart.yaml
- load Istio target version from values.yaml
- fix tests
- provide compatibility for Kyma `main` and `release-2.0` branches

### Links


piotrkpc (dismissed):

piotrkpc (approved):
/lgtm

PR #840, opened 2022-02-17
Leverage heartbeat to send processing Duration instead of own endpoint

### LINKED ISSUE #836: The `pre-main-control-plane-reconciler-e2e` pipeline fails temporarily caused by colliding DB-TXs
**Description**

The pipeline `pre-main-control-plane-reconciler-e2e` is failing casued by colliding DB-TXs. Issue is caused by trying to update the same operation twice at the same time in different threads.

This has to be solved by updating the same operation just once:

https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/pr-logs/directory/pre-main-control-plane-reconciler-e2e

**Expected result**

No DB-TX collisions are not happening during regular reconciliation runs.

**Actual result**

DB collisions are happening and lead to a flaky reconciliation result.

**Steps to reproduce**

**Troubleshooting**

jeremyharisch:
Instead of updating the processingDuration using an own endpoint, the heartbeat is leveraged to also update it. Thus, we do not have two transaction updating the same row in the scheduler_operations table.

Fixes #836

### DIFF
--- a/cmd/mothership/mothership/start/http.go
+++ b/cmd/mothership/mothership/start/http.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/google/uuid"
+	"github.com/kyma-incubator/reconciler/pkg/db"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -96,11 +97,6 @@ func startWebserver(ctx context.Context, o *Options) error {
 	apiRouter.HandleFunc(
 		fmt.Sprintf("/v{%s}/operations/{%s}/callback/{%s}", paramContractVersion, paramSchedulingID, paramCorrelationID),
 		callHandler(o, operationCallback)).
-		Methods(http.MethodPost)
-
-	apiRouter.HandleFunc(
-		fmt.Sprintf("/v{%s}/operations/{%s}/callback/{%s}/processingDuration", paramContractVersion, paramSchedulingID, paramCorrelationID),
-		callHandler(o, updateOperationProcessingDuration)).
 		Methods(http.MethodPost)
 
 	apiRouter.HandleFunc(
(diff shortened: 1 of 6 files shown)


cmd/mothership/mothership/start/http.go
```diff
@@ -917,6 +892,19 @@ func updateOperationStateAndRetryID(o *Options, schedulingID, correlationID, ret
 	return err
 }
 
+func updateOperationStateAndRetryIDAndProcessingDuration(o *Options, schedulingID, correlationID, retryID string, state model.OperationState, processingDuration int, reason ...string) error {
+	err := updateOperationStateAndProcessingDuration(o, schedulingID, correlationID, state, processingDuration, reason...)
+	if err != nil {
+		return err
+	}
+	err = o.Registry.ReconciliationRepository().UpdateOperationRetryID(schedulingID, correlationID, retryID)
```
tobiscr, line 900: Do we need this function? Can't we just add this line also to `updateOperationStateAndProcessingDuration` ?

tobiscr (changes requested):

cmd/mothership/mothership/start/http.go
jeremyharisch: I left it here, since they are dealing with different tables, and ot keep the function names as "short" as possible 😅 
But we could also add it there yes

cmd/mothership/mothership/start/http.go
```diff
@@ -904,6 +855,30 @@ func updateOperationState(o *Options, schedulingID, correlationID string, state
...
+
+		err = rTx.UpdateComponentOperationProcessingDuration(schedulingID, correlationID, processingDuration)
+		if err != nil {
+			o.Logger().Errorf("REST endpoint failed to update operation processingDuration (schedulingID:%s/correlationID:%s) "+
+				"to '%s': %s", schedulingID, correlationID, state, err)
+		}
+		return err
+	}
+	return db.Transaction(o.Registry.Connection(), dbOps, o.Logger())
+}
```
tobiscr, line 900: Refactoring planned in https://github.com/kyma-incubator/reconciler/issues/842

tobiscr (approved):

PR #845, opened 2022-02-17
fix istio versions comparison for new images

piotrkpc:
## Description

Changes proposed:
- remove custom `helperVersion` implementation
- adopt `semver` implementation from `"github.com/coreos/go-semver/semver"`
- adjust API and tests

## Links


werdes72 (approved):
/woof

PR #864, opened 2022-02-24
Add handleDisableJetstream preaction for disabling the jetstream feature flag

VladislavPaskar:
**Description**

There is a `enableJetStream` flag in the eventing helm values which is then used to inject the Jetsream-specific data into the nats ConfigMap, StatefulSet and Pods. There is a problem in the upgrade scenario from `enableJetStream=true` to `enableJetStream=false`, as the JS-related data is not being removed from the statefulSet. 

- [ ] Before merging this PR, another PR https://github.com/kyma-project/kyma/pull/13496 should be merged

**Changes Proposed**
This pr adds a preaction to the eventing component reconciler in order to check this problem and remove the StatefulSet during the upgrade.

**Related Issue**
https://github.com/kyma-project/kyma/issues/13321


pxsalehi (changes requested):
https://github.com/kyma-project/kyma/issues/13321#issuecomment-1054191924

PR #871, opened 2022-02-25
Add prometheus metrics for scmigration reconciler

wozniakjan:
Service Catalog migration reconciler processes high volume of data and prometheus metrics would provide great insight and instrumentation for the whole process. Example output:
```
$ curl localhost:8082/metrics | grep migra

# HELP migrated_bindings_total Migrated bindings total
# TYPE migrated_bindings_total gauge
migrated_bindings_total{instance_id="test-jw148"} 3
# HELP migrated_instances_total Migrated instances total
# TYPE migrated_instances_total gauge
migrated_instances_total{instance_id="test-jw148"} 3
```

### DIFF
--- a/cmd/reconciler/start/service/http.go
+++ b/cmd/reconciler/start/service/http.go
@@ -13,10 +13,15 @@ import (
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/service"
 	"github.com/kyma-incubator/reconciler/pkg/server"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 const (
 	paramContractVersion = "version"
+)
+
+var (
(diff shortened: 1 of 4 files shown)


tillknuesting (approved):

tillknuesting (approved):

tobiscr:
As agreed in Slack, we have to assume that one component reconciler runs with multiple replicas. The monitoring is not able to query each component reconciler for it'S current status. Therefore, we have to find a way to either aggregate all exposed monitoring data on the component-reconciler side or to send them to a central place where they get aggregated (like we do it for occupancy tracking of our worker pools - e.g. #831 ).

wozniakjan:
also attaching another PR that I received for reference from @khlifi411 
https://github.com/kyma-incubator/reconciler/pull/808

PR #872, opened 2022-02-25
Fix metrics error

jeremyharisch:
The situation that no metrics for a specific component with a specific state is not found, should not be an error.

### DIFF
--- a/pkg/scheduler/reconciliation/persistent.go
+++ b/pkg/scheduler/reconciliation/persistent.go
@@ -551,7 +551,7 @@ func (r *PersistentReconciliationRepository) GetComponentOperationProcessingDura
 		return 0, err
 	}
 	if len(operations) == 0 {
-		r.Logger.Infof("No operation for component %s found with desired state %s", component, state)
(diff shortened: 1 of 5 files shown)


pkg/scheduler/reconciliation/persistent.go
```diff
@@ -551,7 +551,8 @@ func (r *PersistentReconciliationRepository) GetComponentOperationProcessingDura
 		return 0, err
 	}
 	if len(operations) == 0 {
-		return 0, errors.Errorf("No operation for component %s found with desired state %s", component, state)
+		r.Logger.Infof("No operation for component %s found with desired state %s", component, state)
```
tobiscr, line 554: Not sure if even `Info` is too much.. do we have to know it as it's important to trace things or is it just interesting when a developer is looking on it?

Maybe `Debug` could be better suited

jeremyharisch: Changed

ruanxin (approved):

PR #876, opened 2022-02-28
Disbale processingDuration Metrics by default - Introduce Feature flags pkg

jeremyharisch:
This PR introduces the simple `features` pkg. Thus, we can use this as a central point to check which features are enabled using env Vars.

Until now it is very trivial, but thus we have a good starting point for further improvements.


pkg/features/features.go
```diff
@@ -0,0 +1,25 @@
+package features
```
jakobmoellersap, line 1: Just a suggestion: Maybe we want to introduce a Repo Pattern here, in case we want to switch from Env Var to something different, e.g. an external Feature Flag Management or DB / Cache-Driven Flags?

jeremyharisch: Tbh I do not know if it would make sense in our use-case, but we can definitive challenge it in future, but for now we need to disable the new Metrics asap, to get it promoted to STAGE

ruanxin (approved):

PR #877, opened 2022-02-28
Propagate State to ReconciliationSequenceConfig to disable cleaner reconciler on normal reconciliations

### LINKED ISSUE #715: Cleaner reconciler runs on non-delete reconciliations
**Description**
The `cleaner` internal component is added to the reconciliation sequence always when starting a reconciliation. But it should only be added to "delete" reconciliations.

**Expected result**

Cleaner does not run on regular reconciliations, only when deleting.

**Actual result**

Cleaner always gets added to the reconciliation sequence regardless of the type of reconciliation.

**Steps to reproduce**
Do a reconciliation in either the CLI or SKR and see cleaner component being deployed.

**Troubleshooting**
When creating the reconciliaiton sequence we only have to add the component when it is a delete reconciliation.

jakobmoellersap:
fixes #715

Allows distinguishing between delete reconciliations and normal reconciliations, making sure the cleaner reconciler is only scheduled in case of delete.

Adds test case in unit test and adopts scheduler test to be status-aware.

Sample of Cluster Deploy without cleaner reconciler

Edit:
Due to a change of ClusterState, this requires an upstream change in the way CLI is handing over cluster state:
https://github.com/kyma-project/cli/blob/3f016792a4f359a395c1db68aa5704af452c98fb/internal/deploy/deploy.go#L129-L131

this will need to be changed to 
```go
status := model.ClusterStatusReconciling
if delete {
	status = model.ClusterStatusDeleting
}
```

### DIFF
--- a/pkg/model/clusterconfigentity.go
+++ b/pkg/model/clusterconfigentity.go
@@ -120,6 +120,7 @@ type ReconciliationSequence struct {
 type ReconciliationSequenceConfig struct {
 	PreComponents  [][]string
 	DeleteStrategy string
+	DesiredStatus  Status
 }
 
 func newReconciliationSequence(cfg *ReconciliationSequenceConfig) *ReconciliationSequence {
@@ -129,10 +130,16 @@ func newReconciliationSequence(cfg *ReconciliationSequenceConfig) *Reconciliatio
 	reconSeq.Queue = append(reconSeq.Queue, []*keb.Component{ //CRDs are always processed at the very beginning (or at the very end in deletion)
 		crdComponent,
 	})
-	cleanupComponent.Configuration = append(cleanupComponent.Configuration, keb.Configuration{Key: DeleteStrategyKey, Value: cfg.DeleteStrategy})
-	reconSeq.Queue = append(reconSeq.Queue, []*keb.Component{
-		cleanupComponent,
-	})
+
+	// if a cluster is pending deletion, we need to add the cleanup component into the reconciliation
+	if cfg.DesiredStatus.IsDeletion() {
+		cleanupComponent.Configuration = append(cleanupComponent.Configuration, keb.Configuration{
+			Key: DeleteStrategyKey, Value: cfg.DeleteStrategy,
+		})
+		reconSeq.Queue = append(reconSeq.Queue, []*keb.Component{
+			cleanupComponent,
+		})
+	}
 
 	return reconSeq
 }
--- a/pkg/model/clusterconfigentity.go
+++ b/pkg/model/clusterconfigentity.go
@@ -118,9 +118,9 @@ type ReconciliationSequence struct {
 }
 
 type ReconciliationSequenceConfig struct {
-	PreComponents  [][]string
-	DeleteStrategy string
-	DesiredStatus  Status
+	PreComponents        [][]string
+	DeleteStrategy       string
+	ReconciliationStatus Status
 }
 
 func newReconciliationSequence(cfg *ReconciliationSequenceConfig) *ReconciliationSequence {
@@ -132,7 +132,7 @@ func newReconciliationSequence(cfg *ReconciliationSequenceConfig) *Reconciliatio
 	})
 
 	// if a cluster is pending deletion, we need to add the cleanup component into the reconciliation
-	if cfg.DesiredStatus.IsDeletionInProgress() {
+	if cfg.ReconciliationStatus.IsDeletionInProgress() {
 		cleanupComponent.Configuration = append(cleanupComponent.Configuration, keb.Configuration{
 			Key: DeleteStrategyKey, Value: cfg.DeleteStrategy,
 		})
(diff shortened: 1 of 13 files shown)


pkg/model/clusterconfigentity.go
```diff
@@ -129,10 +130,16 @@ func newReconciliationSequence(cfg *ReconciliationSequenceConfig) *Reconciliatio
 	reconSeq.Queue = append(reconSeq.Queue, []*keb.Component{ //CRDs are always processed at the very beginning (or at the very end in deletion)
 		crdComponent,
 	})
-	cleanupComponent.Configuration = append(cleanupComponent.Configuration, keb.Configuration{Key: DeleteStrategyKey, Value: cfg.DeleteStrategy})
-	reconSeq.Queue = append(reconSeq.Queue, []*keb.Component{
-		cleanupComponent,
-	})
+
+	// if a cluster is pending deletion, we need to add the cleanup component into the reconciliation
+	if cfg.DesiredStatus.IsDeletion() {
```
tobiscr, line 135: Shouldn 't we use here`IsDeleteCandidate`  ?

pkg/scheduler/invoker/localinvoker_test.go
```diff
@@ -65,7 +65,11 @@ func runLocalReconciler(t *testing.T, simulateError bool) (reconciliation.Reposi
 		SchedulingID: reconEntity.SchedulingID,
 	})
 	require.NoError(t, err)
-	require.Len(t, opEntities, 7)
+	if clusterStateMock.Status.Status.IsDeletion() {
```
tobiscr, line 68: Shoulnd't it be `IsDeleteCandidate` ?

pkg/scheduler/invoker/remoteinvoker_test.go
```diff
@@ -33,7 +33,11 @@ func TestRemoteInvoker(t *testing.T) {
 		SchedulingID: reconEntity.SchedulingID,
 	})
 	require.NoError(t, err)
-	require.Len(t, opEntities, 7)
+	if clusterStateMock.Status.Status.IsDeletion() {
```
tobiscr, line 36: as above

pkg/scheduler/service/scheduler_test.go
```diff
@@ -86,7 +86,11 @@ func requiredReconciliationEntity(t *testing.T, reconRepo reconciliation.Reposit
 		SchedulingID: recons[0].SchedulingID,
 	})
 	require.NoError(t, err)
-	require.Len(t, ops, 3)
+	if state.Status.Status.IsDeletion() {
```
tobiscr, line 89: as above

tobiscr (dismissed):

pkg/model/clusterconfigentity.go
jakobmoellersap: I asked @clebs if we should use deleteCandidate and I think we decided for deletion. To be honest I don't quite remember why anymore. Could you jump in and help me here again?

jakobmoellersap: Did a change to accurately reflect the Deletion after discussion with @clebs , but this now requires a small upstream change in the CLI in order to not break it.

pkg/model/clusterconfigentity.go
```diff
@@ -120,6 +120,7 @@ type ReconciliationSequence struct {
 type ReconciliationSequenceConfig struct {
 	PreComponents  [][]string
 	DeleteStrategy string
+	DesiredStatus  Status
```
clebs, line 123: I would suggest a different name for this field, it is not a status that needs to be reached but actually the status the reconciliation for which we are creating the sequence is in right now.

Maybe something like `ReconciliationStatus` would be better.

pkg/scheduler/invoker/remoteinvoker_test.go
```diff
@@ -33,7 +33,11 @@ func TestRemoteInvoker(t *testing.T) {
 		SchedulingID: reconEntity.SchedulingID,
 	})
 	require.NoError(t, err)
-	require.Len(t, opEntities, 7)
+	if clusterStateMock.Status.Status.IsDeletionInProgress() {
+		require.Len(t, opEntities, 7, "reconciliation sequence has 7 ops (5 + crds + cleaner)")
```
clebs, line 37: Cool description for the assertion.

pkg/scheduler/invoker/localinvoker_test.go
clebs: It is actually correct to have `isDelettionInProgress()` since when `run` and `runOnce` are called the status should be always either `Reconciling` or `Deleting`.

What is wrong is the test status created in `testClusterState()`. It needs to be changed to an in progress status.

jakobmoellersap: I think as of now i check on isDelettionInProgress(). Nevertheless Im not sure where to change the test status.

jakobmoellersap: I changed the test behavior as discussed

jakobmoellersap:
@clebs could you check out again

clebs:
Hi @jakobmoellersap,
the tests failing do not seem to be related with your PR. The jobs running the unit and integration tests you changed are green.

Maybe there have been some changes in `main` since you opened this branch, I would suggest rebasing this PR with `main` and try again.

jakobmoellersap:
@clebs i think protecode has issues currently, not only failing on this job, other than that it should be good.

clebs (approved):

PR #885, opened 2022-03-01
Update CODEOWNERS

adityabhatia:
Add new members to CODEOWNERS: @jakobmoellersap @adityabhatia


adityabhatia:
TODO: Also found that some users in CODEOWNERS do not exist anymore, which needs to be cleaned

tobiscr (approved):

PR #897, opened 2022-03-04
feat: create GORM compatible postgres-layer

### LINKED ISSUE #851: POC: verify effort to replace custom ORM with `GORM` library
**Description**

The decision for using a custom ORM layer has to be challenged after it was decided to drop the support for SQLite. We should investigate the effort for removing / replacing the custom ORM by the well established [`go-pg`](https://github.com/go-pg/pg) library.

AC:
* Investigate how a migration from customer ORM to `go-pg` library is possible without a "big bag" change
* Demonstrate the efforts for migrating existing logic in one of our data repositories (e.g. `cluster.Inventory`) to use `go-pg` internally.
* Share identified pros / deltas within the team and make a suggestion whether a migration would be a valuable investment.

**Reasons**

After changed requirements on DB-ORM support level, we should challenge whether a customer ORM is the maintenance effort still worth. Establishing a mainstream-library could reduce here maintenance efforts and potential risks for security gaps.

**Attachments**

jakobmoellersap:
fixes #851 

this adds OrmConnection that is backwards-compatible to OrmConnection
but also has full fledged compatibility towards GORM and makes use of a
synced connection pool


jakobmoellersap:
Closing due to preference in goqu

Issue #794, opened 2022-02-08
Cleanup SQLite related code


ruanxin:

janmedrek:
Functionalities of the Reconciler were migrated to the new architecture, this issue will not be worked on.

PR #905, opened 2022-03-07
Create DB Views for calls to speed up SQL lookup on Max aggregrations for Inventory

### LINKED ISSUE #733: Improve DB queries which take >10ms
**Description**

The reconciler is still firing DB queries which have sometimes processing efforts far above 10ms.

These queries have to be investigated and optimised (e.g. by adding further indexes etc).

Please use [GCP DB insight](https://console.cloud.google.com/sql/instances/kyma-mps-stage-730c/insights;duration=PT1H;sort_by=TOTAL_EXEC_TIME?project=sap-ti-dx-kyma-mps-stage&ref=https:%2F%2Faccounts.google.com%2F&pli=1&rapt=AEjHL4NVNKfj7pzBRTsQ1T1AuWQNDNB-BXGd7SfU0Ab4b8Cmn82SDV7j9ukgGQ77YcA2_ieFUcXoYS1-L3UT1kPyy233ECozPw) for getting an overview of bad performing SQL statemets.

AC:
* GCP DB insights doesn't show any SQL queries fired by reconciler which are much higher than 10ms.

**Reasons**

Ensure DB is well performing even under high load.

**Tasks**

- [x] https://github.com/kyma-incubator/reconciler/pull/910
- [ ] https://github.com/kyma-incubator/reconciler/issues/851
- [x] https://github.com/kyma-incubator/reconciler/pull/905
- [x] https://github.com/kyma-incubator/reconciler/issues/920
- [x] https://github.com/kyma-incubator/reconciler/issues/921

**Attachments**

jakobmoellersap:
v_active_inventory_cluster_configs adds a view to select all currently
active cluster configurations

v_active_inventory_cluster_config_status_history adds a view to select
all status changes for currently active cluster configurations

v_active_inventory_cluster_latest_status adds a view to select the
latest status from the cluster configs and the status history


pkg/cluster/statussqlfilter.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	createdColName, err := statusColHdr.ColumnName(StatusCreatedAtColumnName)
+	if err != nil {
+		return "", err
+	}
+	switch dbType {
+	case db.Postgres:
+		return fmt.Sprintf(`%s IN ('%s', '%s', '%s') AND %s <= NOW() - INTERVAL '%.0f SECOND'`,
+			statusColName, model.ClusterStatusReady, model.ClusterStatusReconcileErrorRetryable, model.ClusterStatusDeleteErrorRetryable, createdColName, rif.reconcileInterval.Seconds()), nil
+	case db.SQLite:
```
ruanxin, line 61: We are deprecating `SQLite`, and already stop to use it in any case, there is dedicate issue for cleanup the related code https://github.com/kyma-incubator/reconciler/issues/794.
This switch case we don't need to introduce.

pkg/cluster/statussqlfilter.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	createdColName, err := statusColHdr.ColumnName(CreatedAtColumnName)
+	if err != nil {
+		return "", err
+	}
+	switch dbType {
+	case db.Postgres:
+		return fmt.Sprintf(`%s = '%s' AND %s >= NOW() - INTERVAL '%.0f SECOND'`,
+			runtimeIDColName, rif.runtimeID, createdColName, rif.interval.Seconds()), nil
+	case db.SQLite:
```
ruanxin, line 87: Same here as https://github.com/kyma-incubator/reconciler/pull/905/files#r823744214

pkg/cluster/statussqlfilter.go
jakobmoellersap: This is not introduced but instead is unified into one file. You can see we just moved this function to statussqlfilter.go from statusfilter.go. we didn't want to touch the logic because of this exact refactoring in #794, but we can also remove it now. What do you think?

pkg/cluster/statussqlfilter.go
jakobmoellersap: Same reasoning above

pkg/cluster/statussqlfilter.go
ruanxin: ok, sorry for this misunderstanding, then let's keep it.

ruanxin (approved):

PR #910, opened 2022-03-08
fix: introduce access lock for unique db pointer and connection pooling

jakobmoellersap:


jakobmoellersap:
seems like we have quite some isolation issues...

jeremyharisch:
Before introducing this into the reconciler we need to adress the following:
- How does the WorkerPool behave when the workerPool is fully occupied?
- Will there be an error? - We need one, since if it only blocks this could slow down our routines.
- Is it traceable?

After those questions are answered successfully, we can implement it to our needs.

The WorkerPool size should then be less or equal to our Postgres configuration. AFAIK it is set to 1000 connections, but we have to double check.

tobiscr (approved):

pkg/cluster/inventory.go
```diff
@@ -27,6 +28,7 @@ type Inventory interface {
 	ClustersNotReady() ([]*State, error)
 	CountRetries(runtimeID string, configVersion int64, maxRetries int, errorStatus ...model.Status) (int, error)
 	WithTx(tx *db.TxConnection) (Inventory, error)
+	DBStats() *sql.DBStats
```
tobiscr, line 31: Would move this to the Connection-Interface

tobiscr (changes requested):

cmd/mothership/mothership/start/http.go
```diff
@@ -128,6 +128,7 @@ func startWebserver(ctx context.Context, o *Options) error {
 	metrics.RegisterOccupancy(o.Registry.OccupancyRepository(), o.Config.Scheduler.Reconcilers, o.Logger())
 	metrics.RegisterProcessingDuration(o.Registry.ReconciliationRepository(), o.Logger())
 	metrics.RegisterWaitingAndNotReadyReconciliations(o.Registry.Inventory(), o.Logger())
+	metrics.RegisterDbPool(o.Registry.Inventory(), o.Logger())
```
tobiscr, line 131: QQ.. why are we using the inventory as source for the DBStats? All repositories are sharing the same sql.Db instance (=Pool) as they are using all the same conneciton (see https://github.com/kyma-incubator/reconciler/blob/main/internal/persistency/registry.go#L33) - shouldn't we add this `Stats()` function to the DB-connection interface ?

tobiscr (changes requested):

cmd/mothership/mothership/start/http.go
jakobmoellersap: Quick question, the code you send me shows that a new repository is also initializing a new DB Pool due to `conn, err := cf.NewConnection()` no? nevertheless, we can surely move the logic to the connection layer, i for sure agree

jakobmoellersap:
@tobiscr pinging for rereview

tobiscr (approved):

Issue #884, opened 2022-03-01
Enhance cleaner-module in mothership to NOT remove the latest reconciliation


tobiscr:
**Description**

The existing deletion strategy for removing reconciliation data entries has to be adjusted a bit: 

1.  any reconciliation which is older than Y days is deleted except it is most recent reconciliation - this should survive always.

2. any reconciliation between now and Y days is NOT deleted if

    2.1 the reconciliation is one of the most recent X reconciliations of a cluster
    2.2 or if the reconciliation finished with a non-ready state

This is required to avoid that histories of a non-successfully reconciled cluster got empty over time.

**Reasons**

Avoid deletion of the latest reconciliation which is in a non-successfully state to avoid "history less" reconciliation entries.

**Attachments**

Tomasz-Smelcerz-SAP:
See also: https://github.com/kyma-incubator/reconciler/issues/668

Tomasz-Smelcerz-SAP:
PR for the issue: https://github.com/kyma-incubator/reconciler/pull/912

PR #912, opened 2022-03-08
Improve cleaner component

Tomasz-Smelcerz-SAP:
**Description**
Improve cleaner component according to: https://github.com/kyma-incubator/reconciler/issues/884

**Changes:**
- Extend cleaning logic according to AC
- Rename command-line flags to better reflect new logic
- Extend test mocks to return different results for multiple invocation of GetReconciliations() function.

See also: https://github.com/kyma-incubator/reconciler/issues/668

### DIFF
--- a/pkg/model/clusterentity.go
+++ b/pkg/model/clusterentity.go
@@ -17,7 +17,7 @@ type ClusterEntity struct {
 	RuntimeID  string            `db:"notNull"`
 	Runtime    *keb.RuntimeInput `db:"notNull"`
 	Metadata   *keb.Metadata     `db:"notNull"`
-	Kubeconfig string            `db:"notNull"`
+	Kubeconfig string            `db:"notNull,encrypt"`
 	Contract   int64             `db:"notNull"`
 	Deleted    bool              `db:"notNull"`
 	Created    time.Time         `db:"readOnly"`
--- a/pkg/model/clusterentity.go
+++ b/pkg/model/clusterentity.go
@@ -17,7 +17,7 @@ type ClusterEntity struct {
 	RuntimeID  string            `db:"notNull"`
 	Runtime    *keb.RuntimeInput `db:"notNull"`
 	Metadata   *keb.Metadata     `db:"notNull"`
-	Kubeconfig string            `db:"notNull,encrypt"`
+	Kubeconfig string            `db:"notNull"`
 	Contract   int64             `db:"notNull"`
 	Deleted    bool              `db:"notNull"`
 	Created    time.Time         `db:"readOnly"`
--- a/pkg/scheduler/service/cleaner.go
+++ b/pkg/scheduler/service/cleaner.go
@@ -10,18 +10,18 @@ import (
 )
 
 type CleanerConfig struct {
-	PurgeEntitiesOlderThan       time.Duration
-	CleanerInterval              time.Duration
-	KeepLatestEntitiesCount      uint
-	KeepUnsuccessfulEntitiesDays uint
+	PurgeEntitiesOlderThan  time.Duration
+	CleanerInterval         time.Duration
+	KeepLatestEntitiesCount uint
+	MaxEntitiesAgeDays      uint
 }
 
 func (c *CleanerConfig) keepLatestEntitiesCount() int {
 	return int(c.KeepLatestEntitiesCount)
 }
 
-func (c *CleanerConfig) keepUnsuccessfulEntitiesDays() int {
-	return int(c.KeepUnsuccessfulEntitiesDays)
+func (c *CleanerConfig) maxEntitiesAgeDays() int {
+	return int(c.MaxEntitiesAgeDays)
 }
 
 type cleaner struct {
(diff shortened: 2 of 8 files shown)


adityabhatia:
Batch delete of reconciliation addressed here https://github.com/kyma-incubator/reconciler/pull/914

pkg/scheduler/service/cleaner.go
```diff
@@ -67,154 +69,194 @@ func (c *cleaner) purgeReconciliations(transition *ClusterStatusTransition, conf
 
 //Purges reconciliations using rules from: https://github.com/kyma-incubator/reconciler/issues/668
 func (c *cleaner) purgeReconciliationsNew(transition *ClusterStatusTransition, config *CleanerConfig) {
-	now := time.Now()
 
-	latestReconciliations, err := c.getLatestReconciliations(config.keepLatestEntitiesCount(), transition)
+	clusters, err := transition.inventory.GetAll()
```
jakMoeller, line 73: I think this can be a very expensive operation if we get all clusters, is there any way we can instead use a query here that already takes into account the age through config.maxEntitiesAgeDays() and the count and status criteria? Maybe I overestimate this, but I think that getting all clusters can have thousands of records through which you iterate both `deleteRecordsByAge` and `deleteRecordsByCountAndStatus`, even if they end up as NoOp. Maybe a delete driven by the database is more efficient here. Let me know what you think :) I see that you are using these methods maybe also for readability and configurability but I think the delete could also be written as one single query

jakobmoellersap, line 73: I think this can be a very expensive operation if we get all clusters, is there any way we can instead use a query here that already takes into account the age through config.maxEntitiesAgeDays() and the count and status criteria? Maybe I overestimate this, but I think that getting all clusters can have thousands of records through which you iterate both deleteRecordsByAge and deleteRecordsByCountAndStatus, even if they end up as NoOp. Maybe a delete driven by the database is more efficient here. Let me know what you think :) I see that you are using these methods maybe also for readability and configurability but I think the delete could also be written as one single query

Tomasz-Smelcerz-SAP: @jakobmoellersap: TLDR - We will improve things with extensions to the ReconciliationRepository (currently under development). The logic that is currently used is not much better, it also removes records one by one, but it doesn't have to list the clusters. Still I think after the improvements we'll run faster than current logic.

We want to use our object model (simple ORM), not direct queries, unless it's the only option.
So we are extending the functionality of ReconciliationRepository to make this code more efficient in the future - it's already reflected in the functions composition.
We could probably achieve a "single-delete statement" solution here even with the object model, but It would require substantial coding effort and it may be not necessary. Perhaps just the single-delete statement of entries older than "Y" days is enough to cut down the record set to a reasonable amount, e.g: 100 per cluster.
Then we can filter this out in the memory and remove remaining records with another single-delete statement using: 'WHERE IN ("id1", "id2", ...') syntax.

As for the clusters you are right, there may be many of them, I'd expect up to 500 per environment.
But we always have to delete records in the context of the single cluster. I don't see any other way of doing that, except of some stored procedure that runs entirely in the database (we don't want it for now AFAIK).

Finally, the cleaner is a cron job and it is expected to be a long-running process, so maybe it's not an issue. If, after improvements, we can process a cluster in a 10 seconds (easily doable), then we can process 360 clusters in an hour - that should be enough.

pkg/model/clusterentity.go
```diff
@@ -17,7 +17,7 @@ type ClusterEntity struct {
 	RuntimeID  string            `db:"notNull"`
 	Runtime    *keb.RuntimeInput `db:"notNull"`
 	Metadata   *keb.Metadata     `db:"notNull"`
-	Kubeconfig string            `db:"notNull,encrypt"`
+	Kubeconfig string            `db:"notNull"`
```
tobiscr, line 20: this kind of data should stay encrypted actually

tobiscr (changes requested):

pkg/scheduler/service/cleaner.go
jakobmoellersap: I totally agree with the TLDR.

Without extending further on it, I think this "ORM" (which i think is just a row mapper since we do not manage relations) is already being challenged as low prio issue, so perfectly ok that we limit our effort here for now.

I think the stored procedure is not the only way to optimize deletes here, but this we can also talk about again then in detail when it comes to our other optimisation efforts on our DB layer in general.

Thanks for the explanation!

Tomasz-Smelcerz-SAP: @jakobmoellersap One more thing. If this approach is not efficient enough, I'd consider a view-based approach. We're currently introducing views to optimize some other cleaning process. Once it's established, we can do the same here.
But I think it won't be necessary here. We'll see.

pkg/model/clusterentity.go
Tomasz-Smelcerz-SAP: Thanks for spotting that 👍 I did that to test the dump locally and forgot to remove :(

adityabhatia (approved):

tobiscr (approved):

PR #913, opened 2022-03-08
Add hydra-maester startup shift

veichtj:
Add a startup shift for check if hydra sync needs to be called, which should enforce hdyra-maester to synchronize OAuth2Clients to hydra and prevent failing pipelines :https://github.com/kyma-project/kyma/issues/13498

### DIFF
--- a/pkg/reconciler/instances/ory/hydra/syncer.go
+++ b/pkg/reconciler/instances/ory/hydra/syncer.go
@@ -35,6 +35,7 @@ const (
 	hydraPodName           = "app.kubernetes.io/name=hydra"
 	hydraMaesterPodName    = "app.kubernetes.io/name=hydra-maester"
 	hydraMaesterDeployment = "ory-hydra-maester"
+	startupShift           = -2 * time.Second
 )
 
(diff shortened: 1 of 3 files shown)


pkg/reconciler/instances/ory/hydra/syncer.go
```diff
@@ -74,8 +75,7 @@ func restartHydraMaesterDeploymentNeeded(context context.Context, client kuberne
 		return false, err
 	}
 	logger.Debugf("Earliest hydra-maester restart time: %s ", earliestHydraMaesterPodStartTime.String())
-
-	return earliestHydraPodStartTime.After(earliestHydraMaesterPodStartTime), nil
+	return earliestHydraPodStartTime.After(earliestHydraMaesterPodStartTime.Add(startupShift)), nil
```
cnvergence, line 78: Let's apply this as a temporary solution for syncing Ory, but I'm not generally okay with this.
We should aim for another mechanism checking if the synchronization is needed between ory-hydra and ory-hydra-maester, hard-coded shift time.Duration should be removed in the future.

cnvergence (approved):
> Let's apply this as a temporary solution for syncing Ory, but I'm not generally okay with this. We should aim for another mechanism checking if the synchronization is needed between ory-hydra and ory-hydra-maester, hard-coded shift time.Duration should be removed in the future.

We can create the issue for it.

cnvergence (approved):

PR #914, opened 2022-03-08
Reconciliations bulk delete

adityabhatia:


pkg/scheduler/reconciliation/reconciliation.go
```diff
@@ -20,6 +20,7 @@ const (
 type Repository interface {
 	CreateReconciliation(state *cluster.State, cfg *model.ReconciliationSequenceConfig) (*model.ReconciliationEntity, error)
 	RemoveReconciliation(schedulingID string) error
+	RemoveReconciliations(reconEntities []*model.ReconciliationEntity) error
```
Tomasz-Smelcerz-SAP, line 23: I think this interface method is requiring "too much" - you have to load entire object into memory just to delete it. It would be much better to only require a slice of schedulingIDs (strings).

pkg/scheduler/reconciliation/inmemory.go
```diff
@@ -106,6 +106,18 @@ func (r *InMemoryReconciliationRepository) RemoveReconciliation(schedulingID str
 	return nil
 }
 
+func (r *InMemoryReconciliationRepository) RemoveReconciliations(reconEntities []*model.ReconciliationEntity) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	for _, recon := range reconEntities {
+		delete(r.reconciliations, recon.RuntimeID)
```
Tomasz-Smelcerz-SAP, line 114: I think the entire implementation could be just iteration over schedulingIDs and delegating the work to the already existing method.
This may look inefficient, but the in-memory implementation is very fast and it's used only in tests AFAIK.
Considering the above, I'd change the signature of this method to only require a slice of `schedulingIDs`.

adityabhatia:
Block size 5000
Screen Shot 2022-03-09 at 00 38 36

Block size 200
Screen Shot 2022-03-09 at 00 29 03

Existing implementation
Screen Shot 2022-03-09 at 00 18 44

pkg/scheduler/reconciliation/inmemory.go
adityabhatia: done

pkg/scheduler/reconciliation/inmemory.go
```diff
@@ -106,6 +106,22 @@ func (r *InMemoryReconciliationRepository) RemoveReconciliation(schedulingID str
 	return nil
 }
 
+func (r *InMemoryReconciliationRepository) RemoveReconciliations(schedulingIDs []string) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	for _, schedulingID := range schedulingIDs {
+		for _, operation := range r.operations[schedulingID] {
```
Tomasz-Smelcerz-SAP, line 114: What if there is no operations? Remember this is used for tests where we can expect unusual data patterns. I still consider delegating the job to the `RemoveReconciliation(schedulingID string)` function a better option. Unless I don't see something?

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -618,3 +647,21 @@ func (r *PersistentReconciliationRepository) GetAllComponents() ([]string, error
...
+	for i := 0; i < count; i += blockSize {
+		j += blockSize
+		if j > count {
+			j = count
+		}
+
+		blocks = append(blocks, slice[i:j])
+	}
+	return blocks
+}
```
jakobmoellersap, line 667: just a small thing here, feel free to leave as is if you think its more readable :) 
```suggestion
func chunkSlice(slice []string, chunkSize int) [][]string {
	if len(slice) == 0 {
		return nil
	}

	// initialize with final size for alloc
	chunks := make([][]string, (len(slice)+chunkSize-1)/chunkSize)

	i, left, limit := 0, 0, len(slice)-chunkSize

	for left < limit {
		right := left + chunkSize
		chunks[i], left = slice[left:right], right
		i++
	}
	chunks[i] = slice[left:]

	return chunks
}
```
```

File: pkg/scheduler/reconciliation/inmemory.go (hunk 3 above)
Comment (adityabhatia) [reply pkg/scheduler/reconciliation/inmemory.go]: As discussed, I have extracted the common code out to a separate re-use function.

File: pkg/scheduler/reconciliation/persistent.go [hunk 5]
```diff
@@ -174,6 +160,49 @@ func (r *PersistentReconciliationRepository) RemoveReconciliation(schedulingID s
...
+		for _, schedulingIDsBlock := range schedulingIDsBlocks {
+			var args []interface{}
+
+			// format scheduling IDs
+			sIdsDuplicate := make(map[string]interface{}, len(schedulingIDsBlock))
+
+			var buffer bytes.Buffer
+
+			for i, schedulingID := range schedulingIDsBlock {
+				if _, ok := sIdsDuplicate[schedulingID]; ok {
```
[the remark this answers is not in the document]
adityabhatia: I was following the conventional way of WhereIn(...) implementation followed in the repo. By going through the use-cases I found out that in this case duplicate scheduling IDs should not ideally be expected, so removed it.

pkg/scheduler/reconciliation/persistent.go
adityabhatia: Thanks for the suggestion. I have adapted my function to make it more readable and incorporated returned slice capacity. Inline calculations (like ceiling above), in my opinion are efficient, but not intuitive.

Tomasz-Smelcerz-SAP (dismissed):

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -618,3 +638,23 @@ func (r *PersistentReconciliationRepository) GetAllComponents() ([]string, error
 
 	return components, nil
 }
+
+func splitStringSlice(slice []string, blockSize int) [][]string {
+	sliceLength := len(slice)
+	if sliceLength == 0 {
+		return nil
+	}
+	subSlicesCount := (sliceLength + blockSize - 1) / blockSize
```
ruanxin, line 647: can we use `math.Ceil` here to avoid the possibility of overflow? since we add two integer

pkg/scheduler/reconciliation/persistent.go
```diff
@@ -174,6 +160,40 @@ func (r *PersistentReconciliationRepository) RemoveReconciliation(schedulingID s
 	return db.Transaction(r.Conn, dbOps, r.Logger)
 }
 
+func (r *PersistentReconciliationRepository) RemoveReconciliations(schedulingIDs []string) error {
```
ruanxin, line 163: Does it have to be in a transaction? I mean it's fine some of the delete batch failed, right? And put this bulk deletion in one transaction I believe it will have performance issue.

pkg/scheduler/reconciliation/persistent.go
adityabhatia: Added it to prevent any edge cases including integer overflow issue.

pkg/scheduler/reconciliation/persistent.go
adityabhatia: For performance I have some screenshots attached above. Seeing which we decided on splitting the whole batch in blocks (sub-slices) of 200. 
Transaction enhancements are WIP and will be a part of upcoming PRs.

pkg/scheduler/reconciliation/persistent.go
ruanxin: in this case, we can rewrite this to `subSlicesCount := (sliceLength - 1) / blockSize + 1` to prevent overflow.

adityabhatia: That could be another approach sure, added it.

ruanxin (approved):

PR #915, opened 2022-03-09
add label of global_accountId in reconciliation_status metric

lauraanddolay:
Implement task https://github.tools.sap/kyma/backlog/issues/2412  which needs to add `global_accountid` for distinguish the trial skr or real customer.


pkg/metrics/reconciliation_status.go
```diff
@@ -53,7 +53,7 @@ func (c *ReconciliationStatusCollector) OnClusterStateUpdate(state *cluster.Stat
 	}
 
 	c.reconciliationStatusGauge.
-		WithLabelValues(state.Cluster.RuntimeID, state.Cluster.Runtime.Name).
+		WithLabelValues(state.Cluster.RuntimeID, state.Cluster.Runtime.Name, state.Cluster.Metadata.GlobalAccountID, state.Cluster.Metadata.ServicePlanID).
```
ebensom, line 56: Shouldn't we use plan name instead of plan ID as it's much more user friendly?

lauraanddolay: Sounds great and I'll change to `state.Cluster.Metadata.ServicePlanName`.

ebensom (approved):

tobiscr (approved):

jeremyharisch (dismissed):

tobiscr (approved):

Issue #456, opened 2021-11-23
Istio reconciler support istio 1.12


strekm:
**Description**

Implement necessary changes to install / upgrade istio in 1.12 version. For this story we do not need multiple version support. 

- [ ] istioctl 1.12 is used for installation / upgrade
- [ ] istio running in 1.12 version
- [ ] sidecars running in 1.12 version
- [ ] reconciler bumped in kyma cli

**Reasons**

Keeping Istio up to date

**Attachments**

part of https://github.com/kyma-project/kyma/issues/12668

varbanv:
@dariusztutaj is this caused by the istio bump: https://github.com/kyma-incubator/reconciler/issues/468

mjakobczyk:
Due to `istioctl` bug in version 1.12 for now it is not possible to bump it to the latest version.
Opened issues:
- https://github.com/istio/istio/issues/36261
- https://github.com/istio/istio/issues/36258
In order to allow using older `istioctl` versions (still passing the general Istio requirements) the restrictions for binary version checking have been loosen (now also max one minor below comparing to target Istio version).

strekm:
so atm there are no changes needed as long as we use istioctl 1.11 to install istio 1.12

mjakobczyk:
During bump it turned out that there are two different problems.

The first problem is connected with `MutatingWebhookConfiguration`. Example log from the pipeline:
```log
2022-01-31T22:49:30.950Z	INFO	actions/performer.go:199	Patching istio-sidecar-injector MutatingWebhookConfiguration...
2022-01-31T22:49:31.104Z	DEBUG	service/runner.go:132	Runner: Reconcile action of 'istio-configuration' with version 'local' failed: Could not patch MutatingWebhookConfiguration: the server rejected our request due to an error in our request
```
It looks like API has changed a bit.

As @cnvergence mentioned in our case it would be sufficient to:
1. Install Istio with ` istioctl apply` command
2. Patch `istio-revision-tag-default` mutating webhook instead of `istio-sidecar-injector`

Links:
- https://docs.google.com/document/d/1P1mNlrJOs-KHVqcKTrFLrvCUiLgyZ-Vlf7ezaZNvD6g/edit#heading=h.xw1gqgyqs5b
- https://github.com/istio/istio/pull/33773

The second problem revolves around version mismatch and will be addressed in a separate issue available here: https://github.com/kyma-incubator/reconciler/issues/744

strekm:
[...]

PR #916, opened 2022-03-09
idempotently disable sidecar injection into kube-system

piotrkpc:
## Links
Related to: https://github.com/kyma-incubator/reconciler/issues/456

### DIFF
--- a/pkg/reconciler/instances/istio/actions/performer.go
+++ b/pkg/reconciler/instances/istio/actions/performer.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"reflect"
 	"time"
 
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/chart"
@@ -201,7 +202,7 @@ func (c *DefaultIstioPerformer) PatchMutatingWebhook(context context.Context, ku
 		if err != nil {
 			return err
 		}
-		err = c.addLabelSelectorToWebhook(whConf, webhookNameToChange, requiredLabelSelector)
+		err = c.setLabelSelectorToWebhook(whConf, webhookNameToChange, requiredLabelSelector)
 		if err != nil {
 			return err
 		}
@@ -219,15 +220,20 @@ func (c *DefaultIstioPerformer) PatchMutatingWebhook(context context.Context, ku
 	return nil
 }
 
-func (c *DefaultIstioPerformer) addLabelSelectorToWebhook(whConf *v1.MutatingWebhookConfiguration, webhookNameToChange string, requiredLabelSelector metav1.LabelSelectorRequirement) error {
+func (c *DefaultIstioPerformer) setLabelSelectorToWebhook(whConf *v1.MutatingWebhookConfiguration, webhookNameToChange string, requiredLabelSelector metav1.LabelSelectorRequirement) error {
 	var whFound bool
 	for i := range whConf.Webhooks {
 		if whConf.Webhooks[i].Name == webhookNameToChange {
 			whFound = true
-			whConf.Webhooks[i].NamespaceSelector.MatchExpressions = append(
-				whConf.Webhooks[i].NamespaceSelector.MatchExpressions,
-				requiredLabelSelector,
-			)
+			matchExpressions := whConf.Webhooks[i].NamespaceSelector.MatchExpressions
+			var hasRequiredLabel bool
+			for j := range matchExpressions {
+				hasRequiredLabel = reflect.DeepEqual(matchExpressions[j], requiredLabelSelector)
+			}
+			if !hasRequiredLabel {
+				matchExpressions = append(matchExpressions, requiredLabelSelector)
+				whConf.Webhooks[i].NamespaceSelector.MatchExpressions = matchExpressions
+			}
 		}
 	}
 	if !whFound {
--- a/pkg/reconciler/instances/istio/actions/performer.go
+++ b/pkg/reconciler/instances/istio/actions/performer.go
@@ -202,7 +202,7 @@ func (c *DefaultIstioPerformer) PatchMutatingWebhook(context context.Context, ku
 		if err != nil {
 			return err
 		}
-		err = c.setLabelSelectorToWebhook(whConf, webhookNameToChange, requiredLabelSelector)
+		err = c.addNamespaceSelectorIfNotPresent(whConf, webhookNameToChange, requiredLabelSelector)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


pkg/reconciler/instances/istio/actions/performer_test.go
```diff
@@ -292,6 +292,29 @@ func Test_DefaultIstioPerformer_PatchMutatingWebhook(t *testing.T) {
...
+		whConfName := "istio-revision-tag-default"
+		kubeClient := mocks.Client{}
+		mutatingWebhookConf := createIstioMutatingWebhookConf(whConfName)
+		clientset := fake.NewSimpleClientset(mutatingWebhookConf)
+		kubeClient.On("Clientset").Return(clientset, nil)
+		wrapper := NewDefaultIstioPerformer(nil, nil, nil)
+
+		// when
+		err := wrapper.PatchMutatingWebhook(context.TODO(), &kubeClient, log)
+		require.NoError(t, err)
```
mjakobczyk, line 307: I would add first `// then` clause here.

pkg/reconciler/instances/istio/actions/performer_test.go
```diff
@@ -292,6 +292,29 @@ func Test_DefaultIstioPerformer_PatchMutatingWebhook(t *testing.T) {
...
+		intermediateWhConf, err := clientset.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(context.TODO(), whConfName, metav1.GetOptions{})
+		require.NoError(t, err)
+		err = wrapper.PatchMutatingWebhook(context.TODO(), &kubeClient, log)
```
mjakobczyk, line 310: Above this statement I would add: `// and when`.

pkg/reconciler/instances/istio/actions/performer_test.go
```diff
@@ -292,6 +292,29 @@ func Test_DefaultIstioPerformer_PatchMutatingWebhook(t *testing.T) {
...
+		require.NoError(t, err)
```
mjakobczyk, line 311: Above this statement I would add: `// and then`.

pkg/reconciler/instances/istio/actions/performer_test.go
```diff
@@ -292,6 +292,29 @@ func Test_DefaultIstioPerformer_PatchMutatingWebhook(t *testing.T) {
...
+
+		// then
```
mjakobczyk, line 313: I would get rid of this line.

pkg/reconciler/instances/istio/actions/performer_test.go
mjakobczyk, line 307: From my point of view `when` section should contain at most one instruction. In some specific cases more. But any following assertion is part of `then` as we validate the state after performing an action.

What I really like is creating a transparent flow, sometimes combining many `when` and `then`. I think it is not a problem to add another `// and when` bound with `// and then` - it highly increases readability and underlines that there was once again a logic performed.

pkg/reconciler/instances/istio/actions/performer_test.go
```diff
@@ -292,6 +292,29 @@ func Test_DefaultIstioPerformer_PatchMutatingWebhook(t *testing.T) {
 		require.Contains(t, gotNew.Webhooks[0].NamespaceSelector.MatchExpressions, want)
 		require.NotContains(t, gotOld.Webhooks[0].NamespaceSelector.MatchExpressions, want)
 	})
+
+	t.Run("patching mutating webhook should be idempotent", func(t *testing.T) {
```
mjakobczyk, line 296: In general I'm OK with this description but I would rather call define more verbosely such as:
`should patch mutating webhook only once when patch was performed twice` or something similar.

pkg/reconciler/instances/istio/actions/performer.go
```diff
@@ -201,7 +202,7 @@ func (c *DefaultIstioPerformer) PatchMutatingWebhook(context context.Context, ku
 		if err != nil {
 			return err
 		}
-		err = c.addLabelSelectorToWebhook(whConf, webhookNameToChange, requiredLabelSelector)
+		err = c.setLabelSelectorToWebhook(whConf, webhookNameToChange, requiredLabelSelector)
```
mjakobczyk, line 205: Why did you change the function signature? I don't necessarily find `set` word in this context as appropriate. In OOP languages it would imply that there could be a hard assignment performed underneath which is not the case here. We in fact include one more label selector the list of matched expressions if it not present there. I know what you wanted to achieve - to stay idempotent we should not add it twice, so add would imply we mindlessly include it to the list. How about renaming it to: `addLabelSelectorfNotPresent`? Or maybe even: `addLabelSelectorfNotPresentin`? Reading it out loud with parameters names should suggest that we should validate the existence in webhooks so we don't need to include it in function name.

pkg/reconciler/instances/istio/actions/performer.go
```diff
@@ -219,15 +220,20 @@ func (c *DefaultIstioPerformer) PatchMutatingWebhook(context context.Context, ku
 	return nil
 }
 
-func (c *DefaultIstioPerformer) addLabelSelectorToWebhook(whConf *v1.MutatingWebhookConfiguration, webhookNameToChange string, requiredLabelSelector metav1.LabelSelectorRequirement) error {
+func (c *DefaultIstioPerformer) setLabelSelectorToWebhook(whConf *v1.MutatingWebhookConfiguration, webhookNameToChange string, requiredLabelSelector metav1.LabelSelectorRequirement) error {
 	var whFound bool
 	for i := range whConf.Webhooks {
 		if whConf.Webhooks[i].Name == webhookNameToChange {
 			whFound = true
```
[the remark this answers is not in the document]
piotrkpc: This is really good catch! Thanks.

pkg/reconciler/instances/istio/actions/performer.go
```diff
@@ -219,15 +220,20 @@ func (c *DefaultIstioPerformer) PatchMutatingWebhook(context context.Context, ku
...
-			whConf.Webhooks[i].NamespaceSelector.MatchExpressions = append(
-				whConf.Webhooks[i].NamespaceSelector.MatchExpressions,
-				requiredLabelSelector,
-			)
+			matchExpressions := whConf.Webhooks[i].NamespaceSelector.MatchExpressions
+			var hasRequiredLabel bool
+			for j := range matchExpressions {
+				hasRequiredLabel = reflect.DeepEqual(matchExpressions[j], requiredLabelSelector)
```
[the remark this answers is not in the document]
piotrkpc: fixed and new unit test added

pkg/reconciler/instances/istio/actions/performer_test.go
piotrkpc: isn't the convention to execute testee object with when? I'll add more explicit comment instead.

pkg/reconciler/instances/istio/actions/performer.go
piotrkpc: I've changed it to `addNamespaceSelectorIfNotPresent`

mjakobczyk (approved):

mjakobczyk (approved):

Issue #917, opened 2022-03-09
istio reconciler should run istioctl install instead istioctl upgrade on subsequent control loop


piotrkpc:
**Description**
Running `istioctl` upgrade is not an idempotent operation when there is nothing to upgrade. It actually changes mutation webhook causing problem in reconciliation.

**Expected result**
We should run `istioctl install` in subsequent control loop.
```
pkopec@piotr-nb ~ » kyma provision k3d --port 8080:80@loadbalancer                                                                                                                                            79 ↵
  Checking if port flags are valid
  Checking if k3d registry of previous kyma installation exists
  Checking if k3d cluster of previous kyma installation exists
? Do you want to remove the existing k3d cluster? Type [y/N]: y
  Deleted k3d registry of previous kyma installation
  Deleted k3d cluster of previous kyma installation
- k3d status verified
- Created k3d registry 'kyma-registry:5001'
- Created k3d cluster 'kyma'
pkopec@piotr-nb ~ » 
pkopec@piotr-nb ~ »                                                                                                                                                                                          130 ↵
pkopec@piotr-nb ~ » istioctl install --set profile=default -y                                                                                                                                                130 ↵
✔ Istio core installed                                                                                                                                                                                             
✔ Istiod installed                                                                                                                                                                                                 
✔ Ingress gateways installed                                                                                                                                                                                       
✔ Installation complete                                                                                                                                                                                            Making this installation the default for injection and validation.

Thank you for installing Istio 1.12.  Please take a few minutes to tell us about your install/upgrade experience!  https://forms.gle/FegQbc9UvePd4Z9z7
pkopec@piotr-nb ~ » istioctl install --set profile=default -y
✔ Istio core installed                                                                                                                                                                                             
✔ Istiod installed                                                                                                                                                                                                 
✔ Ingress gateways installed                                                                                                                                                                                       
✔ Installation complete                                                                                                                                                                                            Making this installation the default for injection and validation.

Thank you for installing Istio 1.12.  Please take a few minutes to tell us about your install/upgrade experience!  https://forms.gle/FegQbc9UvePd4Z9z7
pkopec@piotr-nb ~ » istioctl analyze                              
Info [IST0102] (Namespace default) The namespace is not enabled for Istio injection. Run 'kubectl label namespace default istio-injection=enabled' to enable it, or 'kubectl label namespace default istio-injection=disabled' to explicitly mark it as not needing injection.
[...]

PR #919, opened 2022-03-10
Replace istioctl upgrade usage from Istio reconciler with apply

mjakobczyk:
## Changes proposed

- replace usage of `istioctl upgrade` with `istioctl apply` in `DefaultCommander` implementation

## Links

See: https://github.com/kyma-incubator/reconciler/issues/917

### DIFF
--- a/pkg/reconciler/instances/istio/istioctl/commander.go
+++ b/pkg/reconciler/instances/istio/istioctl/commander.go
@@ -89,34 +89,7 @@ func (c *DefaultCommander) Install(istioOperator, kubeconfig string, logger *zap
 }
 
 func (c *DefaultCommander) Upgrade(istioOperator, kubeconfig string, logger *zap.SugaredLogger) error {
-
-	kubeconfigPath, kubeconfigCf, err := file.CreateTempFileWith(kubeconfig)
-	if err != nil {
-		return err
-	}
-
-	defer func() {
-		cleanupErr := kubeconfigCf()
-		if cleanupErr != nil {
-			logger.Error(cleanupErr)
-		}
-	}()
-
-	istioOperatorPath, istioOperatorCf, err := file.CreateTempFileWith(istioOperator)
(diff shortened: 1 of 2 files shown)


pkg/reconciler/instances/istio/istioctl/commander.go
```diff
@@ -89,34 +89,7 @@ func (c *DefaultCommander) Install(istioOperator, kubeconfig string, logger *zap
...
-		cleanupErr := istioOperatorCf()
-		if cleanupErr != nil {
-			logger.Error(cleanupErr)
-		}
-	}()
-
-	cmd := execCommand(c.istioctl.path, "upgrade", "-f", istioOperatorPath, "--kubeconfig", kubeconfigPath, "--skip-confirmation")
-
-	return c.execute(cmd, logger)
+	return c.Install(istioOperator, kubeconfig, logger)
```
strekm, line 92: would that work for istio 1.11?

strekm (changes requested):
i'm not quite sure if that method will also work for previous versions of istio, e.g. 1.11

pkg/reconciler/instances/istio/istioctl/commander.go
mjakobczyk: Yes, it would. Please [check my comment with testing results](https://github.com/kyma-incubator/reconciler/issues/917#issuecomment-1064007168).

mjakobczyk:
I verified this PR against Istio `1.11.4` and `1.12.3`. Results are available [here](https://github.com/kyma-incubator/reconciler/pull/919#discussion_r823696869).

strekm (approved):
/meow

piotrkpc:
I've verified this PR using test with our testing repo with label `area/istio`. Looks ok, both standalone installation and upgrade procedure from 2.0.4 -> main.

PR #929, opened 2022-03-11
Remove reconciliations by filter

Tomasz-Smelcerz-SAP:
**Description**

Add a "bulk delete"  feature to the reconciliations repository in order to optimize cleaner module performance.

The new function accepts a select-filter (or combination of such filters) to delete records in one shot using DELETE statement without loading the data into reconciler's process memory.
The direct use case for this feature is to remove all records older than certain date except a single most-recent one, even if it matches the "older than" condition - see: https://github.com/kyma-incubator/reconciler/issues/884

**Changes**
- extend reconciliation.Repository interface with `RemoveReconciliations(filter Filter) error` function
- provide necessary implementations for SQL and in-memory repositories.
- extend logging in TxConnection.Exec() function to debug-level logging of transactional SQL statements like "DELETE"
- improve cleaner deletion logic with the new feature

**See also**
https://github.com/kyma-incubator/reconciler/pull/912

### DIFF
--- a/pkg/db/transaction.go
+++ b/pkg/db/transaction.go
@@ -3,13 +3,14 @@ package db
 import (
 	"database/sql"
 	"fmt"
-	"github.com/google/uuid"
-	"github.com/pkg/errors"
-	"go.uber.org/zap"
 	"math/rand"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
 )
 
 const txMaxRetries = 5
@@ -153,6 +154,7 @@ func (t *TxConnection) Query(query string, args ...interface{}) (DataRows, error
 }
 
 func (t *TxConnection) Exec(query string, args ...interface{}) (sql.Result, error) {
+	t.logger.Debugf("Transaction Exec(): %s | %v", query, args)
 	return t.tx.Exec(query, args...)
 }
 
--- a/pkg/scheduler/reconciliation/filter.go
+++ b/pkg/scheduler/reconciliation/filter.go
@@ -157,6 +157,27 @@ func (ws *WithSchedulingID) FilterByInstance(i *model.ReconciliationEntity) *mod
 	return nil
 }
 
+type WithNotSchedulingID struct {
+	SchedulingID string
+}
+
+func (wns *WithNotSchedulingID) FilterByQuery(q *db.Select) error {
+	column, err := columnName(q, "SchedulingID")
+	if err != nil {
+		return err
+	}
+
+	q.WhereRaw(fmt.Sprintf("%s<>$%d", column, q.NextPlaceholderCount()), wns.SchedulingID)
+	return nil
+}
+
+func (wns *WithNotSchedulingID) FilterByInstance(i *model.ReconciliationEntity) *model.ReconciliationEntity {
+	if i.SchedulingID != wns.SchedulingID {
+		return i
+	}
+	return nil
+}
+
 type WithRuntimeIDs struct {
 	RuntimeIDs []string
 }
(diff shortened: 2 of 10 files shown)


Tomasz-Smelcerz-SAP:
**Important**
TLDR: Be aware that local database performance does NOT reflect the actual, "production-ready" instance.

This PR introduces a DELETE statement that looks like that:
```
DELETE FROM scheduler_reconciliations sr WHERE sr.scheduling_id in (
    SELECT scheduling_id FROM scheduler_reconciliations as tab_sel where
    (created<'2022-03-20 00:00:00.000') and
    runtime_id='f21ab551-14c8-44ce-a90b-31ccba3e91f3' and
    (scheduling_id<>'d671505f-b64d-43bd-bdad-aff053b3da7a--fc589022-ee71-4b1f-98eb-6709f0745d6b') 
    ORDER BY created DESC   
)
```
It's not optimal, some indexes are missing, but it performs well enough for the time.
Because of a foreign key constraint in the `scheduler_operations` table, every record deleted from `scheduler_reconciliations` table results in additional deleted recrods from the `scheduler_operations` table (cascade delete)
The local postgresql in the docker container performs VERY badly in such scenario. The delete takes a couple minutes for 25k entries in `scheduler_reconciliations` and 500k corresponding entries in the `scheduler_operations` table.
But the real database handles the load much better.

The actual results from DEV environment:
`select count(*) from scheduler_reconciliations => 5766`
`select count(*) from scheduler_operations => 121377`
Execution time for delete statement that removes all the records from both tables: 1395 ms

pkg/db/transaction.go
```diff
@@ -153,6 +154,7 @@ func (t *TxConnection) Query(query string, args ...interface{}) (DataRows, error
 }
 
 func (t *TxConnection) Exec(query string, args ...interface{}) (sql.Result, error) {
+	t.logger.Debugf("Transaction Exec(): %s | %v", query, args)
```
jakobmoellersap, line 157: Do we want to add this also for the other statements or as opt-in on the calling service layer (e.g. Repository)? I feel like only having this on exec seems a bit arbitrary

pkg/scheduler/reconciliation/filter.go
```diff
@@ -157,6 +157,27 @@ func (ws *WithSchedulingID) FilterByInstance(i *model.ReconciliationEntity) *mod
...
+	SchedulingID string
+}
+
+func (wns *WithNotSchedulingID) FilterByQuery(q *db.Select) error {
+	column, err := columnName(q, "SchedulingID")
+	if err != nil {
+		return err
+	}
+
+	q.WhereRaw(fmt.Sprintf("%s<>$%d", column, q.NextPlaceholderCount()), wns.SchedulingID)
```
jakobmoellersap, line 170: I think this is not strictly necessary, but could we make use of the existing positional argument support instead of building the entire query on demand? We still make use of the SQL Injection Regex so I dont think its a big problem but I think here we could use it.

pkg/scheduler/reconciliation/reconciliation.go
```diff
@@ -22,6 +23,7 @@ type Repository interface {
 	RemoveReconciliationByRuntimeID(runtimeID string) error
 	RemoveReconciliationBySchedulingID(schedulingID string) error
 	RemoveReconciliationsBySchedulingID(schedulingIDs []string) error
+	RemoveReconciliations(filter Filter) error
```
jakobmoellersap, line 26:
```suggestion
	RemoveReconciliationsByFilter(filter Filter) error
```
Not a must-have but would align with the rest

jakobmoellersap (approved):
Just smaller comments, but Implementation is LGTM :) Feel free to do with them as you like

pkg/scheduler/reconciliation/filter.go
adityabhatia: That would have been good, but there is no support for an operator other than "=" in Where() at the moment.

pkg/db/transaction.go
adityabhatia: In my opinion, this was used for debugging and can be removed in a follow-up PR.

Issue #902, opened 2022-03-07
Setup Prometheus monitoring for component reconcilers


tobiscr:
**Description**

Monitoring of component reconcilers is mandatory for establishing operational readiness. Currently, the monitoring is limited to a few figures and cannot be extended in an easy way. 

To achieve an extendable and future-proof monitoring approach, a pure Prometheus based solution has to be establishe.

AC:

* Component reconcilers have to export monitoring values by providing Prometheus scrape endpoints
* Prometheus is already running in each KCP cluster and has to be configured:
     *  Define [ServiceMonitor](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md#related-resources) CRs to scrape component-reconciler metrics
     * Custom [PrometheusRules](https://github.com/prometheus-operator/prometheus-operator#customresourcedefinitions) have to be defined to provide metrics (e.g. aggregated avg-values calculated on base of all component-reconciler metrics)
* the custom implementation for reporting metrics from component-reconcilers to the mothership-reconciler (e.g. occupancy tracking, component processing KPIs etc.) has to be removed: the mothership reconciler should not be used for any calculation of metrics based on DB queries. Such information has to be either directly reported to or calculated by the Prometheus server.

- [x] #903

- [x] #907
- [x] #930 
- [x] add grafana dashboard to monitor exposed metrics

**Reasons**

Establish standardised monitoring approach for reconciler (mothership + component reconciler).

**Attachments**

janmedrek:
Functionalities of the Reconciler were migrated to the new architecture, this issue will not be worked on.

PR #930, opened 2022-03-11
add ComponentProcessingDurationMetric to component reconciler

ruanxin:
**Description**

sub issue of #902

Expose the reconcile processing duration time of each component reconciler as Histogram metrics.

### DIFF
--- a/cmd/reconciler/start/service/reconciler.go
+++ b/cmd/reconciler/start/service/reconciler.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"context"
+	"github.com/kyma-incubator/reconciler/pkg/metrics"
+	"github.com/prometheus/client_golang/prometheus"
 
 	reconCli "github.com/kyma-incubator/reconciler/internal/cli/reconciler"
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/service"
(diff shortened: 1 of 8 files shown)


cmd/reconciler/start/service/reconciler.go
```diff
@@ -11,7 +13,14 @@ func StartComponentReconciler(ctx context.Context, o *reconCli.Options, reconcil
...
 	}
-	recon, err := reconCli.NewComponentReconciler(o, reconcilerName)
+
+	durationMetric := metrics.NewComponentProcessingDurationMetric(o.Logger())
+	err := prometheus.Register(durationMetric.Collector)
+	if err != nil {
+		return nil, nil, err
+	}
+	reconcilerMetricsSet := metrics.NewReconcilerMetricsSet(durationMetric)
+	recon, err := reconCli.NewComponentReconciler(o, reconcilerName, reconcilerMetricsSet)
```
tobiscr, line 23: As we are always enabling this metric and no config-values are needed, can't we directly create the MetricSet inside of the `ComponentReconciler` struct (or even in the `NewInstall` fct)?

tobiscr (changes requested):

cmd/reconciler/start/service/reconciler.go
ruanxin: The MetricSet only need when componentReconciler started in web service mode, for other usage, it's not necessary to have. like in `pkg/reconciler/instances/busolamigrator/busola.go`, it's not for web service mode, right? Or am I understood wrong...
```
func init() {
	log := logger.NewLogger(false)

	log.Debugf("Initializing component reconciler '%s'", ReconcilerName)
	reconciler, err := service.NewComponentReconciler(ReconcilerName)
	if err != nil {
		log.Fatalf("Could not create '%s' component reconciler: %s", ReconcilerName, err)
	}
```

tobiscr (approved):

PR #931, opened 2022-03-14
occupancy tracking local autoscaling testing findings

khlifi411:


pkg/scheduler/worker/workerpool.go
```diff
@@ -238,3 +244,33 @@ func (w *Pool) RunningWorkers() (int, error) {
...
+		w.logger.Debug("Worker pool has nil occupancy observer already")
+		return
+	}
+	w.occupancyObserver = nil
+	w.logger.Infof("Worker pool unregistered its occupancy observer: %v", observer)
+}
+
+func (w *Pool) Notify() error {
+	if w.occupancyObserver == nil {
+		return fmt.Errorf("worker pool failed to notify occupancy observer: %v", w.occupancyObserver)
```
jeremyharisch, line 268: This Error message is a bit misleading, since the error is not that the notify failed but instead that no observer is registered

pkg/scheduler/worker/workerpool.go
```diff
@@ -238,3 +244,33 @@ func (w *Pool) RunningWorkers() (int, error) {
...
+	}
+	err := w.occupancyObserver.UpdateOccupancy()
+	if err != nil {
+		return err
```
jeremyharisch, line 272: Maybe wrap error, with an additional message that the notify failed, just to have a clearer understanding where from the code the error comes from

pkg/scheduler/worker/workerpool.go
khlifi411: Thanks! I have changed the error message to `no observer was registered`

pkg/scheduler/worker/workerpool.go
khlifi411: Here, the error will be displayed after the message `worker pool failed to notify occupancy observer:` in the Debug log in line 182, so that we know from which component this error message is coming.

jeremyharisch (approved):

PR #934, opened 2022-03-14
Remove reconciliations by runtimeID

adityabhatia:
- Adds functionality to remove reconciliations by runtimeID
- Rename previous removeReconciliation() -> removeReconciliationsBySchedulingID(), removeReconciliations() -> removeReconciliationsBySchedulingID()
- Test removeReconciliationsBySchedulingID() and removeReconciliationsBySchedulingID()

PR #940, opened 2022-03-15
Add purge functionality and API endpoint

Cortey:
**Description**
Created an API Endpoint to handle `kcp reconciliations purge` requests
See also:
https://github.com/kyma-project/control-plane/pull/1453
https://github.com/kyma-incubator/reconciler/pull/934
https://github.com/kyma-project/control-plane/issues/1365


cmd/mothership/mothership/start/http.go
```diff
@@ -174,6 +184,46 @@ func callHandler(o *Options, handler func(o *Options, w http.ResponseWriter, r *
 	}
 }
 
+func deleteReconcilationsByCluster(o *Options, w http.ResponseWriter, r *http.Request) {
+	params := server.NewParams(r)
+	_, err := params.Int64(paramContractVersion)
```
adityabhatia, line 189: optional: you could use inline 'if' conditions if returned values scope is not required outside.
e.g. if _, err := params.Int64(paramContractVersion); err != nil {...}

adityabhatia (approved):

PR #953, opened 2022-03-21
Use semicolon instead of colon to seperate paths to support windows 

jeremyharisch:
### Changes proposed

- Use `;` to separate paths instead of `:`. Since in Windows `:` can be used inside paths, i.e.  `C:\Users\`, which lead to fail the Kyma CLI on Windows Systems. Without this change Kyma CLI is not able to work on Windows distributions
- If `istioctl` is non executable, try to set tit to executable once and if it fails, let the component fail, instead of instant failure
- Propagate logger to have better debugging

### Links
See: https://github.com/kyma-incubator/reconciler/issues/962

### DIFF
--- a/Dockerfile.cr
+++ b/Dockerfile.cr
@@ -39,7 +39,7 @@ COPY --from=build /configs/ /configs/
 # Add istioctl tools
 COPY --from=istio-1_11_4 /usr/local/bin/istioctl /bin/istioctl-1.11.4
 COPY --from=istio-1_12_3 /usr/local/bin/istioctl /bin/istioctl-1.12.3
-# For multiple istioctl binaries, provide their paths separated with a colon (:) like in the Linux PATH variable.
+# For multiple istioctl binaries, provide their paths separated with a semicolon (;) like in the Linux PATH variable.
 ENV ISTIOCTL_PATH=/bin/istioctl-1.12.3;/bin/istioctl-1.11.4
 
 USER appuser:appuser
--- a/Dockerfile.cr
+++ b/Dockerfile.cr
@@ -40,7 +40,7 @@ COPY --from=build /configs/ /configs/
 COPY --from=istio-1_11_4 /usr/local/bin/istioctl /bin/istioctl-1.11.4
 COPY --from=istio-1_12_3 /usr/local/bin/istioctl /bin/istioctl-1.12.3
 # For multiple istioctl binaries, provide their paths separated with a colon (:) like in the Linux PATH variable.
-ENV ISTIOCTL_PATH=/bin/istioctl-1.12.3:/bin/istioctl-1.11.4
+ENV ISTIOCTL_PATH=/bin/istioctl-1.12.3;/bin/istioctl-1.11.4
 
 USER appuser:appuser
 
--- a/pkg/reconciler/instances/istio/bootstrap.go
+++ b/pkg/reconciler/instances/istio/bootstrap.go
@@ -32,7 +32,7 @@ func istioPerformerCreator(istioProxyReset proxy.IstioProxyReset, provider clien
 
 		resolver, err := newDefaultCommanderResolver(istioctlPaths, logger)
 		if err != nil {
-			logger.Errorf("Could not create '%s' component reconciler: Error parsing env variable '%s': %s", name, istioctlBinaryPathEnvKey, err.Error())
+			logger.Errorf("Could not create '%s' component reconciler: Error creating DefaultCOmmadnerResolver with istioctLPaths '%s': %s", name, istioctlPaths, err.Error())
 			return nil, err
 		}
 
@@ -75,7 +75,7 @@ func newDefaultCommanderResolver(paths []string, log *zap.SugaredLogger) (action
 	}, nil
 }
 
-// parsePaths func parses and validates executable paths. The input must contain a list of full/absolute filesystem paths of binaries, separated by a colon character ':'
+// parsePaths func parses and validates executable paths. The input must contain a list of full/absolute filesystem paths of binaries, separated by a semicolon character ';'
 // isValid function is used to validate every single binary path in the input.
 func parsePaths(input string, isValid func(string) error) ([]string, error) {
 	trimmed := strings.TrimSpace(input)
@@ -85,7 +85,7 @@ func parsePaths(input string, isValid func(string) error) ([]string, error) {
 	if len(trimmed) > istioctlBinaryPathMaxLen {
 		return nil, errors.New(fmt.Sprintf("%s env variable exceeds the maximum istio path limit of %d characters", istioctlBinaryPathEnvKey, istioctlBinaryPathMaxLen))
 	}
-	pathDefs := strings.Split(trimmed, ":")
+	pathDefs := strings.Split(trimmed, ";")
 	var res []string
 	for _, path := range pathDefs {
 		val := strings.TrimSpace(path)
--- a/pkg/reconciler/instances/istio/bootstrap.go
+++ b/pkg/reconciler/instances/istio/bootstrap.go
@@ -24,9 +24,14 @@ func istioPerformerCreator(istioProxyReset proxy.IstioProxyReset, provider clien
 
 	res := func(logger *zap.SugaredLogger) (actions.IstioPerformer, error) {
 		pathsConfig := os.Getenv(istioctlBinaryPathEnvKey)
-		istioctlPaths, err := parsePaths(pathsConfig, ensureFileExecutable, logger)
+		istioctlPaths, err := parsePaths(pathsConfig)
 		if err != nil {
 			logger.Errorf("Could not create '%s' component reconciler: Error parsing env variable '%s': %s", name, istioctlBinaryPathEnvKey, err.Error())
+			return nil, err
+		}
+		err = ensureFilesExecutable(istioctlPaths, logger)
+		if err != nil {
+			logger.Errorf("Could not create '%s' component reconciler: One or more istioctl binaries are not executable '%s': %s", name, istioctlBinaryPathEnvKey, err.Error())
 			return nil, err
 		}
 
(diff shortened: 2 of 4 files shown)


pkg/reconciler/instances/istio/bootstrap.go
```diff
@@ -24,15 +24,15 @@ func istioPerformerCreator(istioProxyReset proxy.IstioProxyReset, provider clien
...
+		istioctlPaths, err := parsePaths(pathsConfig, validatePath, logger)
 		if err != nil {
 			logger.Errorf("Could not create '%s' component reconciler: Error parsing env variable '%s': %s", name, istioctlBinaryPathEnvKey, err.Error())
 			return nil, err
 		}
 
 		resolver, err := newDefaultCommanderResolver(istioctlPaths, logger)
 		if err != nil {
-			logger.Errorf("Could not create '%s' component reconciler: Error parsing env variable '%s': %s", name, istioctlBinaryPathEnvKey, err.Error())
+			logger.Errorf("Could not create '%s' component reconciler: Error creating DefaultCommadnerResolver with istioctLPaths '%s': %s", name, istioctlPaths, err.Error())
```
mjakobczyk, line 35:
```suggestion
			logger.Errorf("Could not create '%s' component reconciler: Error creating DefaultCommanderResolver with istioctlPaths '%s': %s", name, istioctlPaths, err.Error())
```

pkg/reconciler/instances/istio/bootstrap.go
```diff
@@ -101,17 +101,31 @@ func parsePaths(input string, isValid func(string) error) ([]string, error) {
...
 	}
 	mode := stat.Mode()
+	logger.Debugf("%s mode: %s", path, mode)
 	if (!mode.IsRegular()) || mode.IsDir() {
 		return errors.New(fmt.Sprintf("\"%s\" is not a regular file", path))
 	}
 	if uint32(mode&0111) == 0 {
-		return errors.New(fmt.Sprintf("\"%s\" is not executable", path))
+		logger.Debugf("%s is not executable, will chmod +x", path)
+		var fileMode os.FileMode = 0777
```
mjakobczyk, line 116: After you really extended the logic of the function could you provide some tests? We lack them here

pkg/reconciler/instances/istio/bootstrap.go
```diff
@@ -101,17 +101,31 @@ func parsePaths(input string, isValid func(string) error) ([]string, error) {
...
+		if err := os.Chmod(path, fileMode); err != nil {
```
mjakobczyk, line 117: I believe this logic should not be a part of this function. It was designed to check whether path exists and is executable

mjakobczyk (changes requested):

pkg/reconciler/instances/rma/integration_action.go
```diff
@@ -37,7 +37,6 @@ type IntegrationAction struct {
 	client       IntegrationClient
 	mux          sync.Mutex
 	archives     map[string][]byte
-	pwds         map[string]string
```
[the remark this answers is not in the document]
jeremyharisch: This change came from rebasing main into my branch, but since it does not gets used anywhere it can be removed.

pkg/reconciler/instances/istio/bootstrap.go
jeremyharisch: I can create a function which will then be called there. We need it here, since to persist file permission in Windows is very weirdly behaving. I tested it the full day yesterday on a local Windows VM , and that is the only way I got it to work . But yes, I will move it into a separate function.

pkg/reconciler/instances/istio/bootstrap.go
jeremyharisch: Will do so

pkg/reconciler/instances/istio/bootstrap.go
jeremyharisch: done

pkg/reconciler/instances/istio/bootstrap_test.go
```diff
@@ -113,9 +115,68 @@ func TestParsePaths(t *testing.T) {
...
+	})
+	t.Run("chmodExecutable should return an error if pathToFile is empty", func(t *testing.T) {
+		pathToFile := ""
+		err := chmodExecutbale(pathToFile, zap.NewNop().Sugar())
+		require.Error(t, err)
+	})
+}
+
+func TestValidatePath(t *testing.T) {
+	t.Run("validatePath should return nil and change mode of file to 0777", func(t *testing.T) {
```
mjakobczyk, line 155: You don't need to repeat the name of the tested function as this one is already in the outer testing function

pkg/reconciler/instances/istio/bootstrap_test.go
```diff
@@ -113,9 +115,68 @@ func TestParsePaths(t *testing.T) {
...
+		err := os.WriteFile(pathToFile, d1, 0000)
+		require.NoError(t, err)
+		err = validatePath(pathToFile, zap.NewNop().Sugar())
+		require.NoError(t, err)
+		stat, err := os.Stat(pathToFile)
+		require.NoError(t, err)
+		require.Equal(t, os.FileMode(0777), stat.Mode())
+
+	})
+	t.Run("validatePath should return error if file is not existing", func(t *testing.T) {
```
mjakobczyk, line 172:
```suggestion
	t.Run("should return error when file does not exist", func(t *testing.T) {
```

pkg/reconciler/instances/istio/bootstrap_test.go
```diff
@@ -113,9 +115,68 @@ func TestParsePaths(t *testing.T) {
...
+		pathToFile := "not-existing"
+		err := validatePath(pathToFile, zap.NewNop().Sugar())
+		require.Error(t, err)
+	})
+	t.Run("validatePath should return error if path is empty", func(t *testing.T) {
```
mjakobczyk, line 177:
```suggestion
	t.Run("should return error when path is empty", func(t *testing.T) {
```

pkg/reconciler/instances/istio/bootstrap_test.go
```diff
@@ -113,9 +115,68 @@ func TestParsePaths(t *testing.T) {
...
+		stat, err := os.Stat(pathToFile)
+		require.NoError(t, err)
+		require.Equal(t, os.FileMode(0777), stat.Mode())
+	})
+	t.Run("chmodExecutable should return an error if file is not existing", func(t *testing.T) {
+		pathToFile := "not-existing"
+		err := chmodExecutbale(pathToFile, zap.NewNop().Sugar())
+		require.Error(t, err)
+	})
+	t.Run("chmodExecutable should return an error if pathToFile is empty", func(t *testing.T) {
```
mjakobczyk, line 147:
```suggestion
	t.Run("should return an error when path to file is empty", func(t *testing.T) {
```

pkg/reconciler/instances/istio/bootstrap.go
```diff
@@ -75,44 +75,64 @@ func newDefaultCommanderResolver(paths []string, log *zap.SugaredLogger) (action
...
 			return nil, err
 		}
 		res = append(res, val)
 	}
 	return res, nil
 }
 
-// TODO: test it
-func validatePath(path string) error {
+func validatePath(path string, logger *zap.SugaredLogger) error {
```
mjakobczyk, line 103: The contract for this function before was to return an error if file was corrupted or not executable. If you want to change it then please adjust the name appropriately as well so it reflects its real purpose. I would suggest: `ensureExecutable` (we don't need to repeat `path` imho as it is already a function parameter) or `ensureFileExecutable` or even `ensureFileExecutableFrom` (implicitly from path when you keep reading function parameters too)

mjakobczyk:
@jeremyharisch we should correct every place in which we define `ISTIOCTL_PATH` env. From what I have in mind these will be:
- [Dockerfile in Reconciler](https://github.com/kyma-incubator/reconciler/blob/fb4311702b3582d2a2324b6a60d8869a925481c1/Dockerfile.cr#L43)
- control-plane - I don't know exactly where, it should be checked out

mjakobczyk (approved):
After all the adjustments it looks fine 💯

Dockerfile.cr
```diff
@@ -40,7 +40,7 @@ COPY --from=build /configs/ /configs/
 COPY --from=istio-1_11_4 /usr/local/bin/istioctl /bin/istioctl-1.11.4
 COPY --from=istio-1_12_3 /usr/local/bin/istioctl /bin/istioctl-1.12.3
 # For multiple istioctl binaries, provide their paths separated with a colon (:) like in the Linux PATH variable.
```
ruanxin, line 42: - Please remove this comment if changed to `;` as separator.
- Is this ENV value only used by that `parsePaths` function? If it is, then it doesn't have to change, right? If it's for compatible with windows path, then will this change break linux path recognition?

pkg/reconciler/instances/istio/bootstrap.go
```diff
@@ -24,15 +24,15 @@ func istioPerformerCreator(istioProxyReset proxy.IstioProxyReset, provider clien
 
 	res := func(logger *zap.SugaredLogger) (actions.IstioPerformer, error) {
 		pathsConfig := os.Getenv(istioctlBinaryPathEnvKey)
-		istioctlPaths, err := parsePaths(pathsConfig, validatePath)
+		istioctlPaths, err := parsePaths(pathsConfig, ensureFileExecutable, logger)
```
ruanxin, line 27: I suggest not mix the `parsePaths` and `ensureFileExecutable` together to follow single responsibility. We can iterate the returned `istioctlPaths` to check `ensureFileExecutable`, then we don't need to pass `isValid func(string) error` as an function to `parsePaths`, and it's also easier to write test

pkg/reconciler/instances/istio/bootstrap_test.go
```diff
@@ -9,23 +11,23 @@ import (
 )
 
 func TestParsePaths(t *testing.T) {
-	alwaysValidFn := func(path string) error { return nil }
+	alwaysValidFn := func(path string, logger *zap.SugaredLogger) error { return nil }
```
ruanxin, line 14: If we extract valid check out of `parsePaths`, we don't need this `alwaysValidFn`

Dockerfile.cr
jeremyharisch: I will change the comment. It does not break linux, it was just for staying consistent with linux env var guidelines, but it will only be used with parsePath function. In the CLi scenario, this env var will be set by Kyma CLI, but since Windows Default path is something like `C:\<...>`, we need to change to a symbol which is also not allowed in Windows paths

ruanxin (approved):

mjakobczyk (approved):

adityabhatia (approved):

PR #955, opened 2022-03-21
Safe cleanup of service catalog instances

wozniakjan:
Followup to https://github.com/kyma-incubator/reconciler/pull/875

As part of service catalog migration and cleanup, it's desired to actually delete the instances and bindings but this PR introduces safer deletion algorithm
* check if service manager knows about the instance
  * if it does - instance must be migrated
  * if it doesn't - halt and error if the instance is ready and has a broker with prefix `sm-`


szwedm (approved):
/lgtm

szwedm (approved):
/lgtm

Issue #781, opened 2022-02-04
Redesign skipped unit test


ruanxin:

jakobmoellersap:
We concluded that we will permanently skip/remove this test in favor of more stable pipeline tests with happy path

jakobmoellersap:
Before TestMothership, we will have to introduce https://github.com/kyma-incubator/reconciler/issues/1008 and then move the E2E Test Execution into a new Profile

Sawthis:
@jakobmoellersap what is status of this issue? According to our rule it should be marked as release blocker because it was opened more than 14 days ago.

jakobmoellersap:
This is not a failing test and it also isnt failing, its just taking so much time to execute that we took it out of our daily pipeline. I think it should not be a release blocker since all of the e2e functionality is tested in much more extensive pipelines as well. In the team we agreed to permanently leave this out because of this.

PR #956, opened 2022-03-21
Rework E2E Integration Test Suite for Mothership and Component Reconciler

### LINKED ISSUE #150: Enhance mothership reconciler integration test (add component reconciler mock)
**Description**

The existing integration tests of the mothership reconcilers have to be extended by a component-reconciler mock to verify the correct behaviour of the mothership reconciler in following cases:

- [x]  Happy path: mothership calls component reconcilers and gets successful result
- [x]  At least one component reconcilers returns consistently an error
- [ ]  One or more component reconcilers are not reachable via network (target endpoint down)
- [ ]  A component reconcilers returns an invalid response (e.g. 500 server error)
- [x] A component reconciler is overloaded (pool, filled 429 Too Many Requests)

**Reasons**

Stabile mothership reconciler by automated testing of expected use cases.

**Attachments**

### LINKED ISSUE #790: Component reconciler should return 429 HTTP error(too many requests) if worker pool is fully allocated
**Description**

A component reconciler should explicitly reject incoming requests with a 429 HTTP response if its worker pool is full allocated.

The response payload should indicate which component-reconciler was sending the response (e.g. pod name) and what the configured worker-pool capacity is.

AC:
* If all workers in the component reconciler's worker pool are busy, a 429 HTTP response (too many requests) should be replied to indicate that the component reconciler is overloaded
* The HTTP payload should include the name of the component reconciler pod and the current worker pool capacity
* An integration test is implemented which verifies that the 429 HTTP response is[ properly handled by the mothership reconciler ](https://github.com/kyma-incubator/reconciler/blob/main/pkg/scheduler/invoker/remoteinvoker.go#L73-L81)and triggers another HTTP call (if this also fails, the operation has to be marked to be failed an re-scheduled later) -> requires #150 

**Reasons**

Send error response if a component reconciler is overloaded.

**Attachments**

jakobmoellersap:
fixes #790
fixes #150 
partial fix for #781

Also reworks the E2E Integration Test Suite and makes sure we spot several cleanup mistakes and race conditions that could happen during test execution and also during pool submission of reconciliations.

Main features:

- Creates a much more flexible Integration Test Case that is able to do better response verification and expectation setting and can also test parallel execution scenarios

- Spots and resolves 500 Internal Server Error from Component Reconcilers that can occur under heavy parallel submission load to a component reconciler as the Pool can be overloaded if not checked properly. This is tested by the 429 Rate Limiting Test with heavy parallelization.

- Simplifies the Suite Execution by isolating the database (using the database integration test suite) and also uses testify to manage logging behavior, allowing us for timestamped test logs, making it easier to spot hickups in timeouts or goroutines that hang up or get orphaned.

- Removes orphaned downloaded external Archive when executing Integration Tests

- Fixes Regression issue in Container suite that caused the container to not be deleted correctly

- Introduces Ryuk from Testcontainers to delete orphaned postgres instances if tests panic (failure is handled by teardown logic)

- Fixes an issue for the test container suite where one package in multiple directory subfolders leads to non-shared container test suites, this has been replaced with a directory hash.

- Reintroduces Mothership E2E Test

- Optimizes Integration Test Suite by not having to specify alternative relative paths for different folders using the DB Test Suite, making consumption much easier

- Including a new mothership CLI Option to define the Bookkeeper Interval instead of hardcoding at 45 seconds

- Various smaller fixes to make sure the mothership / component reconciler CLI Options are properly handled

The cleanup also makes our current deletion appear much more flawed than originally anticipated. It turns out that our kube Adapter still makes use of Helm for deletion of resources and will thus suffer from unremoved finalizers blocking namespace deletion. The only way I got this to work was to separate the test using CRDs into a different namespace.


cmd/reconciler/start/service/component_integration_test.go
```diff
@@ -0,0 +1,414 @@
...
+func (a *NoOpReconcileAction) Run(context *service.ActionContext) (err error) {
+	// prepare logger
+	contextLogger := eventLog.ContextLogger(context, eventLog.WithAction("no-op-action"))
+	contextLogger.Infof("Waiting to simulate Op...")
+	time.Sleep(a.WaitTime)
+	return nil
+}
+
+func (s *reconcilerIntegrationTestSuite) setDefaultValuesFromTestCase(testCase *reconcilerIntegrationTestCase) {
+	if testCase.model.Namespace == "" && testCase.settings.namespace != "" {
```
ruanxin, line 34: Why we do overwrite from settings only when original value is empty? Why not allow force overwrite?

cmd/reconciler/start/service/component_integration_suite_test.go
```diff
@@ -0,0 +1,313 @@
...
+			settings.url,
+		)
+
+		// this is done as Cleanup of the externally downloaded component
+		if settings.deleteAfterTest {
+			s.NoError(s.chartFactory.Delete(version))
+		}
+	})
+
+	componentReconcilerServerContext, cancel := context.WithCancel(s.testContext)
```
ruanxin, line 194: I think use `context.WithTimeout(parent Context, timeout time.Duration)` will do the same purpose and avoid write time.Sleep in the `s.T().Cleanup()`

cmd/reconciler/start/service/component_integration_test.go
jakobmoellersap: I just did this for more transparent behavior, I will add this.

pkg/db/postgres_container_runtime.go
```diff
@@ -17,46 +19,64 @@ type PostgresContainerRuntime struct {
...
+	EncryptionKeyFile EncryptionKeyFileConfig
+}
+
+//id calculates a hash for ContainerSettings based on the name and image of a directory as well as a cumulated hash
+//over all files relevant for a migration
+func (p PostgresContainerSettings) id() string {
+	configHash, _ := file.HashDir(
+		string(p.migrationConfig()),
+		p.containerName()+p.containerImage(),
+		file.HashFnv(".sql"),
```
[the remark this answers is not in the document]
ruanxin: ok, I think here we support the scenario to test using self defined migration files.

ruanxin (dismissed):

cmd/reconciler/start/service/component_integration_suite_test.go
jakobmoellersap: This does not work as WithTimeout will start the timer when creating the context. What we want instead is starting the timeout after cancel was called. I tried this and you can see that the server immediately shuts down at https://status.build.kyma-project.io/view/gs/kyma-prow-logs/pr-logs/pull/kyma-incubator_reconciler/956/pre-main-reconciler-integration-k3d/1510998581635452928#1:build-log.txt%3A643-654

ruanxin: aha, sorry, then I understood in wrong direction..

ruanxin (approved):

jakobmoellersap:
TW review needed on https://github.com/kyma-incubator/reconciler/pull/956/files#diff-8af83c29df95880b0e5ccbb86a050a6a105e681609307dccaffaab31ef92117d

cmd/reconciler/start/service/http.go
```diff
@@ -120,9 +123,24 @@ func reconcile(ctx context.Context, w http.ResponseWriter, req *http.Request, o
 		return
 	}
 
+	// this mutex is necessary because if we have heavy parallel submissions, it can happen that the worker pool was not
+	// full during the if statement execution, but got filled by another goroutine from the router and then leads to
```
NHingerl, line 127:
```suggestion
	// full during the if statement execution, but was filled by another goroutine from the router, which then leads to
```

NHingerl (approved):

ruanxin (approved):

PR #957, opened 2022-03-22
Exclude pods in Pending phase from Istio Proxy Reset

cnvergence:
This fixes an issue, when customer pods with sidecar fail to reconcile, whenever they are stuck in a constant pending phase.
Added the function for checking the pod's state.

### DIFF
--- a/pkg/reconciler/instances/istio/reset/data/gatherer.go
+++ b/pkg/reconciler/instances/istio/reset/data/gatherer.go
@@ -59,14 +59,26 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
 		for _, container := range pod.Spec.Containers {
 			containsPrefix := strings.Contains(container.Image, image.Prefix)
 			hasSuffix := strings.HasSuffix(container.Image, image.Version)
-			isTerminating := pod.Status.Phase == "Terminating"
-			isPending := pod.Status.Phase == "Pending"
+			ready := isPodReady(pod)
 
-			if containsPrefix && !hasSuffix && !isTerminating && !isPending {
+			if containsPrefix && !hasSuffix && ready {
 				outputPodsList.Items = append(outputPodsList.Items, *pod.DeepCopy())
 			}
 		}
 	}
 
 	return
+}
+
+func isPodReady(pod v1.Pod) bool {
+	if pod.Status.Phase != v1.PodRunning {
+		return false
+	}
+	for _, condition := range pod.Status.Conditions {
+		if condition.Status != v1.ConditionTrue {
+			return false
+		}
+	}
+	//deletion timestamp determines whether pod is terminating or running (nil == running)
+	return pod.ObjectMeta.DeletionTimestamp == nil
 }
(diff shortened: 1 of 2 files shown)


pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -60,8 +60,9 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
 			containsPrefix := strings.Contains(container.Image, image.Prefix)
 			hasSuffix := strings.HasSuffix(container.Image, image.Version)
 			isTerminating := pod.Status.Phase == "Terminating"
+			isPending := pod.Status.Phase == "Pending"
 
-			if containsPrefix && !hasSuffix && !isTerminating {
+			if containsPrefix && !hasSuffix && !isTerminating && !isPending {
```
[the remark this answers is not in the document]
cnvergence: Good idea, I will try it
also I think there is no pod.phase Terminating. so we should clean it up a bit

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -59,13 +59,26 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
...
+func isPodReady(pod v1.Pod) bool {
+	if pod.Status.Phase != v1.PodRunning {
+		return false
+	}
+	for _, condition := range pod.Status.Conditions {
+		if condition.Status != v1.ConditionTrue {
+			return false
+		}
+	}
+	//deletion timestamp determines whether pod is terminating or running (nil == running)
```
mjakobczyk, line 82: Instead of this comment I would rather see a function with a proper name and docs:
```go
func isPodRunning(pod v1.Pod) bool {
  return pod.ObjectMeta.DeletionTimestamp == nil
}
```

pkg/reconciler/instances/istio/reset/data/gatherer_test.go
```diff
@@ -72,8 +74,10 @@ func Test_Gatherer_GetPodsWithDifferentImage(t *testing.T) {
 		pods.Items = []v1.Pod{
 			*podWithExpectedImage,
 			*podWithExpectedImageTerminating,
+			*podWithExpectedImagePending,
```
mjakobczyk, line 77: I had a look at `should get one pod from the list` test case - could you please add one more assertion to verify which exactly pod was returned in the list? There are many of them so we should be sure that appropriate one was returned.

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -59,13 +59,26 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
...
 		}
 	}
 
 	return
 }
+
+func isPodReady(pod v1.Pod) bool {
+	if pod.Status.Phase != v1.PodRunning {
+		return false
+	}
```
mjakobczyk, line 76: Missing new line

pkg/reconciler/instances/istio/reset/data/gatherer.go
mjakobczyk, line 82: What is the difference between checking for pod running status like this or with `pod.Status.Phase` as you did above? Can't we reuse that logic for checking running status?

mjakobczyk (changes requested):

pkg/reconciler/instances/istio/reset/data/gatherer.go
cnvergence: There is no Terminating phase, so we need to check also for DeletionTimestamp

mjakobczyk (approved):
Looks decent after changes 💯

PR #963, opened 2022-03-23
feat: Cleanup cluster entities

adityabhatia:


pkg/scheduler/service/cleaner.go
```diff
@@ -53,18 +53,30 @@ func (c *cleaner) Run(ctx context.Context, transition *ClusterStatusTransition,
...
+		c.logger.Infof("%s Cleaner will remove unnecessary reconciliations", CleanerPrefix)
 		c.purgeReconciliationsNew(transition, config)
 	} else {
-		c.logger.Infof("%s Cleaner will remove entities older than %s", CleanerPrefix, config.PurgeEntitiesOlderThan.String())
+		c.logger.Infof("%s Cleaner will remove reconciliations older than %s", CleanerPrefix, config.PurgeEntitiesOlderThan.String())
 		c.purgeReconciliationsOld(transition, config)
 	}
 
+	// delete cluster entities
+	clusterInventoryCleanupDays := 20 // days default
```
Tomasz-Smelcerz-SAP, line 69: I think that the default should not be set here, but should instead come from configuration/flags definition

Tomasz-Smelcerz-SAP (approved):

pkg/scheduler/service/cleaner.go
adityabhatia: Makes sense, let's take a decision on this in a follow-up

PR #970, opened 2022-03-28
tests: Test suite for scheduler/reconciliation, scheduler/service & cluster pkgs

adityabhatia:


pkg/scheduler/service/transition_test.go
```diff
@@ -44,99 +41,123 @@ func TestTransition(t *testing.T) {
...
+	err = transition.StartReconciliation(clusterState.Cluster.RuntimeID, clusterState.Configuration.Version, &SchedulerConfig{
 		PreComponents:  nil,
 		DeleteStrategy: "",
-	}
+	})
+	require.NoError(t, err)
 
 	//cleanup at the end of the execution
-	defer func() {
+	cleanupFn := func() {
```
jakobmoellersap, line 51: I think you can simply use AfterTest for this instead of using a cleanup function

pkg/scheduler/service/scheduler_test.go
```diff
@@ -140,66 +135,58 @@ func createClusterStates(t *testing.T, inventory cluster.Inventory) []string {
...
+	for i := 0; i < 25; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			time.Sleep(time.Until(startAt))
+			err := scheduler.Run(ctx, &ClusterStatusTransition{
+				conn:      dbConn,
+				inventory: inventory,
+				reconRepo: reconRepo,
+				logger:    logger.NewLogger(true),
```
jakobmoellersap, line 178: if you use the testLogger here you dont have to initialize a new one

pkg/scheduler/service/scheduler_test.go
```diff
@@ -140,66 +135,58 @@ func createClusterStates(t *testing.T, inventory cluster.Inventory) []string {
...
+	require.NoError(t, err)
 
-	t.Run("Multiple scheduler watching same inventory", func(t *testing.T) {
-		//initialize WaitGroup
-		var wg sync.WaitGroup
+	clusterRuntimeIDs := createClusterStates(t, inventory)
 
-		inventory, err := cluster.NewInventory(dbConnection(t), true, cluster.MetricsCollectorMock{})
-		require.NoError(t, err)
+	scheduler := newScheduler(logger.NewLogger(true))
```
jakobmoellersap, line 150: if you use the testLogger here you dont have to initialize a new one

pkg/scheduler/service/scheduler_test.go
```diff
@@ -19,65 +19,60 @@ import (
...
+	requiredReconciliationEntity(t, reconRepo, clusterState)
+}
 
-	t.Run("Test run", func(t *testing.T) {
-		reconRepo := reconciliation.NewInMemoryReconciliationRepository()
-		scheduler := newScheduler(logger.NewLogger(true))
+func (s *reconciliationTestSuite) TestSchedulerRun() {
+	t := s.T()
+	reconRepo := reconciliation.NewInMemoryReconciliationRepository()
+	scheduler := newScheduler(logger.NewLogger(true))
```
jakobmoellersap, line 35: if you use the testLogger here you dont have to initialize a new one

pkg/scheduler/reconciliation/reconciliation_suite_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/kyma-incubator/reconciler/pkg/model"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	"go.uber.org/zap"
+	"path/filepath"
+	"sync"
+	"testing"
+)
+
+type reconciliationTestSuite struct {
```
jakobmoellersap, line 18: if you split the transaction relevant tests (e.g. rollback checks) into a second test suite, all others could run with the inbuilt Tx Connection I think, but I will double check this on a second review. Not sure if we want this at all, just a thought as you implemented this with an isolated suite and custom Connection methods

jakobmoellersap (changes requested):
Overall looks good! I think a lot of these tests could also work without the custom Connection method but its mainly flavor for me how you want to use that.

There are a lot of custom initializations of loggers needed for dependencies or tested objects, I think here we should use the test logger instead. Later on (like i did in #790) we can implement a Zap Logger here that is compatible with testing.T and the go test buffers as right now the inventory logs would e.g. not be synced / flushed correctly in parallel execution mode of some tests. I tried to mark a few initializations but maybe I missed some.

One thing to note is that this PR changes some files the same way as #790 to make the test case usable. Depending on which is merged first, we will need to rebase the other and merge changes in the later pr.

pkg/cluster/cluster_suite_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"testing"
+)
+
+type clusterTestSuite struct {
+	suite.Suite
+	containerSuite   *db.ContainerTestSuite
+	testContext      context.Context
+	testLogger       *zap.SugaredLogger
+	serverStartMutex sync.Mutex
+	debugLogs        bool
```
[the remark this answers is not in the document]
adityabhatia: done, removed it from the other unused tests suites as well

pkg/scheduler/service/transition_test.go
adityabhatia: This was not generic enough amongst tests of this package. For more generalisation, the tests need to be completely reworked on, which might go out of scope of the current PR. I've added an AfterTest(), which only serves this specific test for now.

pkg/scheduler/service/scheduler_test.go
adityabhatia: used the test logger in this suite

pkg/scheduler/reconciliation/reconciliation_suite_test.go
adityabhatia: I thought about it too, IMO it would an overkill, since most tests require multiple transactions (aka dbConnection) for tests. But sure we could think about it an a follow-up PR I'd say.

pkg/cluster/inventory_test.go
```diff
@@ -30,13 +30,18 @@ var clusterStatuses = []model.Status{
...
+	inventory := s.newInventory(&conn)
 	expectedCluster := test.NewCluster(t, "1", 1, false, test.Production)
 
-	removeAllClusters(t, inventory)       //cleanup before the test runs
-	defer removeAllClusters(t, inventory) //cleanup after test is finished
+	removeAllClusters(t, inventory) //cleanup before the test runs
+	defer (func() {
+		removeAllClusters(t, inventory) //cleanup after test is finished
+		require.NoError(t, conn.Close())
+	})()
```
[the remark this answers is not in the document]
adityabhatia: done

jakobmoellersap (approved):
LGTM! 🔥  (but needs rebase)

Issue #967, opened 2022-03-25
Istio Component Reconciler should always patch MutatingWebHook


veichtj:
**Description**
The istio component reconciler is patching the istio MutatingWebHook at the moment as a PostReconcileAction, what means it does the patch only if previous steps (pre-version check and install / upgrade) were successful.
If the install of istio failed by whatever reason the patch is not applied and the MutatingWebHook misses the label that is mandatory on gardener clusters to having them reconciled properly.

**Expected result**
The istio component reconciler should always try to patch the MutatingWebHook ignoring the fact if istio installation worked successfully or not. This should prevent gardener clusters from failing in reconcilation.

**Actual result**
The MutatingWebHook is only patched if previous istio reconcilation steps where successful.

**Steps to reproduce**

PR #971, opened 2022-03-28
Always patch MutatingWebHook

dariusztutaj:
Added patching MutatingWebHook even in case of error during install/update istio

Related issues:
#967


pkg/reconciler/instances/istio/action.go
```diff
@@ -89,6 +89,7 @@ func (a *MainReconcileAction) Run(context *service.ActionContext) error {
 
 		err = performer.Install(context.KubeClient.Kubeconfig(), istioManifest.Manifest, istioStatus.TargetVersion, context.Logger)
 		if err != nil {
+			_ = performer.PatchMutatingWebhook(context.Context, context.KubeClient, context.Logger)
```
mjakobczyk, line 92: We should always catch an error

pkg/reconciler/instances/istio/action.go
```diff
@@ -97,6 +98,10 @@ func (a *MainReconcileAction) Run(context *service.ActionContext) error {
 
 		err = performer.Update(context.KubeClient.Kubeconfig(), istioManifest.Manifest, istioStatus.TargetVersion, context.Logger)
 		if err != nil {
+			errPatchMutatingWebhook := performer.PatchMutatingWebhook(context.Context, context.KubeClient, context.Logger)
```
mjakobczyk, line 101: We should just name error env as `err`

pkg/reconciler/instances/istio/action.go
```diff
@@ -97,6 +98,10 @@ func (a *MainReconcileAction) Run(context *service.ActionContext) error {
...
+			if errPatchMutatingWebhook != nil {
+				context.Logger.Info("Could not patch MutatingWebhookConfiguration")
```
mjakobczyk, line 103: In case error appears we should propagate it to the upper layer. Even if we decided to log it then it should be `Error` instead of `Info`

mjakobczyk (changes requested):

dariusztutaj:
Closing in case of too many changes have been made incorrectly, I lost. I'll create new clean PR

PR #981, opened 2022-03-30
Fix: Race condition in occupancy tracker for component reconcilers

khlifi411:
**List of changes**
- Added mutex for CR occupancy tracker for `occupancyCallbackURL`

### DIFF
--- a/pkg/reconciler/service/occupancytracker.go
+++ b/pkg/reconciler/service/occupancytracker.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"sync"
 	"time"
 )
 
@@ -24,6 +25,7 @@ type OccupancyTracker struct {
 	occupancyID          string
 	occupancyCallbackURL string
 	ticker               *time.Ticker
+	sync.Mutex
 }
 
 func newOccupancyTracker(debug bool) *OccupancyTracker {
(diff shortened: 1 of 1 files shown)


pkg/reconciler/service/occupancytracker.go
```diff
@@ -137,7 +139,9 @@ func (t *OccupancyTracker) AssignCallbackURL(callbackURL string) {
 		t.ticker.Reset(defaultInterval)
 	} else if t.occupancyID != "" {
 		var err error
+		t.Lock()
 		t.occupancyCallbackURL, err = parseOccupancyCallbackURL(callbackURL, t.occupancyID)
```
jakobmoellersap, line 143: I think this only locks for parsing the callback URL, is this really the lock condition?

pkg/reconciler/service/occupancytracker.go
```diff
@@ -24,6 +25,7 @@ type OccupancyTracker struct {
 	occupancyID          string
 	occupancyCallbackURL string
 	ticker               *time.Ticker
+	sync.Mutex
```
jakobmoellersap, line 28: I think the mutex is never initialized in the new method

pkg/reconciler/service/occupancytracker.go
khlifi411: Here, I am locking the assignment of `t.occupancyCallbackURL` which will be read by the `(*OccupancyTracker).Track()` method (from a different go routine)

pkg/reconciler/service/occupancytracker.go
khlifi411: Here, I am using struct and not a pointer to it, so it will be initialized to an unlocked Mutex state according to the docs.
`Mutex {0,0}`

PR #982, opened 2022-03-30
Fix sec findings 20220330

tobiscr:


jakobmoellersap (commented):
Shouldn't Protecode run green here with your changes?

jakobmoellersap (approved):

Issue #676, opened 2022-01-14
Improvements to Connectivity Proxy removal logic


janmedrek:
**Description**

Due to our template-apply approach, it might be possible that the Connectivity Proxy reconciler lefts some resources after its removal. Implementing a labeling mechanism to enable precise identification of CP components is needed to preserve the state of the cluster in a good shape.

**Reasons**

We do not want any leftovers in case the component is removed.

**Acceptance Criteria**
- [ ] There are no leftovers after Connectivity Proxy is removed by the reconciler

janmedrek:
I have tried installing and removing the Connectivity Proxy on my cluster, the only observation I have is that `connectivity-proxy-registry-config` Secret in `kyma-system` persists.

There were reports of stuff not being deleted in https://github.tools.sap/kyma/backlog/issues/2214 but I was not able to reproduce it. We should still have that case in mind.

janmedrek:
What would need to be done:
- [ ] Install Connectivity Proxy on the Cluster
- [ ] Remove it
- [ ] Check whether any resources tied to the Connectivity Proxy are abandoned on the cluster
  - `kyma-system` and `istio-system` are namespaces where secrets tend to stay
- [ ] Enhance removal logic to cover those resources

[Remove action in Reconciler](https://github.com/kyma-incubator/reconciler/blob/4b5c4d2ee641944b42a1c26b592275ff3f8f196d/pkg/reconciler/instances/connectivityproxy/command.go#L120)

janmedrek:
I can see the `connectivity-proxy-status` ConfigMap is still present in the `kyma-system` namespace after the Proxy was removed, is this intentionally applied by the reconciler?

mvshao:
Yes, the reconciler intentionally applied `connectivity-proxy-status` ConfigMap during cluster creation/reconciliation. 
Here is a result from the `kubectl get configmap -n kyma-system` command (cluster was created about ~50 minutes ago, Proxy ~5 minutes)

`connectivity-proxy 1      4m24s`
`connectivity-proxy-status                       4      47m`

Content of `connectivity-proxy-status` during reconciliation
```
apiVersion: v1
data:
  last-reconciliation: 2022-05-20 09:27:18.750992714 +0000 UTC m=+89751.808678866
  name: connectivity-proxy
  status: running
  version: ""
kind: ConfigMap
metadata:
  creationTimestamp: "2022-05-20T08:34:35Z"
  labels:
[...]

PR #986, opened 2022-03-31
The logic for removal of remaining certificates after deletion of the Connectivity-Proxy

mvshao:
**Description**
After the Connectivity-Proxy reconciler removes the component there are still leftovers in `istio-system` namespace. 

**Changes**
- added logic for removal remaining certificates `cc-certs` and `cc-certs-cacert`

**Related issue**
#676


pkg/reconciler/instances/connectivityproxy/command.go
```diff
@@ -134,5 +135,18 @@ func (a *CommandActions) Remove(context *service.ActionContext) error {
 	if err != nil {
 		return errors.Wrap(err, "Error during removal")
 	}
+
+	context.Logger.Info("Deleting secrets")
+
+	_, err = context.KubeClient.DeleteResource(context.Context, "secret", "cc-certs", "istio-namespace")
```
akgalwas, line 141: I performed a test and noticed that secrets are not deleted. I think the reason is that service account for Connectivity Proxy doesn't have delete permission. Please see [this](https://github.com/kyma-project/control-plane/blob/main/resources/kcp/charts/component-reconcilers/templates/rbac.yaml).

akgalwas (changes requested):

akgalwas:
The following PR adds privileges for removing secrets in Istio namespace: https://github.com/kyma-project/control-plane/pull/1578

pkg/reconciler/instances/connectivityproxy/action.go
```diff
@@ -25,6 +25,11 @@ func (a *CustomAction) Run(context *service.ActionContext) error {
 	if context.Task.Type == model.OperationTypeDelete {
 		context.Logger.Info("Requested cluster removal - removing component")
 		if err := a.Commands.Remove(context); err != nil {
+			context.Logger.Error("Failed to remove Connectivity Proxy: %v", err)
+			return err
+		}
+		if err := a.Commands.RemoveIstioSecrets(context); err != nil {
```
koala7659, line 31: This does not need to be called since `RemoveIstioSecrets` is called by Commands.Remove method (see L139 in [pkg/reconciler/instances/connectivityproxy/command.go)  After this  change you can delete `RemoveIstioSecrets`   method from the interface and itd mocked conterpart. I would suggest  make this method as package private

pkg/reconciler/instances/connectivityproxy/command.go
```diff
@@ -134,5 +135,22 @@ func (a *CommandActions) Remove(context *service.ActionContext) error {
 	if err != nil {
 		return errors.Wrap(err, "Error during removal")
 	}
+
+	return a.RemoveIstioSecrets(context)
+}
+
+func (a *CommandActions) RemoveIstioSecrets(context *service.ActionContext) error {
```
koala7659, line 142: I would suggest  make this method as package private since it is the only place when it is executed.

koala7659: Maybe the name can be a little bit more clear, like RemoveIstioSecretsFromShoot, but such name is loo long?

koala7659 (changes requested):
Please check my comments

pkg/reconciler/instances/connectivityproxy/command.go
mvshao: Personally, I think the `RemoveIstioSecretsFromShoot` is a bit too long. We are performing the deletion on the shoot, so we can figure out this function is linked to the shoot.

pkg/reconciler/instances/connectivityproxy/mocks/commands.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.1.0. DO NOT EDIT.
+// Code generated by mockery 2.9.0. DO NOT EDIT.
```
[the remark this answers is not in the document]
mvshao: Why we can get rid of this file? The file also contains the other mocked functions e.g. `Remove`. Maybe we need only to delete the generated `RemoveIstioSecrets`?

koala7659 (approved):
LGTM

akgalwas (approved):

Issue #984, opened 2022-03-31
[Istio reconciler] proxy reset job should put annotations on failed pods


strekm:
**Description**

Istio proxy reset post action is responsible to reset workloads after Istio reset. Rollout restart is only executed on pods which have parents. After determining entities restart is performed and progress is tracked. In some cases rollout restart ends up in errors (e.g. image pull back off) and action reports an error. This leads to failed kyma reconciliations. Stand alone pods are omitted, which might cause connectivity issues since Istio not guarantee that old sidecars will work with new Istio. 

Proposed solution is to annotate troublesome pods instead of failing proxy reset action. 

AC:
- [x] pods which can't be restarted should be annotated
- [x] omitted pods should be annotated
- [x] action never fails
- [x] troubleshooting guide with workaround created (delete pods)
- [x] reconciler bumped in control-plane

**Expected result**

proxy reset never fails, failed pods are annotated

**Actual result**

proxy reset fails on restarting pods

strekm:
blocked by https://github.com/kyma-incubator/reconciler/issues/997

jakobmoellersap:
#997 isn't blocking since yesterday as @tobiscr rollbacked the issue and corrected the mod files

piotrkpc:
Merged in control-plane: https://github.com/kyma-project/control-plane/pull/1519

PR #989, opened 2022-04-01
istio: handle unhealthy pods during upgrade

piotrkpc:
properly handle unhealthy pods during proxy reset procedure during upgrade

See: https://github.com/kyma-incubator/reconciler/issues/984

### DIFF
--- a/pkg/reconciler/instances/istio/reset/data/gatherer.go
+++ b/pkg/reconciler/instances/istio/reset/data/gatherer.go
@@ -84,3 +84,15 @@ func isPodReady(pod v1.Pod) bool {
 
 	return pod.ObjectMeta.DeletionTimestamp == nil
 }
+
+// RemoveAnnotatedPods removes pods with annotation annotationKey from in podList
+func RemoveAnnotatedPods(in v1.PodList, annotationKey string) (out v1.PodList) {
+	in.DeepCopyInto(&out)
+	out.Items = []v1.Pod{}
+	for i := 0; i < len(in.Items); i++ {
+		if _, ok := in.Items[i].Annotations[annotationKey]; !ok {
+			out.Items = append(out.Items, in.Items[i])
+		}
+	}
+	return
+}
--- a/pkg/reconciler/instances/istio/reset/pod/handler.go
+++ b/pkg/reconciler/instances/istio/reset/pod/handler.go
@@ -135,6 +135,15 @@ func (i *RolloutHandler) ExecuteAndWaitFor(context context.Context, object Custo
 
 	if err != nil {
 		i.log.Warnf("unable to rollout %s/%s/%s: %v", object.Kind, object.Namespace, object.Name, err)
+		errAnnotate := k8sRetry.RetryOnConflict(k8sRetry.DefaultRetry, func() error {
+			client := i.kubeClient
+			err := annotateObject(context, client, object, AnnotationResetWarningKey, AnnotationResetWarningRolloutTimeoutVal)
+			return err
+		})
+
+		if errAnnotate != nil {
+			return errAnnotate
+		}
 		return err
 	}
 
@@ -175,6 +184,11 @@ func annotateObject(context context.Context, client kubernetes.Interface, object
 
 	case "StatefulSet":
 		err := annotatePodsFor(context, getStatefulSetLabelSelector, client, object, key, val)
+		if err != nil {
+			return err
+		}
+	case "Job":
+		err := annotatePodsFor(context, getJobLabelSelector, client, object, key, val)
 		if err != nil {
 			return err
 		}
@@ -223,6 +237,15 @@ func annotatePods(context context.Context, kubeClient kubernetes.Interface, podL
 		}
 	}
 	return nil
+}
+
+func getJobLabelSelector(context context.Context, client kubernetes.Interface, object CustomObject) (string, error) {
+	job, err := client.BatchV1().Jobs(object.Namespace).Get(context, object.Name, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+	labelSelector := buildLabelSelector(job.Spec.Selector.MatchLabels)
+	return labelSelector, nil
 }
 
 func getDeploymentLabelSelector(context context.Context, client kubernetes.Interface, object CustomObject) (string, error) {
--- a/pkg/reconciler/instances/istio/reset/pod/handler.go
+++ b/pkg/reconciler/instances/istio/reset/pod/handler.go
@@ -18,7 +18,7 @@ import (
 
 const (
 	AnnotationResetWarningKey               = "istio.reconciler.kyma-project.io/proxy-reset-warning"
-	AnnotationResetWarningNoOwnerVal        = "pod sidecar could not be updated because OwnerReferences was not found. Istio might not work. Recreate the pod manually to ensure full compatibility."
+	AnnotationResetWarningNoOwnerVal        = "pod sidecar could not be updated because OwnerReferences is Job or was not found . Istio might not work. Recreate the pod manually to ensure full compatibility."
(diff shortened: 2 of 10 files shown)


pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -84,3 +84,15 @@ func isPodReady(pod v1.Pod) bool {
 
 	return pod.ObjectMeta.DeletionTimestamp == nil
 }
+
+// FilterAnnotatedPods removes pods with annotation annotationKey from in podList
+func FilterAnnotatedPods(in v1.PodList, annotationKey string) (out v1.PodList) {
```
mjakobczyk, line 89: The name is misleading. `FilterAnnotatedPods` could mean that you will retrieve pods with the given annotation whereas right now it is opposite.

cnvergence (approved):

cnvergence:
Tested it out on the k3d cluster.
All is good to go, after the initial review, there is added support for Jobs and CronJobs. Annotation logic works properly.

werdes72 (approved):
/meow

pkg/reconciler/instances/istio/reset/pod/handler.go
```diff
@@ -112,10 +134,162 @@ func (i *RolloutHandler) ExecuteAndWaitFor(context context.Context, object Custo
...
 	}
 
-	return i.WaitForResources(context, object)
+	return nil
+}
+
+func annotateObject(context context.Context, client kubernetes.Interface, object CustomObject, key string, val string) error {
+	switch object.Kind {
+	case "Deployment":
+		err := annotatePodsFor(context, getDeploymentLabelSelector, client, object, key, val)
```
mjakobczyk, line 170: I can see a repetitive pattern in this function. For every k8s object you call `annotatePodsFor(...)`. What about creating a dedicated function around switch-case that would simply return us a label selector and then call mentioned function once for returned argument? Wouldn't it be cleaner?

pkg/reconciler/instances/istio/reset/pod/handler.go
```diff
@@ -112,10 +134,162 @@ func (i *RolloutHandler) ExecuteAndWaitFor(context context.Context, object Custo
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func listPodsWithLabelSelector(context context.Context, client kubernetes.Interface, namespace string, labelSelector string) (*v1.PodList, error) {
+	podList, err := client.CoreV1().Pods(namespace).List(context, metav1.ListOptions{LabelSelector: labelSelector})
+	if err != nil {
+		return nil, err
```
mjakobczyk, line 223: Although you can an error here you still report it would additional metadata (like custom message for `errors.New(...)` etc. Would it make any difference to return directly the output of the method call two lines above instead of error checking here and returning the result two lines later?

pkg/reconciler/instances/istio/reset/pod/handler.go
```diff
@@ -112,10 +134,162 @@ func (i *RolloutHandler) ExecuteAndWaitFor(context context.Context, object Custo
...
+func getJobLabelSelector(context context.Context, client kubernetes.Interface, object CustomObject) (string, error) {
+	job, err := client.BatchV1().Jobs(object.Namespace).Get(context, object.Name, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+	labelSelector := buildLabelSelector(job.Spec.Selector.MatchLabels)
+	return labelSelector, nil
+}
+
+func getDeploymentLabelSelector(context context.Context, client kubernetes.Interface, object CustomObject) (string, error) {
```
mjakobczyk, line 251: I have a feeling that this name is kind of misleading. When one uses `get` verb then we should safely assume that there is some underlying property returned or max one external call made. In fact what you do is you fetch labels from the cluster and perform some custom logic on them. What about calling it `build `the way that you did for the inner private method?

pkg/reconciler/instances/istio/reset/pod/handler.go
```diff
@@ -112,10 +134,162 @@ func (i *RolloutHandler) ExecuteAndWaitFor(context context.Context, object Custo
...
+	deployment, err := client.AppsV1().Deployments(object.Namespace).Get(context, object.Name, metav1.GetOptions{})
```
mjakobczyk, line 252: This function could be simplified to return only labels taken from `.Spec.Selector.MatchLabels` and in the layer above `buildLabelSelector(...)` could be called once for any labels coming from here. It would make this contract cleaner. What do you think?

pkg/reconciler/instances/istio/reset/pod/reset/action_test.go
```diff
@@ -145,4 +148,142 @@ func Test_DefaultPodsResetAction_Reset(t *testing.T) {
...
+		require.NoError(t, err)
+		shouldBeCalledWithoutAnnotatePod := func(t testing.TB) {
+			t.Helper()
+			matcher.AssertCalled(t, "GetHandlersMap", kubeClient, fixRetryOpts, v1.PodList{Items: []v1.Pod{simplePod}}, log, debug, fixWaitOpts)
+		}
+		shouldBeCalledWithoutAnnotatePod(t)
+	})
+}
+
+func createLabelSelector(labels map[string]string) string {
```
mjakobczyk, line 247: I kind of don't like the fact that you duplicate the logic used in the production code but I understand that it is for convenience reasons

mjakobczyk, line 247: Unfortunately, both this and `buildLabelSelector` function don't work.
For input:
`labels := map[string]string{"app": "test"}`
my output is:
`"app=test"`
For input:
`labels := map[string]string{"app": "test", "zapp": "test2", "wapp": "test3"}`
my output is:
`"wapp=test3"`

Is this how it should work?

pkg/reconciler/instances/istio/reset/pod/handler.go
```diff
@@ -112,10 +134,162 @@ func (i *RolloutHandler) ExecuteAndWaitFor(context context.Context, object Custo
...
+	if err != nil {
+		return "", err
+	}
+	labelSelector := buildLabelSelector(deployment.Spec.Selector.MatchLabels)
+	return labelSelector, nil
+}
+
+func buildLabelSelector(labels map[string]string) string {
```
mjakobczyk, line 260: Even though there are no unit tests for this function defined, so I can't see what should be the result, but from my point of view and what name with parameter suggest it does not work as intended.

Please see my second comment here: https://github.com/kyma-incubator/reconciler/pull/989/files#r844725156

pkg/reconciler/instances/istio/reset/pod/handler.go
```diff
@@ -55,10 +63,24 @@ type NoActionHandler struct {
...
+		i.log.Debugf("annotating object where we don't perform proxy reset for: %s/%s/%s", object.Kind, object.Namespace, object.Name)
 	}
-
-	return nil
+	err := k8sRetry.RetryOnConflict(k8sRetry.DefaultRetry, func() error {
+		pod, err := i.kubeClient.CoreV1().Pods(object.Namespace).Get(context, object.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		if pod.Annotations == nil {
```
mjakobczyk, line 73: `pod.Annotations` is a map underneath. Is there a chance that it could be empty but not nil?

pkg/reconciler/instances/istio/reset/pod/reset/action_test.go
mjakobczyk: If that is the case then the contract of the function is really unclear when you iterate over all map entries anyways and override the result.

I would suggest both renaming the function and adjusting the implementation as they are misleading. What about `buildLabelSelectorFromFirstEntryIn(map)`? Then in the loop when you find first entry with non-empty value you can simply break. Then we would know for sure what is going on.

PR #990, opened 2022-04-01
Avoid logging client-go warnings in production

moelsayed:
Client-go can be a bit chatty with deprecation warnings to make sure users upgrade their resource versions. With the expected scale of the reconciler, this can generate an excessive amount of logs. 

This PR sets a `logr.Discard()` logger for client-go when verbose logging is disabled, effectively masking all warnings and logged errors coming from client-go. All errors are returned from client-go immediately, so this shouldn't mask any bugs. This is not an ideal solution but it's useful in this case considering the resources consumed by these logs.

This option applies to the local reconciler and individual component reconcilers.

### DIFF
--- a/cmd/mothership/mothership/start/scheduler.go
+++ b/cmd/mothership/mothership/start/scheduler.go
@@ -14,7 +14,7 @@ import (
 
 func startScheduler(ctx context.Context, o *Options) error {
 
-	runtimeBuilder := service.NewRuntimeBuilder(o.Registry.ReconciliationRepository(), logger.NewLogger(o.Verbose))
+	runtimeBuilder := service.NewRuntimeBuilder(o.Registry.ReconciliationRepository(), logger.NewLogger(o.Verbose), o.Verbose)
 	ds, err := service.NewDeleteStrategy(o.Config.Scheduler.DeleteStrategy)
 	if err != nil {
 		return err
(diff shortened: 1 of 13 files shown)


cmd/mothership/mothership/start/scheduler.go
```diff
@@ -14,7 +14,7 @@ import (
 
 func startScheduler(ctx context.Context, o *Options) error {
 
-	runtimeBuilder := service.NewRuntimeBuilder(o.Registry.ReconciliationRepository(), logger.NewLogger(o.Verbose))
+	runtimeBuilder := service.NewRuntimeBuilder(o.Registry.ReconciliationRepository(), logger.NewLogger(o.Verbose), o.Verbose)
```
jakobmoellersap, line 17: I think here you are submitting the Verbose Flag twice. I suggest to omit this and instead use the Logger as intended and switch all log messages to debug unless they really need to be on info level. Then you won't have this issue anymore

jakobmoellersap (changes requested):

cmd/mothership/mothership/start/scheduler.go
jakobmoellersap: Regarding the original issue of client-go, I suggest we mask the Log with our own, overwriting the Level in this case.

moelsayed:
We decided that these logs are too important to discard even on production. So, closing this for now.

PR #994, opened 2022-04-05
Adjust cluster-essentials reconciler logging

moelsayed:


pPrecel (approved):

pkg/reconciler/instances/clusteressentials/clusteressentials.go
```diff
@@ -11,7 +11,7 @@ const ReconcilerName = "cluster-essentials"
 func init() {
 	log := logger.NewLogger(false)
 
-	log.Debugf("Initializing component reconciler '%s'", ReconcilerName)
+	log.Infof("Initializing component reconciler '%s'", ReconcilerName)
```
jakobmoellersap, line 14: Is this intended to be logged at higher level now? Just checking for safety.

jakobmoellersap (approved):
If

pkg/reconciler/instances/clusteressentials/clusteressentials.go
moelsayed: Yes. I think it makes sense to at least log the activation of the component reconciler. This provides feedback without too much noise. Making the reconciler completely silent in production could make it hard to know if it's even working or not.

PR #1019, opened 2022-04-08
Add optional skip for safety check in scmigration

wozniakjan:
https://github.com/kyma-incubator/reconciler/pull/955 introduced additional safety check so migration doesn't accidentally delete instances that were supposed to be migrated. This adds control to disable it in form of environment variable `SKIP_SAFE_DELETION_BROKER_CHECK=true`.


ralikio (approved):

szwedm (approved):
/lgtm

Issue #1008, opened 2022-04-06
Create POC for for mock-based Test Behavior and replacing db.NewTestConnection for DB Integration Tests


jakobmoellersap:
AC:
- Setup Mocking as POC on any Test in the Integration Tests and convert it to a unit test with mocking based on a library

- Mock Library should be smart enough to support all our behaviors probably needs to use reflection or another smart mechanism, needs to be included in POC. If there is no significant decoupling possible without writing custom code, we should favor our current approach (spys as structs initialized and updated on demand in the tests)

- Follow-Up is required after POC that is able to integrate the new tests into different pipelines

- We want to split our current Tests from 80% Integration Testing / 20% Unit Testing to 80% Unit Testing / 20% Integration Testing and it should show how we can conceptually achieve that (e.g. through documentation of the approach supported by the POC)

- We want to setup Test Profiles that are able to introduce different Scenarios we can target in our pipelines

PR #1029, opened 2022-04-12
Use GORM module to build SQL strings

jeremyharisch:


go.sum
```diff
@@ -2068,9 +2115,16 @@ gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C
...
+gorm.io/driver/postgres v1.0.8 h1:PAgM+PaHOSAeroTjHkCHCBIHHoBIf9RgPWGo8dF2DA8=
 gorm.io/driver/postgres v1.0.8/go.mod h1:4eOzrI1MUfm6ObJU/UcmbXyiHSs8jSwH95G5P5dxcAg=
+gorm.io/driver/postgres v1.3.3 h1:y6DU2kJgDNisxfAlmxRaQZOIy4ytnuYrpzpSFYnSfCY=
+gorm.io/driver/postgres v1.3.3/go.mod h1:y0vEuInFKJtijuSGu9e5bs5hzzSzPK+LancpKpvbRBw=
 gorm.io/gorm v1.20.12/go.mod h1:0HFTzE/SqkGTzK6TlDPPQbAYCluiVvhzoA1+aVyzenw=
+gorm.io/gorm v1.21.4 h1:J0xfPJMRfHgpVcYLrEAIqY/apdvTIkrltPQNHQLq9Qc=
 gorm.io/gorm v1.21.4/go.mod h1:0HFTzE/SqkGTzK6TlDPPQbAYCluiVvhzoA1+aVyzenw=
+gorm.io/gorm v1.23.1/go.mod h1:l2lP/RyAtc1ynaTjFksBde/O8v9oOGIApu2/xRitmZk=
+gorm.io/gorm v1.23.4 h1:1BKWM67O6CflSLcwGQR7ccfmC4ebOxQrTfOQGRE9wjg=
+gorm.io/gorm v1.23.4/go.mod h1:l2lP/RyAtc1ynaTjFksBde/O8v9oOGIApu2/xRitmZk=
```
jakobmoellersap, line 1881: Multiple Gorm versions, we should probably ping this and the postgres driver

pkg/cluster/inventory.go
```diff
@@ -164,16 +190,15 @@ func (i *DefaultInventory) createCluster(contractVersion int64, cluster *keb.Clu
 	}
 
 	//create new version
-	q, err := db.NewQuery(i.Conn, newClusterEntity, i.Logger)
+	q, err := db.NewQueryGorm(i.Conn, newClusterEntity, i.Logger)
 	if err != nil {
 		return nil, err
 	}
-	err = q.Insert().Exec()
+	newDbEntity, err := q.Insert(inventoryClusters{})
```
jakobmoellersap, line 199: is it okay that we do an empty insert here? what is the context here?

pkg/cluster/inventory.go
```diff
@@ -275,8 +299,8 @@ func (i *DefaultInventory) MarkForDeletion(runtimeID string) (*State, error) {
 func (i *DefaultInventory) Delete(runtimeID string) error {
 	dbOps := func(tx *db.TxConnection) error {
 		newClusterName := fmt.Sprintf("deleted_%d_%s", time.Now().Unix(), runtimeID)
+		// TODO: Rewrite with grom to stay consistend
```
jakobmoellersap, line 304: I think we should rewrite the entire inventory if we touch it already, otherwise im afraid we could into too many inconsistencies. Buuut if it keeps the PR small then we can do a follow-up (please open an issue for that then so the team can take it).

Either way small typo :D 
```suggestion
		// TODO: Rewrite with gorm to stay consistend
```

pkg/cluster/inventory.go
```diff
@@ -583,25 +608,23 @@ func (i *DefaultInventory) filterClusters(filters ...statusSQLFilter) ([]*State,
...
 	//SQL to retrieve the latest statuses => max(config_version) within max(cluster_version):
 	/*
 		select cluster_version, max(config_version) from inventory_cluster_config_statuses where cluster_version in (
-			select max(cluster_version) from inventory_cluster_config_statuses group by runtime_id
+			select max(cluster_version) from inventory_cluster_config_statuses where deleted = false group by runtime_id
 		) group by cluster_version
 	*/
 	dataRows, err := i.Conn.Query(
-		fmt.Sprintf(
+		fmt.Sprintf( // TODO: Rewrite with grom to stay consistend
```
jakobmoellersap, line 621:
```suggestion
		fmt.Sprintf( // TODO: Rewrite with gorm to stay consistend
```

pkg/db/query_gorm.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+type QueryGorm struct {
+	Conn          Connection
+	gormDB        *gorm.DB
+	columnHandler *ColumnHandler
+	DbEntity      DatabaseEntity
+}
+
+func NewQueryGorm(conn Connection, entity DatabaseEntity, logger *zap.SugaredLogger) (*QueryGorm, error) {
```
jakobmoellersap, line 17: Currently you create new GORM sessions per query multiple times even. I think this is very bad for Performance if we use the default settings. I think we should either create one shared session from the Connection or create one on-demand through an interface method. Right now the danger of misuse is too high imo.

pkg/cluster/inventory.go
jeremyharisch: This is not an empty insert, please have a look at the implementation of the Insert function. It creates a new row in the given table using the default values.

pkg/cluster/inventory.go
```diff
@@ -54,28 +56,51 @@ func (i *DefaultInventory) WithTx(tx *db.TxConnection) (Inventory, error) {
...
+	if err != nil {
+		return 0, err
+	}
 
 	errCnt := 0
-	for _, clusterStatus := range clusterStatuses {
-		clStatusEntity := clusterStatus.(*model.ClusterStatusEntity)
-		if clStatusEntity.Status.IsFinal() {
-			if statusInSlice(clStatusEntity.Status, errorStatus) {
+	for dataRows.Next() {
```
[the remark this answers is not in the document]
jeremyharisch: In this PR I am using GORM just for the SQL building and I still use our Query methods from the `Connection` interface. Thus, the `Find()` method is actually not used to execute the query, as it would normally do. Here, the `Find` method only is used for to add the `FROM` clause in the SQL String.
This behaviour can be changed by setting the DryRun mode in the Gorm configuration, which I also did in this PR. Meaning the "result" of such line (i.e. [here](https://github.com/kyma-incubator/reconciler/pull/1029/files#diff-f921d302f6db810d79d7007b3c8a95dc107a6a7e4da95f92f3a96cbb6f6589aaR77)) is just the SQL with its vars in a gorm struct. Thus it still can be used as a sub-query in other queries, like [here](https://github.com/kyma-incubator/reconciler/pull/1029/files#diff-f921d302f6db810d79d7007b3c8a95dc107a6a7e4da95f92f3a96cbb6f6589aaR708-R713).

But what we could also do, is using the `Find` method to actually execute the queries and thus also using the built-in parser of gorm. But I did not had the time until know to test this feature in regards of compatibility with our `tx` object.

I would suggest to create a follow-up issue for that.

This, btw also resolves the other comment from you (https://github.com/kyma-incubator/reconciler/pull/1029#discussion_r849157163)

pkg/db/query_gorm.go
jeremyharisch: I totally get your point! Tbh in the way I configured it in this PR at the moment with using the `DryRun` mode, we should not get any performance issues, since we are still using our `Connection` interface for the queries. Thus, the connection of the gorm session will not be used atm.

More in my other comment: https://github.com/kyma-incubator/reconciler/pull/1029#discussion_r849261031

But I can also rewrite it as a preparation for the follow-up.

jakobmoellersap: I agree with your conclusion and think we should not include this change right now but lets follow-up with a lower-prio ticket. This for sure still fixes the issues with the DB Query Building!

go.mod
```diff
@@ -79,6 +79,8 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
+require github.com/DATA-DOG/go-sqlmock v1.5.0
```
jakobmoellersap, line 82: This relates to https://github.com/kyma-incubator/reconciler/issues/1008 and could be a partial solve for our DB Unit Tests FYI @adityabhatia @Tomasz-Smelcerz-SAP

jakobmoellersap (approved):

jakobmoellersap (dismissed):

tobiscr:
Awesome, thanks for that awesome improvement in our stack!

jakobmoellersap (approved):

PR #1038, opened 2022-04-14
feat: add exponential retry when validating a cluster on create/update

jakobmoellersap:
This introduces an exponential backoff delay when validating a cluster after creating / updating a cluster as it is trying to list all namespaces to verify that the cluster connectivity is ok for the given kubeconfig.


pkg/kubernetes/clientbuilder.go
```diff
@@ -52,12 +65,30 @@ func (cb *ClientBuilder) Build(validate bool) (kubernetes.Interface, error) {
...
+			Duration: 5 * time.Second,
+			Factor:   2,
+			Jitter:   0,
+			Steps:    4,
+			Cap:      35 * time.Second,
+		}
+		err = wait.ExponentialBackoffWithContext(ctx, backoff, func() (done bool, err error) {
+			_, err = clientSet.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+			if err != nil {
+				cb.logger.Info(errors.Wrapf(err, "validation of connection to Kubernetes cluster %s failed (%.0f seconds elapsed since start), retrying", config.Host, time.Since(start).Seconds()))
```
Tomasz-Smelcerz-SAP, line 83: Isn't logger.Error is better here?

pkg/kubernetes/clientbuilder.go
```diff
@@ -52,12 +65,30 @@ func (cb *ClientBuilder) Build(validate bool) (kubernetes.Interface, error) {
 	}
 
 	if validate {
-		timeout := int64(5)
-		_, err := clientSet.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{
-			TimeoutSeconds: &timeout,
+		ctx, cancel := context.WithCancel(ctx)
```
Tomasz-Smelcerz-SAP, line 68: Do we need another cancellable ctx here?
We're not explicitly cancelling it and the `defer` is invoked only after `wait.ExponentialBackoffWithContext` is done, which means we're not really cancelling.
It looks we rely entirely on the backoff timeout calculated internally inside `wait.ExponentialBackoffWithContext`
I think it's enough to pass the "parent" context to the `wait.ExponentialBackoffWithContext`.

pkg/kubernetes/clientbuilder.go
jakobmoellersap: If we do error logging here then our control-plane might be spammed with outputs on bad connectivity. I left this on error after consulting with @tobiscr but I can also change it. WDYT?

pkg/kubernetes/clientbuilder.go
jakobmoellersap: For me it's more a convention thing that I always open up a context if I propagate it downstream. As upstream we're using the request context, I found it better to create a new context and use this inside the exponentialbackoffwithcontext. 

I agree regarding necessity though: I will exclude it as per your request because its not strictly necessary.

Issue #935, opened 2022-03-14
Istio reconciler support istio 1.13.x


strekm:
**Description**

Implement necessary changes to install / upgrade istio in 1.13 version. For this story we do not need multiple version support. 

- [x] istioctl 1.13.x is used for installation / upgrade
- [x] istio webhook patching updated if needed
- [x] gardener cluster can be hibernated
- [x] istio running in 1.13.x version
- [x] sidecars running in 1.13.x version
- [x] reconciler bumped in kyma cli

**Reasons**

Keeping Istio up to date

**Attachments**
https://github.com/kyma-project/test-infra/issues/5109

mjakobczyk:
Tested installation with images available after bump from here:
- https://github.com/kyma-project/test-infra/pull/5202

Local Istio chart config that got adjusted for testing purposes looked this way:
```yaml
# the rest of the config
# ...

global:
  containerRegistry:
    path: eu.gcr.io/kyma-project
  images:
    # these definitions aren't used by the chart, but they are required by external tool for list generation
    istio_proxyv2:
      name: "proxyv2"
      version: "1.13.2-distroless"
      directory: "external/istio"
      containerRegistryPath: "eu.gcr.io/kyma-project"
    istio_pilot:
      name: "pilot"
      version: "1.13.2-distroless"
      directory: "external/istio"
      containerRegistryPath: "eu.gcr.io/kyma-project"
  tracing:
    enabled: true
```

So far so good. I did not encounter any problems with installation or upgrade:

```bash
{"level":"INFO","time":"2022-04-07T13:16:10.853+0200","message":"Istio in version 1.13.2-distroless successfully installed"}
{"level":"INFO","time":"2022-04-07T13:16:16.275+0200","message":"Component 'istio' has status 'running'"}
{"level":"INFO","time":"2022-04-07T13:16:27.154+0200","message":"Found 1 pods with different istio proxy image ({istio/proxyv2 1.13.2-distroless-distroless})"}
{"level":"INFO","time":"2022-04-07T13:16:41.972+0200","message":"Proxy reset for 1 pods successfully done"}
{"level":"INFO","time":"2022-04-07T13:16:41.972+0200","message":"Component 'istio' has status 'success'"}
```

dariusztutaj:
cli bump waits for #1031

dariusztutaj:
CLI bumped: kyma-project/cli#1244

PR #1039, opened 2022-04-15
Fix of getting pods with istio sidecar upgrade required

barchw:
This fixes issue of proxies not being updated during bump Istio to 1.13.x, caused by prefix inconsistency between Istio images and Solo.io images.

Related issue: 
kyma-project/kyma#13618
#935


pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +56,19 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
 	outputPodsList.Items = []v1.Pod{}
 
 	for _, pod := range inputPodsList.Items {
+		isIstioPod := false
+		if _, ok := pod.Annotations["sidecar.istio.io/status"]; ok {
+			isIstioPod = true
+		} else {
+			continue
+		}
+		if !isPodReady(pod) {
```
mjakobczyk, line 65: Is pod readiness crucial in this scenario? Even if pod is not ready shouldn't it be included as long as it contains different image?

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +56,19 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
...
+			continue
+		}
 		for _, container := range pod.Spec.Containers {
-			containsPrefix := strings.Contains(container.Image, image.Prefix)
+			containsIstioPrefix := strings.Contains(container.Image, "istio") && strings.Contains(container.Image, "proxyv2")
 			hasSuffix := strings.HasSuffix(container.Image, image.Version)
-			readiness := isPodReady(pod)
-
-			if containsPrefix && !hasSuffix && readiness {
+			if isIstioPod && !hasSuffix && containsIstioPrefix {
```
mjakobczyk, line 71: `isIstioPod` boolean flag is unnecessary here. Considering its two values:
- if it was `true` it does not affect this condition,
- if it was `false` this statement would not have been reached at all (`continue` statement would be called above)

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +56,19 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
 	outputPodsList.Items = []v1.Pod{}
 
 	for _, pod := range inputPodsList.Items {
+		isIstioPod := false
```
mjakobczyk, line 59: There could be two improvements  in this piece of code.
1. Instead of creating a new var you could just fail-fast with continue when this check is evaluated to false,
2. this logic could be wrapped into a separate private function for readability

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +56,19 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
...
+			isIstioPod = true
+		} else {
+			continue
+		}
+		if !isPodReady(pod) {
+			continue
+		}
 		for _, container := range pod.Spec.Containers {
-			containsPrefix := strings.Contains(container.Image, image.Prefix)
+			containsIstioPrefix := strings.Contains(container.Image, "istio") && strings.Contains(container.Image, "proxyv2")
```
mjakobczyk, line 69: This var name does not reflect what it really keeps. There are two different checks evaluated:
1. `istio` keyword existence,
2. `proxyv2` keyword existence.
As a result, the env should not be called only as `containsIstioPrefix`. You could either rename it or run two different validation conditions.

pkg/reconciler/instances/istio/reset/data/gatherer.go
mjakobczyk, line 71: I feel like this statement has been already floated with conditions. What about creating a dedicated function to encapsulate logic responsible for performing Istio's containers validation?

pkg/reconciler/instances/istio/reset/data/gatherer_test.go
```diff
@@ -54,6 +54,7 @@ func Test_Gatherer_GetPodsWithDifferentImage(t *testing.T) {
 	podWithDifferentImageSuffixTerminating := fixPodWith("application", "kyma", "istio/proxyv2:1.10.2", "Terminating")
 	podWithDifferentImageSuffixPending := fixPodWith("application", "kyma", "istio/proxyv2:1.10.2", "Pending")
 	podWithDifferentImagePrefix := fixPodWith("application", "kyma", "istio/weirdimage:1.10.2", "Running")
+	podWithSoloImagePrefix := fixPodWith("istio", "custom", "istio-69e5957db6d2/proxyv2:1.10.2", "Running")
```
mjakobczyk, line 57: Two things here:
1. we should not use `Solo` keyword here as we should not rely on or even mention any external dependency,
2. do not be that specific with Istio image hash - as you can see from above it would be better to use sum buzz wording such as `custom` or `weird` image naming.

pkg/reconciler/instances/istio/reset/data/gatherer_test.go
```diff
@@ -103,6 +106,7 @@ func fixPodWith(name, namespace, image, phase string) *v1.Pod {
 			OwnerReferences: []metav1.OwnerReference{
 				{Kind: "ReplicaSet"},
 			},
+			Annotations: map[string]string{"sidecar.istio.io/status":"true"},
```
mjakobczyk, line 109: What about creating some with pods this annotation set to `true` and some to `false`? Did you think about such unit test cases?

pkg/reconciler/instances/istio/reset/data/gatherer_test.go
```diff
@@ -91,7 +94,7 @@ func Test_Gatherer_GetPodsWithDifferentImage(t *testing.T) {
 		// then
 		require.Equal(t, podsWithDifferentImage.Items, expected.Items)
 		require.NotEmpty(t, podsWithDifferentImage.Items)
-		require.Len(t, podsWithDifferentImage.Items, 1)
+		require.Len(t, podsWithDifferentImage.Items, 2)
```
mjakobczyk, line 97: What about validating which exactly pods are included in the list? Would it be possible to validate if the correct ones that you truly expect are here?

mjakobczyk (changes requested):

pkg/reconciler/instances/istio/reset/data/gatherer.go
dariusztutaj: logic changed

pkg/reconciler/instances/istio/reset/data/gatherer_test.go
dariusztutaj: logic changed

pkg/reconciler/instances/istio/reset/data/gatherer_test.go
dariusztutaj: Already adjusted. Does not matter what is the value of this annotation. If sidecar injection is enabled than this annotation exists

pkg/reconciler/instances/istio/reset/data/gatherer.go
dariusztutaj: adjusted

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +56,16 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
 	outputPodsList.Items = []v1.Pod{}
 
 	for _, pod := range inputPodsList.Items {
+		if _, ok := pod.Annotations["sidecar.istio.io/status"]; !ok {
+			continue
+		}
+		if !isPodReady(pod){
+			continue
+		}
```
mjakobczyk, line 64: You could somehow merge these two conditions in one sentence to avoid multiple branching with the same result (both end with `continue`).

pkg/reconciler/instances/istio/actions/performer.go
```diff
@@ -71,8 +71,17 @@ type chartValues struct {
...
-			} `json:"istio_pilot"`
+				Directory string `json:"directory"`
+				ContainerRegistryPath string `json:"containerRegistryPath"`
+			} `json:"istio_pilot"`;
+			IstioProxyV2 struct {
+				Name string `json:"name"`
+				Version string `json:"version"`
+				Directory string `json:"directory"`
+				ContainerRegistryPath string `json:"containerRegistryPath"`
+			} `json:"istio_proxyv2"`
```
mjakobczyk, line 84: Do we need all these properties? Are they all used? I only included the mandatory one not to pass stuff pointlessly.

pkg/reconciler/instances/istio/actions/performer.go
```diff
@@ -91,7 +100,7 @@ type IstioPerformer interface {
 	Update(kubeConfig, istioChart, targetVersion string, logger *zap.SugaredLogger) error
 
 	// ResetProxy resets Istio proxy of all Istio sidecars on the cluster. The proxyImageVersion parameter controls the Istio proxy version, it always adds "-distroless" suffix to the provided value.
```
mjakobczyk, line 102: Please adjust the documentation too.

pkg/reconciler/instances/istio/actions/performer.go
```diff
@@ -363,6 +377,41 @@ func getTargetVersionFromIstioChart(workspace chart.Factory, branch string, isti
...
+		return "", err
+	}
+
+	var chartValues chartValues
+	err = json.Unmarshal(mapAsJSON, &chartValues)
+	if err != nil {
+		return "", err
+	}
+
+	if err!=nil{
```
mjakobczyk, line 402: Missing spaces.

mjakobczyk, line 402: By the way this error handling is doubled - already done a few lines above.

pkg/reconciler/instances/istio/actions/performer.go
```diff
@@ -363,6 +377,41 @@ func getTargetVersionFromIstioChart(workspace chart.Factory, branch string, isti
...
+		return "", err
+	}
+
+	prefix := chartValues.Global.Images.IstioProxyV2.ContainerRegistryPath + "/" + chartValues.Global.Images.IstioProxyV2.Directory
```
mjakobczyk, line 406: Could you use `fmt.Sprintf()` for string formatting instead of raw concatenation?

pkg/reconciler/instances/istio/actions/performer.go
```diff
@@ -363,6 +377,41 @@ func getTargetVersionFromIstioChart(workspace chart.Factory, branch string, isti
...
+
+	if prefix != "" {
```
mjakobczyk, line 408: IMHO it is impossible for the `prefix` to be empty ever. Pay attention that you always concatenate `/` sign.

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +56,16 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
...
 		for _, container := range pod.Spec.Containers {
-			containsPrefix := strings.Contains(container.Image, image.Prefix)
+			containsIstioProxyV2Prefix := strings.Contains(container.Image, image.Prefix)
```
mjakobczyk, line 66: Shouldn't this name stay generic? Keep in mind that `image` property is still called `Prefix` (accessed by `image.Prefix`) - nothing about Istio proxyv2.

pkg/reconciler/instances/istio/actions/performer_test.go
```diff
@@ -465,7 +465,7 @@ func Test_DefaultIstioPerformer_ResetProxy(t *testing.T) {
 		proxyImageVersion := "1.2.0"
 
 		// when
-		err := wrapper.ResetProxy(ctx, kubeConfig, proxyImageVersion, log)
+		err := wrapper.ResetProxy(ctx, kubeConfig, proxyImageVersion, "",log)
```
mjakobczyk, line 468: Maybe it would be more fitting if you created env such as `proxyImageVersion` and make it empty.

mjakobczyk (changes requested):

pkg/reconciler/instances/istio/actions/performer.go
barchw: Changed logic to check whether directory or registry path are empty

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +58,30 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
...
+		}
+
+		type IstioStatusStruct struct {
+			Containers []string `json:"containers"`
+		}
+		istioStatus := IstioStatusStruct{}
+		err := json.Unmarshal([]byte(fmt.Sprintf("{%s}", pod.Annotations["sidecar.istio.io/status"])), &istioStatus)
+		if err != nil {
+			continue
+		}
```
mjakobczyk, line 72: You could wrap it in some func and give it a meaningful name such as: `retrieveIstioContainerNames`

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +58,30 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
 	outputPodsList.Items = []v1.Pod{}
 
 	for _, pod := range inputPodsList.Items {
+		if _, ok := pod.Annotations["sidecar.istio.io/status"]; !ok || !isPodReady(pod) {
+			continue
+		}
+
+		type IstioStatusStruct struct {
```
mjakobczyk, line 65: If you define this struct in-line then it does not have to be exported

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +58,30 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
 	outputPodsList.Items = []v1.Pod{}
 
 	for _, pod := range inputPodsList.Items {
+		if _, ok := pod.Annotations["sidecar.istio.io/status"]; !ok || !isPodReady(pod) {
```
mjakobczyk, line 61: You could also call this piece of code as: `containsIstioSidecarAnnotation`

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +58,30 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
...
+		}
+
 		for _, container := range pod.Spec.Containers {
+			isIstioSidecar := false
+			for _, c := range istioStatus.Containers {
+				if c == container.Name {
+					isIstioSidecar = true
+					break
+				}
+			}
```
mjakobczyk, line 81: Could you wrap it with some function? It's 3rd inner loop

mjakobczyk (changes requested):

pkg/reconciler/instances/istio/reset/data/gatherer.go
```diff
@@ -56,12 +57,19 @@ func (i *DefaultGatherer) GetPodsWithDifferentImage(inputPodsList v1.PodList, im
 	outputPodsList.Items = []v1.Pod{}
 
 	for _, pod := range inputPodsList.Items {
+		if _, containsIstioSidecarAnnotation := pod.Annotations["sidecar.istio.io/status"]; !containsIstioSidecarAnnotation || !isPodReady(pod) {
+			continue
+		}
+
+		istioSidecarNames := getIstioSidecarNamesFromAnnotations(pod.Annotations)
```
cnvergence, line 64: I had only one question if we should really process those annotations.
The name of `istio-proxy` container is the default and it should not change in general.

mjakobczyk (dismissed):

pkg/reconciler/instances/istio/reset/data/gatherer.go
barchw: We discussed this issue and decided to process it just in case, although the name will most likely always stay as "istio-proxy". It also takes care of a situation where there are more than one Istio sidecars (although very unlikely)

mjakobczyk (approved):

PR #1072, opened 2022-05-04
Add preaction to handle enabling/disabling the jestream with file storage

VladislavPaskar:
**Description**

When switching from NATS to NATS Jestream with file storage or the other way around we get the PATCH-action error: 
` kubeClient failed to update StatefulSet 'eventing-nats' (namespace: kyma-system)  with strategy 'PATCH': cannot patch "eventing-nats" with kind StatefulSet: StatefulSet.apps "eventing-nats" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden` 
To fix that we need to recreate the StatefulSet, and in that case the it will be recreated with the updated PVC configuration.

**Changes proposed in this Pull Request**
1. Scenario NATS -> NATS Jetstream with file storage:
    - check that js is enabled and file storage mode is set and remove delete the StatefulSet, so that it would get recreated in the helm charts apply step
2. Scenario NATS Jetstream with file storage -> NATS:
    - check if the jestream is disabled, but there is PVC part in StatefulSet: delete the StatefulSet and remove the orphan PVC(s).

**Related Issue**
https://github.com/kyma-project/kyma/issues/13777

### DIFF
--- a/pkg/reconciler/instances/eventing/preaction/handletogglejsfilestorage.go
+++ b/pkg/reconciler/instances/eventing/preaction/handletogglejsfilestorage.go
@@ -1,0 +1,205 @@
+package preaction
+
+import (
+	"fmt"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/chart"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/instances/eventing/log"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/kubernetes/progress"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/service"
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v3"
+	appsv1 "k8s.io/api/apps/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8s "k8s.io/client-go/kubernetes"
+	"strings"
+)
+
+const (
+	handleEnableJSFileStorageName = "handleToggleJSFileStorage"
+	natsComponentName             = "eventing-nats"
+	statefulSetName               = "eventing-nats"
+	eventingComponentName         = "eventing"
+	fileStorageType               = "file"
+	volumeClaimName               = natsComponentName + "-js-pvc"
+	statefulSetType               = "StatefulSet"
+)
+
+type handleToggleJSFileStorage struct {
+	kubeClientProvider
+}
+
+// newHandleEnableJSFileStorage
+func newHandleToggleJSFileStorage() *handleToggleJSFileStorage {
+	return &handleToggleJSFileStorage{
+		kubeClientProvider: defaultKubeClientProvider,
+	}
+}
+
+// Config holds the global configuration values.
+type Config struct {
+	Global Global
+}
+
+// Global configuration of Jetstream feature.
+type Global struct {
+	Jetstream Jetstream `yaml:"jetstream"`
+}
+
+// Jetstream specific values like enabled flag and storage type.
+type Jetstream struct {
+	Enabled string `yaml:"enabled"`
+	Storage string `yaml:"storage"`
+}
+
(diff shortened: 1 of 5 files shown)


pkg/reconciler/instances/eventing/preaction/handletogglejsfilestorage.go
```diff
@@ -0,0 +1,205 @@
...
+	if jetstreamEnabled && strings.EqualFold(fileStorageType, fmt.Sprint(storageType)) {
+		if !volumeClaimTemplateFound {
+			if err := deleteNatsStatefulSet(context, clientSet, tracker, logger, statefulSet, false); err != nil {
+				return err
+			}
+		} else {
+			logger.With(log.KeyReason, "Nats StatefulSet already contains a PVC").Info("Step skipped")
+		}
+		return nil
+	}
```
mfaizanse, line 126: (nit) this nested ifs can be simplified:
```
if jetstreamEnabled && !volumeClaimTemplateFound && strings.EqualFold(fileStorageType, fmt.Sprint(storageType)) {
		logger.With(log.KeyReason, "Nats StatefulSet does not contains a PVC for JetStream File Storage").Info("Removing StatefulSet")
		if err := deleteNatsStatefulSet(context, clientSet, tracker, logger, statefulSet, false); err != nil {
			return err
		}
		return nil
}
```

mfaizanse, line 126: (nit) this nested ifs can be simplified:
```
if jetstreamEnabled && !volumeClaimTemplateFound && strings.EqualFold(fileStorageType, fmt.Sprint(storageType)) {
		logger.With(log.KeyReason, "Nats StatefulSet does not contains a PVC for JetStream File Storage").Info("Removing StatefulSet")
		if err := deleteNatsStatefulSet(context, clientSet, tracker, logger, statefulSet, false); err != nil {
			return err
		}
		return nil
}
```

pkg/reconciler/instances/eventing/preaction/handletogglejsfilestorage.go
```diff
@@ -0,0 +1,205 @@
...
+
+	logger.With(log.KeyReason, "No actions needed").Info("Step skipped")
+	return nil
+}
+
+// deleteNatsStatefulSet delete the Nats StatefulSet and optionally its assigned PVC.
+func deleteNatsStatefulSet(context *service.ActionContext, clientSet k8s.Interface, tracker *progress.Tracker, logger *zap.SugaredLogger, statefulSet *appsv1.StatefulSet, deletePVC bool) error {
+	// In case of switching from Nats Jestream to NATS the PVCs assigned to the Nats Stateful set should be deleted
+	if deletePVC {
+		err := deleteNatsJsPVC(context, statefulSet, clientSet, logger)
```
mfaizanse, line 145: Can a PVC be deleted if its bind to some StatefulSet?
What if we first delete the StatefulSet and then delete the PVC?
Please test and verify this workflow?

mfaizanse, line 145: Can a PVC be deleted if its bind to some StatefulSet?
What if we first delete the StatefulSet and then delete the PVC?
Please test and verify this workflow?

pkg/reconciler/instances/eventing/preaction/handletogglejsfilestorage.go
VladislavPaskar: I know, but the reason I put the `if !volumeClaimTemplateFound {` one level below is because I wanted to output a meaningful `Step skipped` message. So it's done intentionally

pkg/reconciler/instances/eventing/preaction/handletogglejsfilestorage.go
VladislavPaskar: Yes, it can be deleted and it's being deleted together with the `PV`. I tested it as a part of this ticket.

mfaizanse (approved):

mfaizanse (approved):

ruanxin (approved):

Issue #949, opened 2022-03-17
Enhance the KCP CLI to enable debug log for particular operations during runtime


khlifi411:
**Description**

To simplify debugging, it has to be possible to enable the debug-mode for particular operations happening on a cluster.

It should be possible to enable the debugging via KCP CLI for a reconciliation (in this case all operations will automatically get the debug-mode enabled) or for a particular operation. When the component reconciler picks up an operation, the debug mode has to be enabled and the logger prints debug messages in addition to the usual log output.

AC:
* Enhance KCP CLI to enable debug-mode for a reconciliation (all related operations will run in debug mode)
* Enhance KCP CLI to enable debug-mode for an operation (only during the reconciliation of the particular operation will the reconciler also print debug messages)

**Reasons**

Enable debugging possibilities during runtime.

**Related issues**

Depends on #728

PR #1087, opened 2022-05-11
Enable debug logs for KCP CLI

khlifi411:
**Changes proposed in this PR**

- changed debug endpoints HTTP method from POST to PUT as they are idempotent operations
- changed error code when debug logs feature is disabled
- updated open API external spec for KCP CLI

**Related Issues**
#949

### DIFF
--- a/cmd/mothership/mothership/start/http.go
+++ b/cmd/mothership/mothership/start/http.go
@@ -181,12 +181,12 @@ func startWebserver(ctx context.Context, o *Options) error {
 	apiRouter.HandleFunc(
 		fmt.Sprintf("/v{%s}/operations/{%s}/{%s}/debug", paramContractVersion, paramSchedulingID, paramCorrelationID),
 		callHandler(o, enableOperationDebugLogging)).
-		Methods(http.MethodPost)
+		Methods(http.MethodPut)
 
 	apiRouter.HandleFunc(
 		fmt.Sprintf("/v{%s}/reconciliations/{%s}/debug", paramContractVersion, paramSchedulingID),
 		callHandler(o, enableReconciliationDebugLogging)).
-		Methods(http.MethodPost)
+		Methods(http.MethodPut)
 
 	apiRouter.HandleFunc(
 		fmt.Sprintf("/v{%s}/clusters/{%s}/config/{%s}", paramContractVersion, paramRuntimeID, paramConfigVersion),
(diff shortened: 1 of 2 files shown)


openapi/external_api.yaml
```diff
@@ -58,6 +58,34 @@ paths:
                 $ref: '#/components/schemas/HTTPErrorResponse'
         '500':
           $ref: '#/components/responses/InternalError'
+  /operations/{schedulingID}/{correlationID}/debug:
```
medmes, line 61: I think it will be better if you set debug mode as parameter in order to follow REST practices:
/operations/{schedulingID}/{correlationID}/**?debug=1**
The default value is 0, What do you think?

openapi/external_api.yaml
```diff
@@ -77,6 +105,29 @@ paths:
         "500":
           $ref: "#/components/responses/InternalError"
 
+  /reconciliations/{schedulingID}/debug:
```
medmes, line 108: Here as well

cmd/mothership/mothership/start/http.go
```diff
@@ -238,7 +237,8 @@ func startWebserver(ctx context.Context, o *Options) error {
 func enableReconciliationDebugLogging(o *Options, w http.ResponseWriter, r *http.Request) {
 
 	if !features.Enabled(features.DebugLogForSpecificOperations) {
-		w.WriteHeader(http.StatusNotFound)
+		err := fmt.Errorf("enabling debug logs for a reconciliation is disabled")
```
medmes, line 240: In my opinion , it's better to use `errors.New` instead of` fmt.Errorf `because the error in this line is not a custom error message.

medmes:
Except the suggestions, all looks good for me 👍

cmd/mothership/mothership/start/http.go
khlifi411: Yes. that's definitely better. I have applied it! thanks for the suggestion 👍

openapi/external_api.yaml
khlifi411: That's a good idea. But I am not sure if this applies in this case. Because, debug as a url param seems to me like I can disable debug looging for an operation or a reconciliation, which is not true and not needed by this feature because the operation will run once and the next operation debug logs will need to be enabled separately.

openapi/external_api.yaml
khlifi411: replied to the first comment above

openapi/external_api.yaml
medmes: Hmm, understood, then it make sense to not change it then :)

jeremyharisch (approved):

Issue #1056, opened 2022-04-26
Audit Log Shipment logs too much data


jakobmoellersap:
**Description**

When writing Audit Logs in the reconciler, we currently use request-based Middleware Audit Logging. While this takes care to cover all the details in a request, it can lead to unintended side-effects like logging the kubeconfig on the cluster creation endpoint when the audit data should be sanitized. We should switch to application-layer logging instead and have tighter control over audit log shipment.

**Reasons**

- Have more finegrained control over content of Audit Logs

**Attachments**

PR #1091, opened 2022-05-12
redact audit log details by key

jakobmoellersap:
fix https://github.com/kyma-incubator/reconciler/issues/1056


adityabhatia (approved):
💯

PR #1092, opened 2022-05-16
Increase payload limit to 100KB

ruanxin:


cmd/mothership/mothership/start/http.go
```diff
@@ -127,11 +127,6 @@ func startWebserver(ctx context.Context, o *Options) error {
 		callHandler(o, createOrUpdateCluster)).
 		Methods(http.MethodPost, http.MethodPut)
 
-	apiRouter.HandleFunc(
```
jakobmoellersap, line 130: This for sure is not intended right?

jakobmoellersap (changes requested):

cmd/mothership/mothership/start/http.go
ruanxin: actually intended, if you check the block above, it's duplicated. I'm really questioning the review process of other team...

jakobmoellersap (approved):

Issue #1107, opened 2022-06-08
Connectivity Proxy reconciliation fails for SKRs without Service Catalog and BTP Operator


wozniakjan:
**Description**

Kyma has deprecated service catalog https://github.com/kyma-project/kyma/issues/13422, and it is getting removed completely in https://github.com/kyma-project/kyma/pull/13675. BTP-Operator is also going to be optional component https://github.com/kyma-project/control-plane/issues/1755 and there can be an SKR without it. In that case we should either disable connectivity-proxy or make it installable without both service catalog as well as BTP-Operator.

**Expected result**

cluster provisioning and kyma installation should not fail when both service catalog and BTP-Operator are missing.

**Actual result**

cluster provisioning and kyma installation fail when both service catalog and BTP-Operator are missing.

```
"Runner: failing reconciliation of 'connectivity-proxy' in version '' with profile 'Production': Error while retrieving binding from BTP Operator: Error while listing resources: no matches for /, Resource=serviceinstance"
```

**Steps to reproduce**

spin up SKR / kyma from PR without service catalog and BTP-Operator

```
curl -XPUT "${keb-api}/oauth/v2/service_instances/$instance?accepts_incomplete=true" -i \
    -H "X-Broker-API-Version: 2.14" \
    -H "Content-Type: application/json" \
    --data-binary @- << EOF
{
   "service_id":"47c9dcbf-ff30-448e-ab36-d3bad66ba281",
   "plan_id":"361c511f-f939-4621-b228-d0fb79a1fe15",
   "context":{
       "globalaccount_id":"e449f875-b5b2-4485-b7c0-98725c0571bf",
       "subaccount_id":"${subaccount}",
       "user_id":"${user}",
   },
   "parameters":{
       "name":"$instance",
       "kymaVersion":"PR-13675",
       "overridesVersion":"2.2.0"
   }
}
EOF
```

**Troubleshooting**

janmedrek:
Waiting for confirmation from @wozniakjan if the fix solved all the issues. We will close the issue if there are no further problems.

wozniakjan:
thanks @janmedrek, just confirmed connectivity-proxy it appears operational on develop
```
COMPONENT               CORRELATION_ID                                                               SCHEDULING_ID                                                                PRIORITY   STATE         CREATED AT            UPDATED               REASON                                                                                 
[...]

PR #1108, opened 2022-06-08
Enable reconciliation without Service Catalog and BTP Operator

janmedrek:
Changes introduced:
- Ignore the noMatchErrors for the Bindings 
 
This was done to not fail the reconciliation loop if the Custom Resource Definitions are missing. If there are neither the BTP nor SC ones the Connectivity Proxy will simply not be installed.

Related: https://github.com/kyma-incubator/reconciler/issues/1107


koala7659 (approved):
LGTM

PR #1111, opened 2022-06-10
Use always Kyma Repository as default

tobiscr:


pkg/reconciler/service/reconciler.go
```diff
@@ -97,22 +97,22 @@ func (r *ComponentReconciler) EnableDryRun(dryRun bool) {
 	r.dryRun = dryRun
 }
 
-func (r *ComponentReconciler) newChartProvider(repo *reconciler.Repository) (*chart.DefaultProvider, error) {
-	wsFact, err := r.workspaceFactory(repo)
+func (r *ComponentReconciler) newChartProvider(_ *reconciler.Repository) (*chart.DefaultProvider, error) {
```
ruanxin, line 100: can we remove this `reconciler.Repository` as param if not needed?

ruanxin (approved):

PR #1113, opened 2022-06-21
updated containerd version to fix sec-vuln

khlifi411:
**Description**
Updated containerd to latest

**Related issue(s)**

https://github.tools.sap/kyma/security-scans/issues/18094


jakobmoellersap (changes requested):
I think the indirects are not updated, so its only done in the replace. I think we should also update the indirects

jakobmoellersap (approved):

PR #1118, opened 2022-06-24
Stop retry if err is no such host

ruanxin:
no such host error log can be found in this comment: https://github.tools.sap/kyma/backlog/issues/2654#issuecomment-1064333

### DIFF
--- a/pkg/reconciler/service/runner.go
+++ b/pkg/reconciler/service/runner.go
@@ -107,6 +107,13 @@ func (r *runner) Run(ctx context.Context, task *reconciler.Task, callback callba
 		retry.Attempts(uint(task.ComponentConfiguration.MaxRetries)),
 		retry.Delay(r.retryDelay),
 		retry.LastErrorOnly(false),
+		retry.RetryIf(func(err error) bool {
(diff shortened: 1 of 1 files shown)


ruanxin:
Hi @jakobmoellersap yes, regarding the temporary network outage, we have this concern and discussed. The purpose of this issue is related to previous [release blocker](https://github.tools.sap/kyma/backlog/issues/2654), we skip this error retry, try to fail faster because of non exists target clusters. 

We have default configured 15 times delete retry until one cluster consider as delete error, and 150 times retry until one cluster consider as reconcile error.
```
	defaultMaxReconcileErrRetries  = 150
	defaultMaxDeleteErrRetries     = 15
```
So if it's temporary network outage, there is highly chance in next several retries, it will back to normal.

And some background about previous issue, due to some unexpected sync problem, many target clusters (around 1k) in DEV has been removed by external service, but not synced in reconciler DB, it took several hours for reconciler workers to mark them as delete error, and it leads to only a few workers can be assigned to normal cluster reconciliation, which leaves the impression that new provisioned cluster is not installing kyma.

PR #1121, opened 2022-07-04
Remove ory finalizers during reconciler delete action

werdes72:
It solves the issue with race condition during kyma deprovisioning.


pkg/reconciler/instances/ory/k8s/ory_crds.go
```diff
@@ -0,0 +1,143 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	k8sRetry "k8s.io/client-go/util/retry"
+)
+
+type OryCRDsHandler interface {
+	FindAndDeleteOryFinalizers(kubeconfigData string, logger *zap.SugaredLogger) error
```
tobiscr, line 20: Interfaces are actually only needed if you plan to provide multiple implementations / allow others to come up with a custom implementation. Otherwise it's cleaner to leave them away and introduce them later when you have a concrete need.

tobiscr (approved):

pkg/reconciler/instances/ory/k8s/ory_crds.go
Tomasz-Smelcerz-SAP: In addition to what @tobiscr wrote, there are also some unexported methods in the interface? That's a bit unusual coding style. Any reason for that?

werdes72: I used the interface to match with other ory component code. Unexported methods are removed.

pkg/reconciler/instances/ory/k8s/ory_crds.go
```diff
@@ -0,0 +1,133 @@
...
+	h.logger.Debugf("Dropping finalizers for all ory custom resources of type: %s.%s/%s", crdef.Resource, crdef.Group, crdef.Version)
+	defer h.logger.Debugf("Finished dropping finalizers for ory custom resources of type: %s.%s/%s", crdef.Resource, crdef.Group, crdef.Version)
+
+	customResourceList, err := h.dynamic.Resource(crdef).Namespace(v1.NamespaceAll).List(context.Background(), metav1.ListOptions{})
+	if err != nil && !apierr.IsNotFound(err) {
+		return err
+	}
+
+	if customResourceList == nil {
+		return nil
```
veichtj, line 84: Maybe some debug log would be good here as well

pkg/reconciler/instances/ory/k8s/ory_crds.go
veichtj: Interfaces are typically used in the ory reconciler code to allow for easier mocking. So like @tobiscr said, if you don't  use that mechanism you could introduce it later as well

pkg/reconciler/instances/ory/k8s/ory_crds.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	for i := range customResourceList.Items {
+		instance := customResourceList.Items[i]
```
veichtj, line 88: once more ;-)

tobiscr (approved):

veichtj (approved):

PR #1139, opened 2022-07-26
Only patch new istio webhook

veichtj:
Remove logic to patch either the old istio webhook (<1.12) or the new one. Just rely on new existing one.


dariusztutaj (approved):
/meow

PR #1151, opened 2022-08-09
Add an eventing preaction to handle the Helm charts update

VladislavPaskar:
**Changes in this Pull Request**

- `SERVER_NAME` env var could act as an indicator for updating the NATS helm charts from existing version to the latest `0.17.3` version
- the changes in the charts touch the StatefulSet Spec, which requires the recreation of the STS
- the PVC stays untouched, which means: `no messages in the stream will be lost, in case there are any`

**Related Issue**
https://github.com/kyma-project/kyma/issues/14961


VladislavPaskar:
not required anymore.

PR #1152, opened 2022-08-10
Revert Only patch new istio webhook(#1139)

cnvergence:
Bring back support for older clusters with only istio-sidecar-injector mutatingwebhook


strekm (approved):
/meow

PR #1157, opened 2022-08-17
breaking change: upgrade to go 1.19 - Protecode vulnerabilities

adityabhatia:
This should solve all protecode security vulnerabilities associated with golang versions 1.17.x and 1.18.x


adityabhatia:
closing this PR since test-infra doesn't support 1.19 as of now

PR #1160, opened 2022-08-19
Add FinalizerIntercepter

chrkl:
The FinalizerIntercepter adds existing Kubernetes finalizers to rendered
manifests of LogPipeline and LogParser kind to avoid them to be removed in every reconciliation cycle.

See also https://github.com/kyma-project/kyma/issues/15118


jakobmoellersap (approved):

tobiscr (approved):

pkg/reconciler/service/install.go
```diff
@@ -59,6 +59,13 @@ func (r *Install) Invoke(ctx context.Context, chartProvider chart.Provider, task
 			},
 			newClusterWideResourceInterceptor(),
 			&NamespaceInterceptor{},
+			&FinalizerInterceptor{
```
akgalwas, line 62: Removing finalizers by Reconciler can cause issues for components other than  `logging`.  It would make sense to apply `FinalizerInterceptor`for all resource types.

akgalwas (approved):

PR #1163, opened 2022-08-22
Add preaction to delete configMaps no longer used in eventing

raypinto:
As part of the preaction we delete the following two configMaps:
- "eventing-dashboards-event-types-summary"
- "eventing-dashboards-delivery-per-subscription"


Ressetkk:
/override pre-main-reconciler-cli-kyma-prev-to-last-release-upgrade-k3d

Issue #1159, opened 2022-08-18
Support Istio injection pod labels in Proxy Reset


cnvergence:
**Description**
With some changes to Istio sidecar injection logic, there is still some lack of consistency in the docs
Example:
- https://istio.io/latest/docs/ops/configuration/mesh/injection-concepts
- https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy

It seems like this logic is up to date:
```
The injector is configured with the following logic:
If either label is disabled, the pod is not injected
If either label is enabled, the pod is injected
If neither label is set, the pod is injected if .values.sidecarInjectorWebhook.enableNamespacesByDefault is enabled. This is not enabled by default, so generally, this means the pod is not injected.
```
This is meaning that with the latest versions of Istio `sidecar.istio.io/inject` label will take precedence over `sidecar.istio.io/inject ` annotation on the pod and we should support it in our Proxy Reset as well.

**Reasons**
With Istio injection change to disabled by default, we would need to support the labels as well for opt-in/opt-out in the service mesh.
Sidecar annotation is reported to cause misunderstanding how the sidecar injection is working under the hood:
- https://github.com/istio/istio/issues/6476
- https://github.com/istio/istio/issues/7404

cnvergence:
Current sidecar injection behaviour:

| default policy (enableNamespacesByDefault) | namespaceSelector (label: "istio-injection"]) | Pod override annotation (sidecar.istio.io/inject) | Sidecar injected? |
|:------------------------------------------:|:---------------------------------------------:|:-------------------------------------------------:|:-----------------:|
|            true	            		             |                 enabled						                 |                    true 						                    |      yes				      |
|                  true			                   |                 enabled						                 |                    false						                    |      no				       |
|                true     		                 |                disabled						                 |                    true						                     |      no				       |
|                  true			                   |                disabled						                 |                    false						                    |      no				       |
|                  true			                   |                  not present                  |                    true						                     |      yes				      |
|                  true			                   |                not present				                |                    false						                    |      no				       |
|                  true			                   |                not present				                |                  not present 			                  |      yes				      |
|                 false   		                 |                 enabled						                 |                    true 						                    |      yes				      |
|                 false   		                 |                 enabled						                 |                    false						                    |      no				       |
|                 false   		                 |                disabled						                 |                    true						                     |      no				       | 
|                 false   		                 |                disabled						                 |                    false						                    |      no				       |
[...]

PR #1176, opened 2022-09-07
Support sidecar injection labels in proxy reset

cnvergence:
See also: 
- https://github.com/kyma-incubator/reconciler/issues/1159


barchw (approved):
/meow

Issue #1173, opened 2022-09-02
Istio reconciler support istio 1.15.x


strekm:
**Description**

Implement necessary changes to install / upgrade istio in 1.15.. version. 

- [x] istioctl 1.15.x is used for installation / upgrade
- [x] gardener cluster can be hibernated
- [x] istio running in 1.15.x version
- [x] sidecars running in 1.15.x version
- [x] reconciler bumped in kyma cli
- [x] reconciler bumped for managed kyma

**Reasons**

Keeping Istio up to date

**Attachments**
https://github.com/kyma-project/kyma/issues/15347

PR #1184, opened 2022-09-15
Add istioctl 1.15.0 to Dockerfile

werdes72:
See: https://github.com/kyma-incubator/reconciler/issues/1173

### DIFF
--- a/Dockerfile.cr
+++ b/Dockerfile.cr
@@ -3,7 +3,7 @@ FROM eu.gcr.io/kyma-project/external/istio/istioctl:1.14.4 AS istio-1_14_4
 FROM eu.gcr.io/kyma-project/external/istio/istioctl:1.15.0 AS istio-1_15_0
 
 # Build image
-FROM golang:1.18.5-alpine3.16 AS build
+FROM golang:1.19.1-alpine3.16 AS build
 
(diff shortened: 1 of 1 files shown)


Dockerfile.cr
```diff
@@ -1,5 +1,6 @@
 # Istioctl source images
 FROM eu.gcr.io/kyma-project/external/istio/istioctl:1.14.4 AS istio-1_14_4
+FROM eu.gcr.io/kyma-project/external/istio/istioctl:1.15.0 AS istio-1_15_0
 
 # Build image
 FROM golang:1.18.5-alpine3.16 AS build
```
barchw, line 6:
```suggestion
FROM golang:1.19.1-alpine3.15 AS build
```

It seems that Istio uses some functionality that was introduced in Go v.1.19

- https://pkg.go.dev/sync/atomic@go1.19.1#Bool
- https://pkg.go.dev/sync/atomic@go1.18.6#Value

barchw (changes requested):

Dockerfile.cr
werdes72: Good catch!

barchw (approved):

tobiscr (approved):

PR #1188, opened 2022-09-22
Fix linting issues

tobiscr:


clebs (approved):

ruanxin (approved):

cnvergence (approved):

clebs (approved):

franpog859 (approved):

pkg/db/columnhandler.go
```diff
@@ -154,7 +154,7 @@ func (ch *ColumnHandler) ColumnName(field string) (string, error) {
 	return "", fmt.Errorf("entity '%s' has no field '%s': cannot resolve column name", ch.entity, field)
 }
 
-//ColumnNamesCsv returns the CSV string of the column names
+// ColumnNamesCsv returns the CSV string of the column names
```
VOID404, line 157: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
```suggestion
// ColumnNamesCSV returns the CSV string of the column names
```

pkg/db/columnhandler_test.go
```diff
@@ -161,7 +161,7 @@ func splitAndTrimCsv(csv string) []string {
 	return result
 }
 
-//mock DB entity used to test the validation logic
+// mock DB entity used to test the validation logic
```
VOID404, line 164: https://go.dev/doc/comment
> [...] doc comments for types start with complete sentences naming the declared symbol. An explicit subject often makes the wording clearer, and it makes the text easier to search, whether on a web page or a command line

VOID404:
As per [the style guide](https://go.dev/doc/comment), Go doc comments should use complete sentences ending with a dot

kwiatekus (approved):

pkg/db/columnhandler_test.go
tobiscr: Fully agree on your points. But a fix would need to change actually all files which use comments to be consistent again. 

As we are in maintenance mode and not changing the old reconciler any more (replacement happens in coming months), we should avoid bigger code changes as the efforts will get lost soon.

Please excuse when I recommend to leave the wrong comment style and focus on clean comments in the new reconciler code bases.

pkg/db/columnhandler.go
tobiscr: Agree on your points. A fix would imply to change the wrong fct-name as well to have it consistant. Please see my comments below, why we try to minimise any changes in the old reconciler (we are close to the transition phase and will drop the old reconciler in next few months).

ebensom (approved):

Issue #1295, opened 2023-02-27
Cleaner: Use only CRDs with `storage` = `true` 


zupa-pomidorowa:
**Description**

Currently, to find the objects to remove, the cleaner iterates over all defined versions of a CRD.
For example: `v1alpha1`, `v1beta1`, `v1` etc.

The cleaner then uses every CRD version to `list` the objects and then tries to remove them.
That approach breaks if there is not a Conversion Webhook anymore, which is often the case during cleanup phase.
This is because the Kyma component, including the controller (operator), along with the defined Conversion Webhook may have been undeployed already  (see: https://github.com/kyma-incubator/reconciler/pull/1193).

To be able to `list` objects without the Conversion Webhook, the cleaner should only list objects from the CRD version marked with `storage`  = `true`.
**This requires that the same version, and only that version, is listed in the CRD's `status.storedVersions`** field.

**NOTE**
_Proper migration of the CRD versions is beyond the scope of the cleaner._
As [explained in the docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version) it is normal that more than one version is actually stored during API migration, as reflected in the CRD's `status.storedVersions` field.
The cleaner, if invoked in a two-stored-versions scenario, will probably fail!
It is because there are still some CR instances stored using the old version, but the cleaner will try to access these using the new version, marked with `storage` = `true`. And there's no Conversion Webhook available.

**Expected result**

Cleaner doesn't try to `list` objects from an old CRD version.

**Actual result**

Cleaner fails to `list` objects from an old CRD version, because there's no  Conversion Webhook anymore.

**Steps to reproduce**

TBD.

**Troubleshooting**
n/a

PR #1296, opened 2023-02-27
CRD filter by 'storage==true'

Tomasz-Smelcerz-SAP:
This PR ensures that only CRD versions marked with `storage` == `true` are used to find instances of CRs to clean-up.
The existing approach is using **all** CRD versions, which leads to failures if the Conversion Webhook does not exist anymore.
See: https://github.com/kyma-incubator/reconciler/issues/1295 for details.

**How to test:**

To reproduce:

1) `kyma provision k3d`
2) `kyma deploy -v`
3) create two ApiRules in v1beta1 (you can find sample [here](https://github.com/kyma-project/api-gateway/blob/main/config/samples/gateway_v1beta1_apirule.yaml))
4) `kyma undeploy -v`

You should observe, there is an error related to the `v1alpha1` version of the ApiRules, as reported in the linked issue.

To verify:

Build the kyma-cli using reconciler from this PR
Repeat the steps
Observe that now only the `v1beta1` version of the ApiRule CRD is involved in the cleanup, and there is no error.

Fixes: https://github.com/kyma-incubator/reconciler/issues/1295

### DIFF
--- a/pkg/reconciler/instances/cleaner/action.go
+++ b/pkg/reconciler/instances/cleaner/action.go
@@ -96,13 +96,24 @@ func findKymaCRDs(crdManifests []*chart.Manifest, logger *zap.SugaredLogger) ([]
 					if !ok {
 						return nil, fmt.Errorf("Error getting attribute \"spec.versions[%d].name\" for %s CRD", i, crdName)
 					}
-					grv := toGRV(group, versionName, namesPlural)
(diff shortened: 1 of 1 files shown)


Tomasz-Smelcerz-SAP:
putting on hold until manually verified...

Tomasz-Smelcerz-SAP:
Verified manually on a `k3d` cluster.

tobiscr (approved):

PR #1307, opened 2023-03-02
Istio reconciler reset proxy more resielient, verify Istio install

videlov:
- Add Istio install verification step to check if istio-system pods are running with the new target version after installation.
- Add additional check before resetting proxies, checks if current Istio version is the target version for proxy reset.


jeremyharisch (dismissed):

pkg/reconciler/instances/istio/action.go
```diff
@@ -274,13 +276,46 @@ func canUpdate(istioStatus actions.IstioStatus) (bool, error) {
...
+	for dpVersion := range istioStatus.DataPlaneVersions {
+		if isDataplaneCompatible, err := isComponentCompatible(dpVersion, istioStatus.TargetVersion, "Data plane"); !isDataplaneCompatible {
+			return false, err
+		}
+	}
+
+	return true, nil
+}
+
+func canResetProxies(istioStatus actions.IstioStatus) (bool, error) {
```
cnvergence, line 288: do we need to return bool here?

pkg/reconciler/instances/istio/action.go
```diff
@@ -139,15 +141,15 @@ func (a *ProxyResetPostAction) Run(context *service.ActionContext) error {
 		return err
 	}
 
-	canUpdateResult, err := canUpdate(istioStatus)
+	_, err = canResetProxies(istioStatus)
```
cnvergence, line 144: can we somehow use this bool?

cnvergence:
LGTM after fixes, just left small comments

pkg/reconciler/instances/istio/action.go
videlov: actually no need for the bool, indeed.

pkg/reconciler/instances/istio/action.go
videlov: removed it.

cnvergence (approved):

jeremyharisch (approved):

PR #1311, opened 2023-03-09
Clean up duplicated webhooks when istioctl fails

cnvergence:
- Clean up conflicted webhooks when istioctl fails to unblock reconciliations.

See also: 
- https://github.com/kyma-project/kyma/issues/17054


pkg/reconciler/instances/istio/webhooks/webhooks.go
```diff
@@ -0,0 +1,82 @@
...
+func isDefaultRevisionDeactivated(ctx context.Context, client kubernetes.Interface) bool {
+	webhooks, err := tag.GetWebhooksWithRevision(ctx, client, tag.DefaultRevisionName)
+	if err != nil {
+		return true
+	}
+	if len(webhooks) == 0 {
+		return true
+	}
+
+	webhook := webhooks[0]
```
triffer, line 73: Why can we rely on the order of the webhooks and always check the first webhook? As far as I understand this `GetWebhooksWithRevision` will return `istio-sidecar-injector` and `istio-revision-tag-default` and we only want to verify that all webhooks in the `istio-sidecar-injector` webhook configuration are deactivated. Should we add a filter for the name to make it more clear?

pkg/reconciler/instances/istio/webhooks/webhooks_test.go
```diff
@@ -0,0 +1,125 @@
...
+		// then
+		require.NoError(t, err)
+		got, err := client.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(ctx, taggedWhName, metav1.GetOptions{})
+		require.NoError(t, err)
+		require.Equal(t, taggedWh, got)
+	})
+	t.Run("should delete conflicted tagged webhook if old one is not deactivated", func(t *testing.T) {
+		// given
+		defaultWh := createDefaultMutatingWebhookWithSelector(defaultWhName, "istio.io/rev", validSelector)
+		taggedWh := createDefaultMutatingWebhookWithSelector(taggedWhName, tag.IstioTagLabel, validSelector)
```
triffer, line 84: We might want to update the test setup as both webhook configurations are created with the label "istio.io/rev" with the value "default".

pkg/reconciler/instances/istio/webhooks/webhooks.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	err = retry.Do(func() error {
+		webhooks, err := tag.GetWebhooksWithTag(ctx, kubeClient, tag.DefaultRevisionName)
+		if err != nil {
+			return err
+		}
+
+		if !isDefaultRevisionDeactivated(ctx, kubeClient) && len(webhooks) > 0 {
+			for _, wh := range webhooks {
```
triffer, line 46: Maybe we can add a comment like this to make it more clear why we can delete this webhooks:
"As the default revision is not deactivated and handles the injection, we are safe to delete all other webhook configurations that were created during the failed installation."

pkg/reconciler/instances/istio/webhooks/webhooks_test.go
```diff
@@ -0,0 +1,125 @@
...
+		client := fake.NewSimpleClientset(defaultWh, taggedWh)
+		provider := &clientsetmocks.Provider{}
+		provider.On("RetrieveFrom", mock.AnythingOfType("string"), mock.AnythingOfType("*zap.SugaredLogger")).Return(client, nil)
+		// when
+		err := DeleteConflictedDefaultTag(ctx, provider, kubeConfig, log)
+
+		// then
+		require.NoError(t, err)
+		got, err := client.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(ctx, taggedWhName, metav1.GetOptions{})
```
triffer, line 93: Adding an assert that the default webhook is not deleted might be a good idea. What do you think?

pkg/reconciler/instances/istio/webhooks/webhooks.go
cnvergence: I will add the comment

pkg/reconciler/instances/istio/webhooks/webhooks_test.go
cnvergence: oh, I think I missed that information

pkg/reconciler/instances/istio/webhooks/webhooks.go
```diff
@@ -0,0 +1,82 @@
...
+		return true
+	}
+	if len(webhooks) == 0 {
+		return true
+	}
+
+	webhook := webhooks[0]
+	for i := range webhook.Webhooks {
+		wh := webhook.Webhooks[i]
+		if fmt.Sprint(wh.NamespaceSelector.MatchLabels) == fmt.Sprint(deactivatedLabel) || fmt.Sprint(wh.ObjectSelector) == fmt.Sprint(deactivatedLabel) {
```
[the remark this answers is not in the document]
cnvergence: tbh, any occurence of never-match label would be enough to count that MWC as deactivated
I can add AND with no problem at all.

pkg/reconciler/instances/istio/actions/performer.go
```diff
@@ -197,6 +198,11 @@ func (c *DefaultIstioPerformer) Install(context context.Context, kubeConfig, ist
 
 	err = commander.Install(mergedCNI, kubeConfig, logger)
 	if err != nil {
+		// In case of error in the istioctl, old mutatingwebhook won't be deactivated, which will block later reconciliations.
+		err2 := webhooks.DeleteConflictedDefaultTag(context, c.provider, kubeConfig, logger)
+		if err2 != nil {
```
triffer, line 203: If we don't want to wrap err2 into err I think we should at least log err2 on error level.

triffer (approved):

ruanxin (approved):

Ressetkk:
As per https://github.com/kyma-project/kyma/issues/17030#issuecomment-1471915673:
> The issue has been fixed on main. Currently we will override the result of the job to be green. As the fix has been already backported to release branch 2.12. Check [here](https://github.com/kyma-project/kyma/commit/86447847e0919cb1062dfa1a98e6c3b8ca4d6bcb). Once we release 2.12.1 then the job should work fine and we should not need to override the result anymore.

/override pre-main-reconciler-cli-kyma-prev-to-last-release-upgrade-k3d

Ressetkk:
/override rakesh-garimella_test_of_prowjob_pre-main-reconciler-cli-kyma-p

PR #1312, opened 2023-03-10
Add migration logic for transition from old to new reconciler

tobiscr:
The old reconciler is checking whether a component has a module-CRD on the target cluster. If this is the case, the component is expected to be managed by the new reconciler and the old reconciler will exclude it from the reconciliation.


tobiscr:
PR blocked by  https://github.com/kyma-project/kyma/issues/17030

janmedrek (approved):

jeremyharisch (approved):

PR #1322, opened 2023-03-24
Add a watcher on pods when deleting the NATS sts.

friedrichwilken:
To avoid a deadlock when deleting and recreating the NATS StatefulSet we have to make sure that the Pods are deleted as well.

### DIFF
--- a/pkg/reconciler/instances/eventing/preaction/handlenatspodmanagementpolicy.go
+++ b/pkg/reconciler/instances/eventing/preaction/handlenatspodmanagementpolicy.go
@@ -2,6 +2,7 @@ package preaction
 
 import (
 	"fmt"
+
 	"github.com/kyma-incubator/reconciler/pkg/reconciler/chart"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/instances/eventing/preaction/handlenatspodmanagementpolicy.go
```diff
@@ -125,18 +128,35 @@ func (r *handleNATSPodManagementPolicy) Execute(context *service.ActionContext,
...
+	// Fetch a list of all Pods as we need to make sure they are deleted as well.
+	listOpts := metav1.ListOptions{LabelSelector: podLabel}
+	pods, err := clientSet.CoreV1().Pods(namespace).List(ctx.Context, listOpts)
+	if err != nil {
+		return err
+	}
+
+	// Watch all Pods.
+	for _, pod := range pods.Items {
+		if err := addToProgressTracker(tracker, logger, podType, pod.GetName()); err != nil {
```
mfaizanse, line 142: How does the tracker figure out the `namespace` of the provided pods?

PR #1330, opened 2023-04-05
Use config file for disabled linters & delete hack/lint.sh

lindnerby:
Description:
* Lint step was deactivated temporarily, so activate it again here
* Move all linters config to the config file
* Move config file to root
* Delete hack/lint.sh
* Adapt for new lint job
* Fix make target lint for local linting
* Linting on pipelines will not download golangci-lint


lindnerby:
Need to re-open since it's not picking up pipeline changes.

PR #1353, opened 2023-06-01
Update AWS LB on Istio GW connection timeout settings via annotation

videlov:
Update AWS LB on Istio GW connection timeout settings via annotation

https://github.com/kyma-project/kyma/issues/17496


triffer (commented):
Why do we need this implementation? In my tests on SKR and locally executing reconciler against a Gardener cluster, the reconciler didn't remove the annotation. Did you experience something different?

It's not the right behaviour in the long term, but we should fix the behaviour that annotations are not reverted in istio module.

videlov:
> Why do we need this implementation? In my tests on SKR and locally executing reconciler against a Gardener cluster, the reconciler didn't remove the annotation. Did you experience something different?
> 
> It's not the right behaviour in the long term, but we should fix the behaviour that annotations are not reverted in istio module.

perhaps it was something wrong with my setup, I had to apply it once again. let me recheck this.

videlov:
no need for explicit setting annotations

PR #1359, opened 2023-06-14
Resolve node port conflict for internal docker registrey

dbadura:
Changes:
- add action which solves the node port conflict.

### DIFF
--- a/pkg/reconciler/instances/serverless/nodeport.go
+++ b/pkg/reconciler/instances/serverless/nodeport.go
@@ -1,0 +1,155 @@
+package serverless
+
+import (
+	"context"
+	"fmt"
+	"github.com/kyma-incubator/reconciler/pkg/reconciler/service"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"math/rand"
+)
+
+var _ service.Action = ResolveDockerRegistryNodePort{}
+
+const (
+	dockerRegistryNodePort = 32_137
+
+	//Available ports according to documentation https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport
+	maxNodePort = 32_767
+	minNodePort = 30_000
+)
+
+const (
+	dockerRegistryService      = "serverless-docker-registry"
+	dockerRegistryNodePortPath = "global.registryNodePort"
+	dockerRegistryPortName     = "http-registry"
+)
+
+type ResolveDockerRegistryNodePort struct {
+}
+
+func (n ResolveDockerRegistryNodePort) Run(svcCtx *service.ActionContext) error {
+	k8sClient, err := svcCtx.KubeClient.Clientset()
+	if err != nil {
+		return errors.Wrap(err, "while getting clientset")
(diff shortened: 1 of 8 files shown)


pkg/reconciler/instances/serverless/nodeport.go
```diff
@@ -0,0 +1,155 @@
...
+	return minRange + number
+}
+
+func setNodePortOverride(overrides map[string]interface{}, path string, port int32) {
+	overrides[path] = port
+}
+
+func getAllNodePortServices(ctx context.Context, k8sClient kubernetes.Interface, namespace string) (*corev1.ServiceList, error) {
+
+	svcs, err := k8sClient.CoreV1().Services(namespace).List(ctx, metav1.ListOptions{})
```
dbadura, line 132: You should get all services from everynamespace

pkg/reconciler/instances/serverless/nodeport.go
```diff
@@ -0,0 +1,155 @@
...
+	if svc != nil {
+		if isDefaultNodePortValue(svc) {
+			return nil
+		}
+		currentNodePort := getNodePort(svc)
+		setNodePortOverride(svcCtx.Task.Configuration, dockerRegistryNodePortPath, currentNodePort)
+		return nil
+	}
+
+	svcs, err := getAllNodePortServices(svcCtx.Context, k8sClient, svcCtx.Task.Namespace)
```
kwiatekus, line 53: This should look up node port svc from all namespaces.

kwiatekus (changes requested):

pkg/reconciler/instances/serverless/nodeport.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("while checking if %s service is installed on cluster", dockerRegistryService))
+	}
+
+	if svc != nil {
+		if isDefaultNodePortValue(svc) {
+			return nil
+		}
+		currentNodePort := getNodePort(svc)
+		setNodePortOverride(svcCtx.Task.Configuration, dockerRegistryNodePortPath, currentNodePort)
```
kwiatekus, line 49: shouldn't the new nodeport value be drawn in this case? I tested the code and it keeps applying the same nodeport ```
kubeClient failed to update Service 'serverless-docker-registry' (namespace: kyma-system)  with strategy 'PATCH': cannot patch "serverless-docker-registry" with kind Service: Service "serverless-docker-registry" is invalid: spec.ports[0].nodePort: Invalid value: 32137: provided port is already allocated
```

**Review by kwiatekus — CHANGES_REQUESTED** [review]

File: pkg/reconciler/instances/serverless/nodeport.go [hunk 4]
```diff
@@ -0,0 +1,155 @@
...
+	k8sClient, err := svcCtx.KubeClient.Clientset()
+	if err != nil {
+		return errors.Wrap(err, "while getting clientset")
+	}
+	svc, err := getService(svcCtx.Context, k8sClient, svcCtx.Task.Namespace, dockerRegistryService)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("while checking if %s service is installed on cluster", dockerRegistryService))
+	}
+
+	if svc != nil {
```
kwiatekus, line 44:
```suggestion
	if svc != nil && isNodePortType(svc) {
```

kwiatekus (approved):

PR #1364, opened 2023-06-21
Fix for generated too long common name in connectivity-proxy istio certificate

koala7659:
The `commonName` field in connectivity-proxy Istio certificate is built from concatenation of `cc-proxy` sub-domain name and SKR cluster host name. 

It is possible when shoot name is too long the resulting commonName string exceeds max allowed length of 64 characters. 

For example shooot `c-92461c5` on stage has a domain name `api.c-92461c5.kyma-stage.shoot.live.k8s-hana.ondemand.com`. In such a case the generated commonName `cc-proxy.api.c-92461c5.kyma-stage.shoot.live.k8s-hana.ondemand.com` is 67 characters long.

This PR should prevent this problem - it will use `cp` prefix for the sub-domain name instead of `cc-proxy`


pkg/reconciler/instances/connectivityproxy/action.go
```diff
@@ -142,6 +142,10 @@ func (a *CustomAction) Run(context *service.ActionContext) error {
 		return errors.Wrap(err, "Error during reconciliation")
 	}
 
+	if err := a.Commands.PatchConfigMap(context, "kyma-system", "connectivity-proxy"); err != nil {
```
koala7659, line 145: This function should not be called every reconcilation.  PatchConfigMap  must be done only if we are upgrading from 2.4.4 to 2.9.2. 

```suggestion
         if (refresh) {
                 if currentVersion == 2.4.4 && installedVersion == 2.9.2 {
	                if err := a.Commands.PatchConfigMap(context, "kyma-system", "connectivity-proxy"); err != nil {
	        }
	}
```

akgalwas: Done

m00g3n (dismissed):

m00g3n (approved):
LGTM

PR #1370, opened 2023-06-28
Remove api prefix from the tunnel's URL

akgalwas:
Gardener no longer allows to create certificate for API server subdomains. So if the dns name is of form cp.api.myserver.com, and api.myserver.com is API server domain, Gardener will fail to create a certificate. 
This PR contains the following changes:
- Changing Connectivity Proxy's tunnel address to not contain api prefix
- Fixing configuration for existent clusters


m00g3n (approved):
LGTM

koala7659 (approved):
LGTM

Issue #1379, opened 2023-08-03
Reconciler fails when processing PVCs with expanded size


kwiatekus:
**Description**
In case PVC was expanded ( say from 20 to 30 Gi ) on the runtime, the reconciler fails to reconcile it

**Expected result**
Reconciler should intercept for PVC resources and check the actual size before applying the PVC manifest

**Actual result**

`'kubeClient failed to update PersistentVolumeClaim ''serverless-docker-registry'' (namespace: kyma-system)  with strategy ''PATCH'': cannot patch "serverless-docker-registry" with kind PersistentVolumeClaim: PersistentVolumeClaim "serverless-docker-registry" is invalid: spec.resources.requests.storage: Forbidden: field can not be less than previous value',`

PR #1381, opened 2023-08-08
Allow expanded pvc reconciliations

kwiatekus:
**Description**
When reconciling PVCs and desired volume is less than actual one take the actual volume from the runtime in order to avoid reconciliation errors and incident reports.

**Related issues**
https://github.com/kyma-incubator/reconciler/issues/1379


tobiscr (approved):

tobiscr (approved):

m00g3n (approved):
LGTM

PR #1392, opened 2023-09-26
Hydra migration

triffer:
- Skip reconciliation for Hydra resources (e.g. Hydra database secret, Hydra synchronization, Hydra rollout) when Hydra is disabled or does not exist in chart config
- Create Hydra database secret `ory-hydra-credentials` in `hydra-deprecated` namespace if it exist, otherwise use `kyma-system`. The data source name in the secret is set depending on whether the namespace `hydra-deprecated` exists or not.

### DIFF
--- a/pkg/reconciler/instances/ory/action.go
+++ b/pkg/reconciler/instances/ory/action.go
@@ -47,7 +47,7 @@ type postDeleteAction struct {
 }
 
 const (
-	databaseSecretName = "ory-hydra-credentials"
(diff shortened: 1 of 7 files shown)


werdes72 (approved):

tobiscr (approved):

Ressetkk:
/override pre-main-ory-reconciler-prod-integration-gcp
As requested by @strekm

Issue #1386, opened 2023-08-20
Connectivity Proxy Reconciler should check the presence of Istio resources before installing Connectivity Proxy


koala7659:
**Description**

Connectivity Proxy installation requires Istio to to be installed on the Kyma Cluster.
After the moment when Istio Kyma module is officially enabled it might be possible that Istio might not be present on the Kyma Cluster where Connectivity Proxy is about to be installed. 
That's why before Istio module officially is enabled - we must add Istio pre check in Connectivity Proxy Reconciler as an installation condition.

**Suggested implementation:**

In reconciliation loop we will check for existence of required Istio CRD like Virtual Service and Gateway. If these CRDs are missing we will skip installation without error.

**Comments:**

This must be delivered before Istio module is officially enabled on Canary (in November).

tobiscr:
Closing issue as it got solved in https://github.com/kyma-incubator/reconciler/pull/1403

PR #1403, opened 2023-10-24
Skip reconcile when Istio CRDs are missing

koala7659:
Issue: https://github.com/kyma-incubator/reconciler/issues/1386

Check in reconciliation action if Istio Virtual Service or Gateway CRDs are missing on the cluster and in such a case skip the reconciliation.

### DIFF
--- a/pkg/reconciler/instances/connectivityproxy/action.go
+++ b/pkg/reconciler/instances/connectivityproxy/action.go
@@ -45,8 +45,8 @@ func (a *CustomAction) Run(context *service.ActionContext) error {
 	}
 	context.Task.Configuration["global.kubeHost"] = strings.TrimPrefix(host, "https://api.")
 
-	// checking Istio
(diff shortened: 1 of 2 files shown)


pkg/reconciler/instances/connectivityproxy/action.go
```diff
@@ -44,6 +45,12 @@ func (a *CustomAction) Run(context *service.ActionContext) error {
 	}
 	context.Task.Configuration["global.kubeHost"] = strings.TrimPrefix(host, "https://api.")
 
+	// checking Istio
+	context.Logger.Debug("Checking if Istio CRDs are present on the cluster")
+	if istioCRDsAreMissing(context) {
+		return nil
```
Disper, line 51: Would it make sense to log that the CRD is missing?

akgalwas (approved):

Issue #1409, opened 2023-11-20
Istio reconciler clean up 


strekm:
**Description**

Since Istio is modular all related implementation and configuration can be cleaned up.

ACs:
- [ ] no dependency to Istio
- [ ] no dependency to Kyma Istio
- [ ] no related configuration
- [ ] no related tests
- [ ] no related documentation

**Reasons**

Removal of not used components

**Attachments**

strekm:
already done

PR #1415, opened 2023-12-06
Remove Istio

triffer:
Istio module is rolled out to Prod, so we can remove the deprecated Istio reconciler.

- https://github.com/kyma-incubator/reconciler/issues/1409


werdes72:
How about this file: https://github.com/kyma-incubator/reconciler/blob/1bfc067e5d88e866b763ddfa089825f117b6bee1/configs/reconciler.yaml#L45

lindnerby (approved):

werdes72 (approved):

PR #1432, opened 2024-01-18
Cache kubeconfig lookups

tobiscr:


pkg/cluster/inventory.go
```diff
@@ -57,6 +65,19 @@ func NewInventory(conn db.Connection, debug bool, collector metricsCollector) (I
 		return nil, err
 	}
 
+	//ensure the cache get's initialized just once
+	mutex.Lock()
+	if kcCache == nil {
+		kcCache = ttlcache.New[string, string](
+			ttlcache.WithTTL[string, string](maxTTL),
```
jeremyharisch, line 72: You need to disable the touch, otherwise TTL will always be renewed and never triggered

tobiscr: aaah, nice catch man! :) 

https://github.com/jellydator/ttlcache/blob/v3/options.go#L73

tobiscr:
Closing are replaced by #1433

PR #1433, opened 2024-01-18
feat: Cache Kubeconfigs

jeremyharisch:
Since we decided to implement the cache as simple as possible with least minimal effort, due to the end of life of the reconciler in a couple of weeks, the implementation is a s simple as possible without adhering to the good practices of Go.

- Introduce Cache mechanism for KubeConfigs
- Each cached item has a TTL of 5min; thus we keep it simple:
    - If Secret gets rotated by the Infrastructure manager, it will be invalid for a maximum of 5min and then gets fetched again and stored in the cache
    - Using a global variable for the cache
    - Set `WithDisableTouchOnHit` for cache, so TTL does not get renewed on successful fetch from cache
- Fixed linting


pkg/cache/kubeConfig.go
```diff
@@ -0,0 +1,82 @@
...
+
+var kubeConfigCache = ttlcache.New[string, string](
+	ttlcache.WithTTL[string, string](5*time.Minute),
+	ttlcache.WithDisableTouchOnHit[string, string](),
+)
+
+// GetKubeConfigFromCache returns the kubeconfig from the cache if it is not expired.
+// If it is expired, it will get the kubeconfig from the secret and set it in the cache.
+func GetKubeConfigFromCache(logger *zap.SugaredLogger, clientSet *kubernetes.Clientset, runtimeID string) string {
+	kubeConfigCache.DeleteExpired()
```
Tomasz-Smelcerz-SAP, line 24: Out of curiosity: Is it necessary? Doesn't the cache evict items automatically/internally?

pkg/cache/kubeConfig.go
```diff
@@ -0,0 +1,82 @@
...
+	kubeConfigFromCache := kubeConfigCache.Get(runtimeID)
+
+	if kubeConfigFromCache.Value() == "" || kubeConfigFromCache.IsExpired() {
+		kubeConfigCache.Delete(runtimeID)
+		kubeConfig := getKubeConfigFromSecret(logger, clientSet, runtimeID)
+		SetKubeConfigInCache(runtimeID, kubeConfig)
```
Tomasz-Smelcerz-SAP, line 30: No error is returned here. The "error condition" is only indicated by the `kubeConfig` being empty string.
If so, I recommend to only put this value into the cache if the string is not empty.

pkg/cache/kubeConfig.go
```diff
@@ -0,0 +1,82 @@
...
+		logger.Errorf("Kubeconfig-secret '%s' for runtime '%s' does not include the data-key 'config'",
+			kubeConfigName, runtimeID)
+	}
+
+	return string(kubeconfig)
+}
+
+// getKubeConfigSecret gets the kubeconfig secret from the cluster.
+func getKubeConfigSecret(logger *zap.SugaredLogger, clientSet *kubernetes.Clientset,
+	runtimeID, kubeConfigName string) (secret *corev1.Secret, err error) {
```
Tomasz-Smelcerz-SAP, line 61: `kubeConfigName` should be renamed to `secretResourceName` or something like that.

pkg/reconciler/instances/rma/integration_action.go
```diff
@@ -362,12 +369,13 @@ func findLatestRevision(releases []*release.Release) *release.Release {
...
 	if err != nil {
-		context.Logger.Debugf("got error %s when converting string to int for configuration: %s, use its default value: %d", err, RmiVmalertGroupsNum, DefaultVMAlertGroupsNum)
+		context.Logger.Debugf("got error %s when converting string to int for configuration: %s, use its default value: %d",
+			err, RmiVmalertGroupsNum, DefaultVMAlertGroupsNum)
 		groups = DefaultVMAlertGroupsNum
 	}
 
 	csum := sha256.Sum256([]byte(id))
-	mrand.Seed(int64(binary.LittleEndian.Uint64(csum[0:8])))
+	mrand.New(mrand.NewSource(int64(binary.LittleEndian.Uint64(csum[0:8]))))
```
Tomasz-Smelcerz-SAP, line 378: Same problem with `math/rand` - the returned object is not used, it looks like this has no effect.

tobiscr:
JFYI-  I have also a proposal (probably overlooked - I mentioned it in your Slack channel :D ): 

https://github.com/kyma-incubator/reconciler/pull/1432

pkg/cache/kubeConfig.go
Tomasz-Smelcerz-SAP: :+1:

Tomasz-Smelcerz-SAP (dismissed):

pkg/cache/kubeConfig.go
```diff
@@ -0,0 +1,84 @@
...
+	secret, err := getKubeConfigSecret(logger, clientSet, runtimeID, secretResourceName)
+	if err != nil {
+		return ""
+	}
+
+	kubeconfig, found := secret.Data["config"]
+	if !found {
+		logger.Errorf("Kubeconfig-secret '%s' for runtime '%s' does not include the data-key 'config'",
+			secretResourceName, runtimeID)
+	}
```
tobiscr, line 56: This is acutally a failure and the process should stop with an error.

The logic currently implies that the developer who is using this method has to know that invalid values can be returned.

Explicit error handling by throwing an error avoid this.

pkg/cache/kubeConfig.go
```diff
@@ -0,0 +1,84 @@
...
+func SetKubeConfigInCache(key string, kubeconfig string) {
+	kubeConfigCache.Set(key, kubeconfig, 5*time.Minute)
+}
+
+// getkubeConfigFromSecret gets the kubeconfig from the secret.
+func getKubeConfigFromSecret(logger *zap.SugaredLogger, clientSet *kubernetes.Clientset, runtimeID string) string {
+	secretResourceName := fmt.Sprintf("kubeconfig-%s", runtimeID)
+	secret, err := getKubeConfigSecret(logger, clientSet, runtimeID, secretResourceName)
+	if err != nil {
+		return ""
```
tobiscr, line 49: Throw the error please, empty kubeconfigs are not valid. If the empty value gets returned, the callers of this method have to deal again with this exceptional case by their own.

pkg/cluster/inventory.go
```diff
@@ -180,32 +177,16 @@ func (i *DefaultInventory) createCluster(contractVersion int64, cluster *keb.Clu
...
-			} else if k8serr.IsForbidden(err) { // configuration failure
-				i.Logger.Warnf("Cluster inventory is not allowed to lookup kubeconfig-secret '%s' for cluster with runtimeID %s: %s",
-					kubeconfigName, cluster.RuntimeID, err)
-				return
-			} else {
-				i.Logger.Errorf("Cluster inventory failed to lookup kubeconfig-secret '%s' for cluster with runtimeID %s: %s",
-					kubeconfigName, cluster.RuntimeID, err)
-			}
+		kubeconfig := cache.GetKubeConfigFromCache(i.Logger, i.clientSet, cluster.RuntimeID)
+		if kubeconfig == "" {
```
tobiscr, line 181: Better to use an error in the response for indicating wrong values.

tobiscr (changes requested):

pkg/cache/kubeConfig.go
```diff
@@ -0,0 +1,84 @@
...
+// If it is expired, it will get the kubeconfig from the secret and set it in the cache.
+func GetKubeConfigFromCache(logger *zap.SugaredLogger, clientSet *kubernetes.Clientset, runtimeID string) string {
+	kubeConfigCache.DeleteExpired()
+	kubeConfigFromCache := kubeConfigCache.Get(runtimeID)
+
+	if kubeConfigFromCache.Value() == "" || kubeConfigFromCache.IsExpired() {
+		kubeConfigCache.Delete(runtimeID)
+		kubeConfig := getKubeConfigFromSecret(logger, clientSet, runtimeID)
+		if kubeConfig != "" {
+			SetKubeConfigInCache(runtimeID, kubeConfig)
```
tobiscr, line 31: Using an error is better to indicate a wrong kubeconfig-value

pkg/cache/kubeConfig.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		return kubeConfig
+	}
+
+	return kubeConfigFromCache.Value()
+}
+
+// SetKubeConfigInCache sets the kubeconfig in the cache.
+func SetKubeConfigInCache(key string, kubeconfig string) {
+	kubeConfigCache.Set(key, kubeconfig, 5*time.Minute)
```
tobiscr, line 41: The method includes just one line of code and this line is also quite self-explaining. Wrapping by another method is not needed actually.

tobiscr (approved):

pkg/cache/kubeConfig.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/jellydator/ttlcache/v3"
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var kubeConfigCache = ttlcache.New[string, string](
+	ttlcache.WithTTL[string, string](5*time.Minute),
```
tobiscr, line 17: Are 5min not too short? We reconcile with 20min intervals.. in theory we could increase the value to something <20min

Issue #1447, opened 2024-01-25
Introduce warden reconciler to cleanup cert volume mounts


kwiatekus:
**Description**

Introduce warden component reconciler that would clean up the removed volume mount

**Reasons**

Warden 0.10.0 introduced a [change](https://github.com/kyma-project/warden/pull/117/files#diff-20dad112397d4e28bacb5d114c2be35642b2e157fdc869f3607aeb41d3fbdde0) that removed volume mounts.

When 0.10.0 was [introduced to stage ](https://github.tools.sap/kyma/management-plane-config/pull/4826/files) reconciler havent cleaned up the volume mounts - causing an Crashloopbackoff [incident](https://github.tools.sap/kyma/backlog/issues/4985) for warden admission on many of the managed clusters.

```
19T09:11:07Z","caller":"admission/main.go:103","message":"failed to save certificate from secretfailed to save server cert to file '/tmp/k8s-webhook-server/serving-certs/server-cert.pem': open /tmp/k8s-webhook-server/serving-certs/server-cert.pem: read-only file system","context":{}}
```

This should be avoided when installing warden 0.10.0 on prod.

kwiatekus:
blocked because cannot [provision](https://sap-btp.slack.com/archives/C01LGCBS196/p1706872119673359) and test on DEV

PR #1448, opened 2024-01-26
Warden volume mount cleanup reconciler

kwiatekus:
**References**
https://github.com/kyma-incubator/reconciler/issues/1447

### DIFF
--- a/pkg/reconciler/instances/warden/volumemounts.go
+++ b/pkg/reconciler/instances/warden/volumemounts.go
@@ -18,7 +18,6 @@ const wardenAdmissionDeploymentName = "warden-admission"
 const wardenAdmissionDeploymentNamespace = "kyma-system"
 const volumeName = "certs"
 
-// TODO: please implement component specific action logic here
 type CleanupWardenAdmissionCertColumeMounts struct {
 	name string
 }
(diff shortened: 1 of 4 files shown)


pPrecel (approved):

pkg/reconciler/instances/warden/warden.go
```diff
@@ -0,0 +1,24 @@
...
+
+	log.Debugf("Initializing component reconciler '%s'", ReconcilerName)
+	reconciler, err := service.NewComponentReconciler(ReconcilerName)
+	if err != nil {
+		log.Fatalf("Could not create '%s' component reconciler: %s", ReconcilerName, err)
+	}
+
+	reconciler.
+		WithPreReconcileAction(&CleanupWardenAdmissionCertColumeMounts{
+			name: "cleanup-warden-admissio-cert-volume-mounts-action",
```
Tomasz-Smelcerz-SAP, line 22: What is this `name` used for?

pkg/reconciler/instances/warden/volumemounts.go
```diff
@@ -0,0 +1,121 @@
...
+			volumeIndex := getVolumeIndexByName(deployment, volumeName)
+			volumeMountIndex := getVolumeMountIndexByName(deployment, volumeName)
+
+			if volumeIndex == -1 || volumeMountIndex == -1 {
+				context.Logger.Debugf("no action needed for warden admission deployment before applying image %s", targetImage)
+				return nil
+			}
+
+			context.Logger.Debugf("warden admission deployment qualifies for Volume[%d] nad VolumeMount[%d] cleanup before applying image %s", volumeIndex, volumeMountIndex, targetImage)
+			data := fmt.Sprintf(`[{"op": "remove", "path": "/spec/template/spec/containers/0/volumeMounts/%d"},{"op": "remove", "path": "/spec/template/spec/volumes/%d"}]`, volumeMountIndex, volumeIndex)
```
Tomasz-Smelcerz-SAP, line 47: It would be nice to have an unit-tested function that creates this JSON-patch data.

pkg/reconciler/instances/warden/volumemounts.go
```diff
@@ -0,0 +1,121 @@
...
+	for p, v := range deployment.Spec.Template.Spec.Volumes {
+		if v.Name == volumeName {
+			return p
+		}
+	}
+	return -1
+}
+
+func getVolumeMountIndexByName(deployment *appsv1.Deployment, volumeMountName string) int {
+	for p, v := range deployment.Spec.Template.Spec.Containers[0].VolumeMounts {
```
Tomasz-Smelcerz-SAP, line 98: This is kind of risky IMO. If any sidecar is injected (Istio?), then the proper container "index" may change from `0` to `1` or so.

pkg/reconciler/instances/warden/warden.go
kwiatekus: yes. Now  it is used in log that is the first line in the `Run` function.

pkg/reconciler/instances/warden/volumemounts.go
kwiatekus: What do you mean?
There is a mock call verification in the unit test (if expected call is recorded)
Wouldn't testing the line 47 be just testing if `fmt.Sprintf` works? I'd like to avoid testing that tbh.

pkg/reconciler/instances/warden/volumemounts.go
kwiatekus: Agreed. Even though we own warden manifest and we do control the amount of container, better be safe than sorry. I changed the logic

Tomasz-Smelcerz-SAP (approved):

PR #1450, opened 2024-01-31
Specify group and version while searching for connectivity-proxy resources and logging improvement

koala7659:
- Reduce log noise from connectivity-proxy reconciller by setting log level of most messages to Debug
- Make sure all reconcilation exit paths are always logged 
- Use API group version when searching for connectivity-proxy service instances
- Update mocks and unit tests


tobiscr (approved):

m00g3n (approved):
LGTM

